PR #12, opened 2017-07-31
Rewrite use case 3 (Action 28)

RubenVerborgh:
Resolves https://www.w3.org/2017/dxwg/track/actions/28.

### DIFF
--- a/ucr/index.html
+++ b/ucr/index.html
@@ -233,7 +233,7 @@ it is necessary that servers can add more detail.</p>
       </p>
     </section>
     <section id="ID3" class="usecase">
-      <h2>Combining multiple types of content sections in a single response [ID3]</h2>
+      <h2>Responses can conform to multiple, modular profiles [ID3]</h2>
       <p class="contributor">Ruben Verborgh</p>
       <p class="tags">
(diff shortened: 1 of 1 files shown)


RubenVerborgh:
> this seems to change the original intent

No, it's a (fundamental) rephrasing of the same use case, following what was discussed during the F2F. The original use case explained the fact that one resource can conform to multiple profiles, by drawing an analogy with multiple content areas in HTML. However, it was not intended as multi-part, but rather that a single (RDF or other) response can conform to different profiles (for instance, because there are multiple content sections). However, the multiple content sections are not a requirement, hence the rephrasing.

The important thing, both in the original and rephrased text, is the conformance to multiple profiles. Multiple sections are just one way to realize that.

>  to choosig alternative responses which I think have been covered.

This is not about choosing alternatives. Instead, it is about single representations that conform to multiple profiles.

> If that was the intent, then we should validate that multipart responses are either in or out of scope.

The intent was multi-profile conformance and has now been clarified in the rewriting. (Note that we did not discuss whether multipart responses during the F2F; I did talk about the intent of this use case.)

PR #18, opened 2017-08-13
Ixchels editorial work

rob-metalinkage:
Passing up into master.


rob-metalinkage:
Merged both Ixcehls commits and more major deduplication efforts as i fixed some syntax at the same time to allow page to render.

PR #20, opened 2017-08-21
Added Use Case #49 => Annotating changes that do not change the infor…

pwin:
…mation content [ID49]


rob-metalinkage:
did this locally to check formatting after conflict resolution and submitted new PR

PR #26, opened 2017-09-06
Add reference to RAMLET in ID32

dr-shorthair:
Appears to be relevant prior art


agbeltran:
@rob-metalinkage should we delete the branch as this has been merged already? thanks

dr-shorthair:
Indeed. See https://www.w3.org/2017/dxwg/wiki/GitHub_etiquette 
"5. If the pull-request has been accepted, then it is recommended to delete the branch, so that the repository does not get too cluttered!"

agbeltran:
Thanks @dr-shorthair - I will delete the branch.

PR #35, opened 2017-09-28
suggested additions to the text of Rver1 and Rver2 as per https://www…

pwin:
….w3.org/2017/dxwg/track/actions/43


jpullmann:
Thanks for your proposal, Peter. There is a pending action to [discover which resources need versioning, and create specific requirements](https://www.w3.org/2017/dxwg/track/actions/44). It is expected to replace this entire (generic) requirement, let's wait for resolution.

jpullmann:
ACTION-43 closed, the request resolved in recent version

PR #42, opened 2017-11-30
Changes to UCR document introduction

agbeltran:
According to action item: https://www.w3.org/2017/dxwg/track/actions/65

### DIFF
--- a/ucr/index.html
+++ b/ucr/index.html
@@ -35,14 +35,34 @@
     </section>
     <section id="Introduction">
       <h1>Introduction</h1>
+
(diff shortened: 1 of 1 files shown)


larsgsvensson (commented):
I'm afraid those changes don't address the issues I had with the _abstract_ (not the _introduction_).
The changes in the introduction are fine with me.

agbeltran:
Thank you @larsgsvensson - my action was about the intro, but yes, the abstract needs to be modified too. Thanks!

PR #47, opened 2017-12-19
dcat v1.0 ontology to serve as a starting point

dr-shorthair:
copied directly from https://www.w3.org/ns/dcat.ttl
plus imports added for FOAF, SKOS, dcterms, prov-o


agbeltran:
Thanks @dr-shorthair - will merge now. Should we also reference it from the document?

dr-shorthair:
OK - see https://github.com/w3c/dxwg/pull/48

Issue #77, opened 2018-01-18
Project relation [RPR]


jpullmann:
<section class='requirement' id='RPR'><h3>Project relation [<a href='https://www.w3.org/TR/dcat-ucr#RPR' target='_blank'>RPR</a>]</h3><p>Provide a means to indicate the relation of Datasets to a project.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID49' target='_blank'>Dataset business context [ID49]</a> </section>

dr-shorthair:
Related to #71

dr-shorthair:
The relation should be to a generic activity (not just a project, which is a very specific kind of activity)
In which case we can make datasets a kind of Entity from the [PROV-O ontology](https://www.w3.org/TR/prov-o/), by adding 
  `dcat:Dataset rdfs:subClassOf prov:Entity .`
Then the `prov:wasGeneratedBy` predicate can be used to satisfy this requirement directly.

dr-shorthair:
Suggest removing the following tags:
[version] [usage control] [status] [service] [profile]

dr-shorthair:
Resolved by #312

Issue #76, opened 2018-01-18
Provenance information [RPIF]


jpullmann:
<section class='requirement' id='RPIF'><h3>Provenance information [<a href='https://www.w3.org/TR/dcat-ucr#RPIF' target='_blank'>RPIF</a>]</h3><p>Provide a way to link to structured information about the provenance of a dataset including:<ul><li>the input data used to create a dataset to the dataset.</li><li>the software used to produce the dataset to the dataset.</li><li>an extensible model different types of agent roles</li><li>funders</li></ul></p><p class='note'>dct:creator, dct:publisher etc are special cases, which require guidance, further roles may be defined in provenance or other richer models. The requirement is to establish an extensible mechanism, and for profiles to specify canonical equivalents for the special case properties of dcat:Dataset</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID9' target='_blank'>Common requirements for scientific data [ID9]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID12' target='_blank'>Modeling data lineage [ID12]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID13' target='_blank'>Modeling agent roles [ID13]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID31' target='_blank'>Modeling funding sources [ID31]</a> </section>

andrea-perego:
About agent roles, I have a punctual proposal which is about relaxing the domain of `dcat:contactPoint`, to allow its use not only for datasets, but also for other resources (e.g., catalogues, catalogue records).

This issue popped up during the development of GeoDCAT-AP, since all the agent roles (`dcat:contactPoint` included) supported in ISO 19115 can be specified for any resource.

Besides this, the "contact point" is probably the most important role for data consumers, not only for datasets. For instance, for a `dcat:Catalog` it is possible to specify the `dct:publisher`, but if I need to ask questions and/or report issues about the catalogue I need to get in touch with the publisher's dedicated contact point.

I wonder whether this (and similar revisions to DCAT) requires the creation of a separate use case.

> :warning: As decided at the end of [the DCAT subgroup telecon of 31 Jan 2018](https://www.w3.org/2018/01/31-dxwgdcat-minutes.html), a separate issue has been created (#95)

nicholascar:
Re-represented as RDA Prov Patterns WG Use Case 41: [http://patterns.promsns.org/usecase/41](http://patterns.promsns.org/usecase/41)

[...]

PR #94, opened 2018-01-19
Create PROV-alignment module

dr-shorthair:
Integration with prov-o, as proposed in #76, #77 and all of https://github.com/w3c/dxwg/issues?q=is%3Aissue+label%3Aprovenance


dr-shorthair:
ALso relates to #76

nicholascar:
Re-represented as RDA Prov Patterns WG Use Case 42: [http://patterns.promsns.org/usecase/42](http://patterns.promsns.org/usecase/42)

riccardoAlbertoni:
We had a similar discussion in the DWBP when designing the DQV. 
In the DQV we decided not to do so! However, I guess there might be reasons to do otherwise in the case of DCAT ...

Some considerations against making dcat:Dataset a subclass of prov:Entity:
-What is the advantage of explicitly declaring the `dcat:Dataset` as a subclass of `prov:Entity`?
Being 'prov:Entity' the domain of properties such as prov:wasGeneratedBy,   `prov:Entity` is automatically everytime we use a PROV property on a dcat:Dataset. So by not declaring `dcat:Dataset` subclass of `prov:Entity`,  you have that `dcat:Dataset` will be  `prov:Entity` only when PROV properties are actually used, which in my opinion might be quite useful.

-Also,  if we want to follow some design best practice such as the "minimal ontological commitment",    we might not want to over-constraint the DCAT vocabulary,  as this can affect the potential reuse of the vocabulary in the future.
For example, in ten years,  future DCAT re-users might want to use an alternative provenance vocabulary, or they might need to define an instance of `dcat:Dataset` also as an instance of a `prov:Entity` disjoint class. 
I acknowledge this might seem a remote eventuality, and that discussion might seem a little academic,  but Do we want to leave out this possible reuses of DCAT?

said that, I guess there are reasons to define `dcat:Dataset` as `prov:Entity` which I am totally overlooking, aren't they?

dr-shorthair:
1. IMO DCAT application is more closely aligned to Provenance concerns even than DQV. A significant fraction of dataset metadata (the scope of DCAT) is Provenance. 

2. The primary concern with linking to another ontology is whether it introduces undesirable entailments.   AFAIK the critical axioms relating to prov:Entity are only that the class is disjoint with Activity and InstantaneousEvent. Everything else is in PROV-O are 'opportunities' rather than 'constraints'. 

So as long as dcat Datasets are understood as Entities and the fundamental distinction between occurrent/continuant (approximately perdurant/endurant) is acceptable, then I see no risk in making the alignment fundamental to DCAT. And in particular, I believe it will accelerate resolution of a significant number of the formal requirements.

riccardoAlbertoni:
Hi Simon, thanks for your reply. 

I am very much in favour of reusing PROV. Actually, I consider myself a great fan of PROV ;)

My comment aimed to bring in the discussion that subclassing `prov:Entity` is not the only option we have on the table to reuse PROV. 

For example,  we might consider recommending the  PROV reuse by referring to the DCT PROV mapping [1] or by including/specializing the prov properties without explicit `prov:Entity` subclassing.

Let me make an example to explain why I am bringing this issue up. 

example 1: assuming we define `dcat:Dataset` as a subclass `prov:Entity`, and we have the following dcat definitions

```
:dataset1 a  dcat:Dataset, 
 # add  some dcterms statements
.

:dataset2 a  dcat:Dataset, 
#add some dctems statements
prov:wasDerivedBy :dataset1
.

....
# a certain number of  dcat datasets defined without any prov specific properties, let say 2000 datasets
:dataset_n a dcat:Dataset,
# add  some dcterms statements
.
```

Considering the example above, if we query for the  `prov:Entity` instances
we will have 2002 instances, but only dataset2 has actual prov statements, which IMO is slightly annoying ...

I am not necessarily against the `prov:Entity` subclassing.  I just think that if we say that every `dcat:Dataset` is a `prov:Entity`, we should make sure that some actual prov statement is attached. Otherwise, the subclassing is not adding much and it might bring to tons of `prov:Entity` without much actual PROV inside.

Does my point have any sense for you?

Cheers, 
Riccardo


[1] https://www.w3.org/TR/prov-dc/

nicholascar:
If we map DCAT Dataset/Agent relations to PROV Entity/Agent qualified relations (`:Dataset_X dct:publisher :Publisher_P == :Dataset_X prov:qualifiedAttribution [a prov:Attribution ; prov:hadRole :publisher ; prov:agent :Publisher_P ]`, then all Datasets will be able to be understood as having PROV properties.

I think we should make these mappings and maximally reuse the qualified Entity/Agent patterning.

dr-shorthair:
@riccardoAlbertoni : 

1. If `dcat:Dataset rdfs:subClassOf prov:Entity .` in the core DCAT graph of definitions, 
then your example 1 will only generate 2002 results if the query engine has reasoning enabled, or if the query explicitly asks for individuals whose type is a subclass of `prov:Entity`. 

2. prov:Entity is a pretty general concept, and the OWA says that even if no `prov:` properties are known, it does not mean they do not exist. 

In fact I would suggest that anyone who is running a reasoner would be well aware of the OWA so would understand the patterns in the 2002 results ...

riccardoAlbertoni:
@dr-shorthair
During the DCAT Call two or three weeks ago,  you pointed out that an agreement on this issue might depend on what kind of "reasoning"  are we expecting consuming the DCAT metadata, and also whether or not we are going to provide constraints with SHACL. I think you were right,  have we made any progress on the two points above?

Perhaps, another relevant point is to decide between

-  a "pay-as-you-go" use of provenance, where we support catalogs in gradually moving to a PROV enabled version when need, but we accept that datasets and catalogs might not always need a full use of the provenance;
 or 
- a "fully-fledged provenance anyway" approach, where we force the use of a fully-fledged provenience in every DCAT dataset and catalog.

I tend to prefer the pay-as-you-go approach, as  I am not sure that having the PROV qualified form is strictly necessary for all the DCAT catalog. For example,  I would prefer to recommend the materialization of DCT-PROV mapping only for certain kind of catalogs, e.g. the catalogs That harvest DCAT dataset from third-party catalogs.

aisaac:
For what it's worth, it's not necessary to declare a class as a subclass of prov:Entity to be able to use PROV statements on the instances of this class. I don't understand the rationale behind the latter part of the sentence "Formally indicate that dcat:Dataset, in a future update to DCAT, is a subClassOf prov:Entity so it can be used as such." at http://patterns.promsns.org/usecase/42 ...

dr-shorthair:
@aisaac yeah, OWA means you can use any property anywhere. Just need to keep an eye out for the consequent entailments, as I discussed above. I don't think there are any adverse consequences since the only constraint on prov:Entity of significance is that it is disjoint with prov:Activity. 

@riccardoAlbertoni very happy to nudge people along to the full glories of PROV with a 'pay as you go' approach. Adding the subclass axiom wouldn't cost anything in the short term unless there were some fierce SHACL rules as well. And I can't see us proposing those in general  (though specific profiles of DCAT might). But I do see a benefit of adding the prov:Entity relationship now because it would tell users '_if you do want to record complex provenance metadata, then use PROV, rather than inventing something new_'.

aisaac:
@dr-shorthair there's no crucial adverse consequence indeed. But it adds one class to the DCAT mix, it diverts attention to what really matters (i.e. which parts of PROV to re-use besides prov:Entity) and it doesn't bring anything clear to meeting the requirements. So for the moment it looks like some dead weight we'd put in DCAT. A small one I reckon, but still some dead weight.

I'm happy to go into the details of requirements #76 and #77 but I'm a bit confused as to where the discussion should go, as #128 is now about the same matter.

riccardoAlbertoni:
+1 to Antoine. 

@antoine
>  there's no crucial adverse consequence indeed

Not sure, maybe the adversities are not so crucial, but  the tons of inferred 
prov:Entity added to dcat:Dataset instances without no actual PROV properties are in my opinion a quite annoying risk coming from the prov:Entity subclassing.

@dr-shorthair
> I do see a benefit of adding the prov:Entity relationship now because it would tell users 'if you do want to record complex provenance metadata, then use PROV, rather than inventing something new'.

I see your goal,  but I  am not sure the subclassing is an effective solution to meet it:  probably a consumer who recognizes what a prov:Entity is,  already knows PROV. 

I suspect that explicitly stating in the DCAT document 
> if you do want to record complex provenance metadata, then use PROV, rather than inventing something new

is still our best option to meet the goal, at least at this stage of the discussion. 

Perhaps we want to return to this specific issues after that we have solved  #128 ...

dr-shorthair:
I was premature in raising this pull-request - should have raised an issue first. Then I realised that CatalogRecord was involved as well as Dataset, so I created #128 which is more general. 

This PR should probably be closed without merging, except that there is some valuable discussion relevant to #128.

dr-shorthair:
Following discussion in #111  and #128 I've moved the proposed PROV alignment into a separate file/graph. I also included a comment on the status of this code, and to indicate that it is floating an idea for how to package dependencies and alignments.

aisaac:
@dr-shorthair if the idea is that there's a separate file that we can review later against (both functional and non-functional) requirements, then it's good. Is the plan also to leave either this ticket or #128 open so that it's indeed still 'floating'? I guess this would help reminding us about it in due time. And quite acceptable from the perspective of our ticket process!

dr-shorthair:
Perhaps merge this and close #94, leaving #128 open

dr-shorthair:
Perhaps merge this and close #94, leaving #128 open

aisaac:
Sounds good!

dr-shorthair:
#134 approved, so I think this PR can be accepted/merged

agbeltran:
Given the resolution in the [DCAT group meeting](https://www.w3.org/2018/02/28-dxwgdcat-minutes.html) about using RDF vocabularies for modularization, at least a discussion step, as described in https://github.com/w3c/dxwg/issues/134, I will review and merge this PR.

agbeltran (approved):

PR #99, opened 2018-02-01
Inserted some paragraphs of type 'issue'

dr-shorthair:
to alert readers to places where changes to the DCAT vocabulary have been proposed.


agbeltran (approved):
Thanks @dr-shorthair - it is good to include these notes, and hopefully we will update them as a result of those issues being resolved.

PR #100, opened 2018-02-01
Added links

agbeltran:
I should have added this directly to your PR before merging, but here it goes as a separate one.


dr-shorthair:
Good.

Issue #64, opened 2018-01-18
Dataset type [RDST]


jpullmann:
<section class='requirement' id='RDST'><h3>Dataset type [<a href='https://www.w3.org/TR/dcat-ucr#RDST' target='_blank'>RDST</a>]</h3><p>Provide a mechanism to indicate the type of data being described and recommend vocabularies to use given the dataset type indicated.</p><p class='note'>Providing examples of scope will provide guidance, without being unnecessarily restrictive. The key requirement is interoperability, achieved by using standardised vocabulary terms. It it unclear whether a canonical registry is required or whether communities should constrain choice via DCAT profiles.</p><br/><em>Related requirements</em>: <a href='https://www.w3.org/TR/dcat-ucr#RDSAT' target='_blank'>Dataset aspects [RDSAT]</a> <br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID8' target='_blank'>Scope or type of dataset with a DCAT description [ID8]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID20' target='_blank'>Modelling resources different from datasets [ID20]</a> </section>

dr-shorthair:
Some existing type vocabularies: 

**DCMI Type** vocabulary - http://dublincore.org/documents/dcmi-terms/#section-7

**ISO 19115 Scope Code** vocabulary - http://registry.it.csiro.au/def/isotc211/MD_ScopeCode https://geo-ide.noaa.gov/wiki/index.php?title=ISO_19115_and_19115-2_CodeList_Dictionaries#MD_ScopeCode

**PARSE.Insight content-types** recommended for Re3data - http://www.re3data.org/schema http://gfzpublic.gfz-potsdam.de/pubman/item/escidoc:1397899 item 15 contentType

The last one takes some burrowing to find inside a PDF, so I've copied the contents of the key table here: 

15 contentType

PARSE.Insight type | Examples of File Formats

Standard office documents | text documents, spreadsheets, presentations
Networkbased data | websites, email, chat history, etc.
Databases | DBASE, MS Access, Oracle, MySQL, etc.
Images | JPEG, JPEG2000, GIF, TIF, PNG, SVG, etc.
Structured graphics | CAD, CAM, 3D, VRML, etc.
Audiovisual data | WAVE, MP3, MP4, Flash, etc.
Scientific and statistical data formats | SPSS, FITS, GIS, etc.
Raw data | device specific output
Plain text | TXT in various encodings
[...]

PR #103, opened 2018-02-02
Associate dct:type with dcat:Dataset

dr-shorthair:
Add an existential qualifier to formally associate dct:type with dcat:Dataset
See #64


makxdekkers:
I do not know what "_an existential qualifier to formally associate_" means. Is it not enough to include a reference in DCAT v2 mentioning dct:type as one of the possible properties in the description of Dataset?

dr-shorthair:
OK - don't use existential qualifier, use recommendation in text instead see #105.

Issue #98, opened 2018-02-01
remove subsumption axiom `dcat:Dataset rdfs:subClassOf dctype:Dataset.`


dr-shorthair:
dctype:Dataset is part of the [DCMI Type Vocabulary](http://dublincore.org/documents/dcmi-terms/#section-7), alongside text, image, sound, software and some other genre terms. 

In DCAT 1.0 dcat:Dataset is a subclass of dctype:Dataset. However, the scope of DCAT 'dataset' is more general than the DCMI Type 'dataset' (see [Scope or type of dataset with a DCAT description [ID8]](https://www.w3.org/TR/dcat-ucr/#ID8)), and also includes 'text' at least, and maybe multimedia types as well - the Datacite-DC-DCAT mapping provided at https://github.com/w3c/dxwg/issues/64#issuecomment-362113915 certainly shows this. 

So, while the DCMI Type Vocabulary could reasonably used as a typing vocabulary for classifying a dcat:Dataset (see #64), it is incorrect to pre-empt this by making dcat:Dataset a subclass of dctype:Dataset.

makxdekkers:
I agree that it is too restrictive to declare dcat:Dataset a subclass of dctype:Dataset.

dr-shorthair:
Resolved in meeting https://www.w3.org/2018/02/07-dxwgdcat-minutes
Implemented in #108

agbeltran:
Re-opening until the PR is merged.

dr-shorthair:
@agbeltran action on the PR is currently with you. 
Do you see any problems?

agbeltran:
@dr-shorthair the PR is actually about two issues (I suggest we do a PR per issue, what do you think)? I'm still reviewing one of the issues.

dr-shorthair:
This issue is fully resolved by #108 so can be closed. 
#64 should remain open until documentation is completed, per the comments. 

I combined two changes in one PR as they are logically bound together - the subsumption axiom is replaced by (... generalised to ...) the dct:type property.

PR #108, opened 2018-02-08
dcat:Dataset type

dr-shorthair:
1. remove subclass axiom dcat:Dataset subclass of dctype:Dataset #98 
2. add recommendation to use dct:type to indicate the type or genre of a data set, with value taken from suitable controlled vocabulary (examples provided) #64


agbeltran (changes requested):
Given the discussion in #64, the current text on dataset types seems too brief. Shouldn't we add more examples apart from this:

```
<p>
	The type or genre of a dataset may be indicated using the <a href="http://purl.org/dc/terms/type">dct:type</a> property:
</p>

<pre>  :dataset-001 dct:type dctype:Text . </pre>
```

agbeltran:
@dr-shorthair looking at your last comment in #64, it seems you weren't intending to close the issue but this PR is just a first step? However, are you intending to close the issue #98 with this PR? Some clarity on this will be helpful.

dr-shorthair:
Yes - PR is a first step. I've made the necessary change to the RDF, and also to section 6 in the Rec document. 

The discussion and examples in section 5 need more work: in particular how to use classifiers from type vocabularies where the items do not have individual URIs.  I think it is useful to merge this PR, but leave #64 open until the requirement has been fully satisfied with more documentation and examples.

agbeltran (approved):

agbeltran:
OK, thanks for clarifying. Merging now.

Issue #80, opened 2018-01-18
Mapping of qualified and non-qualified forms [RQFM]


jpullmann:
<section class='requirement' id='RQFM'><h3>Mapping of qualified and non-qualified forms [<a href='https://www.w3.org/TR/dcat-ucr#RQFM' target='_blank'>RQFM</a>]</h3><p>Specify mapping of qualified to non-qualified forms (lowering mapping). The reverse requires information (qualification) that might not be present/evident.</p><p class='note'>This requirement is still under review</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID19' target='_blank'>Guidance on the use of qualified forms [ID19]</a> </section>

nicholascar:
I think this is an important Req but can it be un-tagged with "provenance"? The reason is that the PROV-O style of qualification should be used for both provenance and non-provenance as a general ontology pattern, therefore this is not provenance-specific.

riccardoAlbertoni:
I suggest to untag "quality",  this is not a quality-specific requirement, it affects the general way in which we want to model thing in dcat.  In some of the related use cases, DQV is mentioned as a source for possible modelling examples but the requirement itself is not related to quality.

dr-shorthair:
I've attempted an axiomatization of the simple form in terms of a qualified forms in the Wiki Page here: https://github.com/w3c/dxwg/wiki/Qualified-relations#simple-relations

dr-shorthair:
See axiomatization of some of the relations here: https://github.com/w3c/dxwg/blob/dcat-issue79-simon/dcat/rdf/dcat-prov.ttl#L63

dr-shorthair:
Branch gone, now here https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-prov.ttl#L63

dr-shorthair:
Issue discussed in DCAT meeting 2019-02-07. As it is a niche concern likely to be only of interest to OWL mavens, it is low priority and will not be addressed further.

agbeltran:
Link to minutes: https://www.w3.org/2019/02/06-dxwgdcat-minutes.html

PR #136, opened 2018-02-26
Added qualified terms issue to publisher property

nicholascar:
representing Issue 80 in the document. Qualification may eventually apply in many places in DACT 1.1/2.0 but this is a current tangible place.


agbeltran (approved):

agbeltran:
Linking this PR to the corresponding issue for reference: https://github.com/w3c/dxwg/issues/80

Issue #65, opened 2018-01-18
Entailment of schema.org [RES]


jpullmann:
<section class='requirement' id='RES'><h3>Entailment of schema.org [<a href='https://www.w3.org/TR/dcat-ucr#RES' target='_blank'>RES</a>]</h3><p>Define schema.org equivalents for DCAT properties to support entailment of schema.org compliant profiles of DCAT records.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID40' target='_blank'>Discoverability by mainstream search engines [ID40]</a> </section>

andrea-perego:
Some work has been carried out in the framework of SDW that can turn to be relevant here:

https://www.w3.org/2015/spatial/wiki/ISO_19115_-_DCAT_-_Schema.org_mapping

The specified mappings were based on existing work, including a mapping of DCAT-AP, GeoDCAT-AP, and StatDCAT-AP to Schema.org:

https://ec-jrc.github.io/dcat-ap-to-schema-org/

makxdekkers:
This would be very useful for tools that could create the relevant schema.org markup for a landing page for the dataset based on the DCAT description.

andrea-perego:
> This would be very useful for tools that could create the relevant schema.org markup for a landing page for the dataset based on the DCAT description.

The DCAT-AP to Schema.org mapping exercise I mentioned earlier in this thread includes also SPARQL queries implementing such mappings:

https://github.com/ec-jrc/dcat-ap-to-schema-org/tree/master/sparql

dr-shorthair:
I propose adding an RDF alignment document for DCAT-schema.org to enable this discussion

dr-shorthair:
More mappings from SDWWG and Project Open Data added in #146

dr-shorthair:
Preliminary DCAT-schema.org alignment here: https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/schema.ttl

dr-shorthair:
This repo https://github.com/earthcubearchitecture-project418/p418Vocabulary#dataset-diagram provides a detailed analysis of how schema.org supports describing a dataset.

azaroth42:
Could someone explain the relevance of `profile-negotiation` here, other than as an instance of negotiating for schema.org vs dcat? :)

dr-shorthair:
@rob-metalinkage responded on the email list: 

[...]

PR #140, opened 2018-02-28
Initial alignment to schema.org

dr-shorthair:
Triggered by #65


davebrowning (approved):
Looks plausible to me, though I'm not a schema.org expert. Given this is marked as provisional/incomplete but is needed as a basis for other work (and an example of modularisation)  then I'm happy to approve for merge.

PR #146, opened 2018-03-06
Additional mappings to schema.org

dr-shorthair:
Added mappings from Project Open Data and SDWWG

### DIFF
--- a/dcat/rdf/schema.ttl
+++ b/dcat/rdf/schema.ttl
@@ -15,6 +15,30 @@
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
+<http://purl.org/dc/elements/1.1/title>
+  owl:equivalentProperty schema:name ;
+.
+dcterms:description
+  owl:equivalentProperty schema:description ;
+.
+dcterms:issued
+  owl:equivalentProperty schema:datePublished ;
+.
+dcterms:language
+  owl:equivalentProperty schema:inLanguage ;
+.
+dcterms:modified
+  owl:equivalentProperty schema:dateModified ;
+.
+dcterms:publisher
+  owl:equivalentProperty schema:publisher ;
+.
+dcterms:spatial
+  owl:equivalentProperty schema:spatial ;
+.
+dcterms:temporal
+  owl:equivalentProperty schema:temporal ;
+.
 schema:DataDownload
   rdfs:subClassOf dcat:Distribution ;
 .
@@ -56,6 +80,7 @@ dcat:distribution
 dcat:downloadURL
   schema:domainIncludes dcat:Distribution ;
   schema:rangeIncludes rdfs:Resource ;
+  owl:equivalentProperty schema:contentUrl ;
 .
 dcat:granularity
   schema:domainIncludes dcat:Dataset ;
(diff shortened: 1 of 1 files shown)


dcat/rdf/schema.ttl
```diff
@@ -15,6 +15,42 @@
...
+dcterms:publisher
+  owl:equivalentProperty schema:publisher ;
+.
+dcterms:spatial
+  owl:equivalentProperty schema:spatial ;
+  owl:equivalentProperty schema:spatialCoverage ;
+.
+dcterms:temporal
+  owl:equivalentProperty schema:datasetTimeInterval ;
+  owl:equivalentProperty schema:temporal ;
```
agbeltran: [schema:temporal](http://schema.org/temporal) and [schema:datasetTimeInterval](http://schema.org/datasetTimeInterval) have both been superseded by [schema:temporalCoverage](http://schema.org/temporalCoverage), so I wonder if we should not include them in the mapping

dcat/rdf/schema.ttl
```diff
@@ -15,6 +15,42 @@
 @prefix skos: <http://www.w3.org/2004/02/skos/core#> .
 @prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
 
+<http://purl.org/dc/elements/1.1/title>
+  owl:equivalentProperty schema:name ;
+.
+dcterms:description
+  owl:equivalentProperty schema:description ;
+.
+dcterms:format
```
agbeltran: It seems to me that dcterms:format should be mapped to schema:fileFormat instead of schema:encodingFormat (that has domain schema:MediaObject)

agbeltran:
@dr-shorthair please see the comments I added in the form of a review

Adding the links to the external resources here, as they only appear in the commits at the moment:
- [Project Open Data](https://project-open-data.cio.gov/metadata-resources/)
- [SDWWG](https://www.w3.org/2015/spatial/wiki/ISO_19115_-_DCAT_-_Schema.org_mapping)

dcat/rdf/schema.ttl
dr-shorthair: I took that straight from https://www.w3.org/2015/spatial/wiki/ISO_19115_-_DCAT_-_Schema.org_mapping#Distribution 
At this stage I didn't challenge the prior work, I merely encoded it. 
However, I did have to apply some judgement about whether to use rdfs:subPropertyOf or owl:equivalentProperty

dcat/rdf/schema.ttl
dr-shorthair: Yes - I noticed that. For now I included everything I found in the prior work. The idea was to have a starting point to work from for a possibly normative alignment. I would also agree to dropping all reference to anything external that is already marked 'deprecated' but maybe we can do that in the next phase?

dr-shorthair:
Since @andrea-perego made important contributions to the prior work, would be good to have his eyes look over this.

andrea-perego:
Thanks, @dr-shorthair . I carried out a quick review, and I made a PR with small changes - see #159.

I need some more time to do a proper review, but I'm perfectly happy to have this PR merged.

I include below some preliminary considerations:
- The .ttl included an `owl:equivalentProperty` axiom for `dc:title`, but DCAT uses instead `dct:title`. So, in PR #159 I added the same axiom for `dct:title`, and also for `dc:description`. However, the question is whether we need to cover also the DC11 terms
- I'm not completely sure about the mappings of properties like `dcat:accessURL`. If I remember well (although I may be mistaken), the equivalent / analogous terms in schema.org have not as range a URI ref, but a string of type `xsd:anyURI`. If this is actually the case, should the mapping be different?

andrea-perego (approved):
Thanks, @dr-shorthair . I carried out a quick review, and I made a PR with small changes - see #159.

I need some more time to do a proper review, but I'm perfectly happy to have this PR merged.

I include below some preliminary considerations:
- The .ttl included an `owl:equivalentProperty` axiom for `dc:title`, but DCAT uses instead `dct:title`. So, in PR #159 I added the same axiom for `dct:title`, and also for `dc:description`. However, the question is whether we need to cover also the DC11 terms
- I'm not completely sure about the mappings of properties like `dcat:accessURL`. If I remember well (although I may be mistaken), the equivalent / analogous terms in schema.org have not as range a URI ref, but a string of type `xsd:anyURI`. If this is actually the case, should the mapping be different?

dr-shorthair:
Thanks Andrea. Ready for the other editors now.

davebrowning (approved):
Consistent with the current state, though it does highlight some issues/gaps (as per thread).  Provides us with a firm baseline work from/enhance to normative status as appropriate

PR #157, opened 2018-03-07
several update proposals to the narrative text

pwin:
... to make it more readable or make the sense clearer


agbeltran (approved):
Thanks, @pwin - the changes read very well.

PR #158, opened 2018-03-07
Added ack + changelog sections

andrea-perego:
Added placeholders for the following sections:
- Acknowledgments
- Change history - as per [ACTION-92](https://www.w3.org/2017/dxwg/track/actions/92)

Editorial changes concerning the case of some bib refs.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -34,7 +34,7 @@
 
 <p class="note">From DCAT 1.0</p>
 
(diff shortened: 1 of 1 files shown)


agbeltran (approved):
Thanks @andrea-perego - for the changes, do we need to provide a list in addition to pointing to the changes in github? Also, do we need to specify changes to the vocabulary file in addition to those changes to the document? In any case, I'll merge this changes now, and we can discuss about what else is needed later.

andrea-perego:
> Thanks @andrea-perego - for the changes, do we need to provide a list in addition to pointing to the changes in github? 

Yep. I added an empty list we can fill in.

> Also, do we need to specify changes to the vocabulary file in addition to those changes to the document? 

I don't think so. The reference document is the spec, not the vocabulary. Anyway, we can always integrate the changelog in the RDF file by using `vann:changes` - as done, e.g., here:

https://www.w3.org/ns/locn.ttl

> In any case, I'll merge this changes now, and we can discuss about what else is needed later.

Thanks!

andrea-perego:
@agbeltran , @dr-shorthair , [ACTION-92](https://www.w3.org/2017/dxwg/track/actions/59) is still pending review. 

Can it be closed?

PR #169, opened 2018-03-20
original DCAT had foaf:Page where it looks like foaf:page was intended.

dr-shorthair:
This looks like an erratum in the original DCAT.

### DIFF
--- a/dcat/rdf/dcat.ttl
+++ b/dcat/rdf/dcat.ttl
@@ -503,7 +503,7 @@ dcat:landingPage
   rdfs:label "صفحة وصول"@ar ;
   rdfs:label "ランディング・ページ"@ja ;
   rdfs:range foaf:Document ;
-  rdfs:subPropertyOf foaf:Page ;
+  rdfs:subPropertyOf foaf:page ;
 .
 dcat:mediaType
   rdf:type rdf:Property ;


dr-shorthair:
However, I wonder if an erratum should be issued for the old DCAT? 
@draggett @kcoyle @pwin what do you think?

draggett:
We could do that, but isn't our aim to replace the old DCAT spec with a new one, or do we continue use for both old and new specs, in which case errata for the old spec have a greater urgency?

akuckartz:
@draggett Publishing this erratum should not take much effort and certainly be possible before the new DCAT becomes a Recommendation. Or are there efforts necessary I am not aware of?

kcoyle:
I know nothing about errata in W3C specs, so I won't comment on that, but I do agree that 1) we should get word out sooner rather than later and 2) that we have had one discussion in which we rejected the idea of versioning the namespace, and I assume that we are not going to want to keep around a prior version of DCAT. The discussions in which Makx is promoting backward compatibility attest to the desire for a single, unified DCAT vocabulary.

To my mind the big issue is the RDF vocabulary file. Is there a way to mark an RDF term as "in error"? or "withdrawn"? Can the vocabulary be changed, not just the written documentation? If foaf:Page is withdrawn from DCAT, there could be current DCAT documents and DCAT-APs that will not validate against the upcoming version of DCAT. If we are lucky, the number of uses is not great and dataset owners can correct any existing uses, but there will need to be some coordination between the vocabulary change and DCAT documentation updates. In fact, it seems darned complicated!

draggett:
If DXWG agree on the errata, it should be simple to publish them so that they are easily found by people looking at the DCAT spec.

The replacement spec for DCAT would be able to obsolete RDF terms defined in the former specification. However, I am unaware of a machine interpretable means to mark RDF terms as withdrawn. This is an interesting use case for annotations for RDF. Other use cases relate to temporal, spatial, provenance, trust and data quality annotations. e.g. when a relation only holds since a given date. There is a potential connection to Property Graphs. I am at an early stage of planning for a W3C workshop on taking stock after two decades of RDF and linked data.

dr-shorthair:
The original DCAT https://www.w3.org/TR/2014/REC-vocab-dcat-20140116/#Property:dataset_landingpage  shows 
`dcat:landingPage rdfs:subPropertyOf foaf:page .` but the machine readable version of the schema https://www.w3.org/ns/dcat.ttl shows `dcat:landingPage rdfs:subPropertyOf foaf:Page .` **The bug is only in the RDF files**, not the document - but it is in both serializations https://www.w3.org/ns/dcat . **These should be fixed right away since this is clearly an unintentional error.** 

Moving forward we are making changes to DCAT but intend to use the same namespace. We will not remove any classes or properties, and are unlikely to deprecate any. There will be some small changes to axiomatization, but the resulting definitions will not change any of the essential semantics or intentions. We will provide a machine-readable artefact that will give people access to the original axiomatization if they prefer - currently it is this RDF graph https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat10.ttl but it might be reformulated some other way for publication (SPARQL, SHACL, something else ...). There are also likely to be some new classes and properties both in DCAT core and also in some optional extensions.

makxdekkers:
@kcoyle @draggett About 'withdrawing' RDF terms, there is already precedent of that in the DCAT [RDF schema](https://www.w3.org/ns/dcat.ttl). At the end of the schema file, there are three classes and five properties that were defined during development that did not make it into the Recommendation. These are all marked with:

`owl:deprecated true`

So that seems to be a machine-readable means to mark RDF terms as withdrawn.

kcoyle:
Thanks, Makx. We should do it that way, then.

dr-shorthair:
However, note that this issue is not concerned with withdrawing anything. It just corrects the object of a triple that is miscoded in the RDF (though correct in the Recommendation document).

andrea-perego:
@kcoyle said:

> Thanks, Makx. We should do it that way, then.

+1 from me. This is also the approach used in vCard:

https://www.w3.org/2006/vcard/ns#

PR #171, opened 2018-03-21
Relax global domain constraints on theme, keyword, landingPage

dr-shorthair:
Per resolutions in https://www.w3.org/2018/03/21-dxwgdcat-minutes


dr-shorthair:
Apologies - four images associated with a different thread got caught up in this commit.

agbeltran (approved):
The changes in the document and ttl files seem ok. I checked the Spanish and the changes in French seem fine too.

Issue #184, opened 2018-03-29
DCAT 1.1 URL in W3C space


andrea-perego:
The URL currently in the ED is: https://www.w3.org/TR/dcat-1-1/

I wonder whether, according to W3C naming conventions for vocabulary TRs, it should instead be:

https://www.w3.org/TR/vocab-dcat-1-1/

It would important to check this before we go for FPWD.

akuckartz:
Is that naming convention documented anywhere?

andrea-perego:
> Is that naming convention documented anywhere?

I don't know if it is documented somewhere. Just noted a consistent pattern (although there are exceptions):
- https://www.w3.org/TR/vocab-dcat/
- https://www.w3.org/TR/vocab-adms/
- http://www.w3.org/TR/vocab-org/
- https://www.w3.org/TR/vocab-dqv/
- https://www.w3.org/TR/vocab-duv/

philarcher:
No, it's not documented (IIRC) and not a hard and fast rule but it is the convention for this kind of thing as Andrea points out. Short URLs in /TR space need Director approval (see @draggett for details). Mind you, there's also https://www.w3.org/TR/odrl-vocab/ and others that way round. 

One thing I found surprising is that the RDF 1.1 specs have URls like https://www.w3.org/TR/rdf11-primer/ i.e. with 1.1 represented as 11. I take this as an indication that RDF 11.0 is unlikely to happen before the heat death of the universe.

dr-shorthair:
I've altered the config.js to match, but I'm not sure if the rec URI should change or not. Also, I just took a punt on 1.1 being the version - I think that is our intention, but if things got really wild it could go to 2.0 ... 

Clearly we will need some guidance from @draggett on these details prior to publication.

andrea-perego:
@dr-shorthair said:

> I've altered the config.js to match, but I'm not sure if the rec URI should change or not. Also, I just took a punt on 1.1 being the version - I think that is our intention, but if things got really wild it could go to 2.0 ...

Considering the discussion around broadening the scope of `dcat:Catalog` (if not of DCAT itself), probably we should seriously consider going for 2.0 (`vocab-dcat-2`).

dr-shorthair:
Yeah - you have a point. And dotty version signifiers are very prone to (mis-)interpretation as well. 
There are W3C precedents in several directions. 

HTML is at v5 - I think this was due to the progressive addition of features.
OWL went to v2 - I think this was because of incompatible changes. 
[...]

Issue #147, opened 2018-03-07
Prepare DCAT document for FPWD


dr-shorthair:
The DXWG is committed to publishing a FPWD by end of Q1 2018. 
Publication of a FPWD is primarily aimed at getting feedback from the community, so the document should make clear where changes to the draft are being contemplated. This may be done by embedding links to the Issues in GitHub.

dr-shorthair:
Intention is to send this to the dxwg plenary for release this week.

dr-shorthair:
Agreed by DCAT Team https://www.w3.org/2018/03/28-dxwgdcat-minutes#x07 
subject to 
1. addition of changelog by @andrea-perego 
2. minor editorial suggestions, accepted at the prerogative of document editors (@jpullmann to contribute)

andrea-perego:
Relevant PR created: #185

dr-shorthair:
Published: https://www.w3.org/TR/2018/WD-vocab-dcat-2-20180508/

PR #187, opened 2018-03-30
Refine the config.js

dr-shorthair:
Some tweaks in response to #184 
But not clear about the URIs for this version vs. the prior version. 
WIll need guidance from @draggett

### DIFF
--- a/dcat/config.js
+++ b/dcat/config.js
@@ -1,7 +1,7 @@
 var respecConfig = {
 //	preProcess: [dfn_index],
     specStatus: "ED",
-    shortName: "vocab-dcat-1-1",
+    shortName: "vocab-dcat-rev",
     canonicalURI: "TR",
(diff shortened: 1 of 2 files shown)


agbeltran:
When referring to the new DCAT version we might need to remove 1.1 and add 'DCAT revision' or similar. This should be modified after today's meeting.

dr-shorthair:
@agbeltran I believe this PR removes all mention of DCAT 1.0 and DCAT 1.1 including in the title. 
It consistently replaces them with 'DCAT 2014' and 'revision of DCAT' instead.

agbeltran:
Thanks, when referring to DCAT (the new version) should we say DCAT rev all along? I can merge and then we refine if needed.

agbeltran (approved):

dr-shorthair:
rE #147

PR #190, opened 2018-04-05
profiledesc

nicholascar:
Establishment of an independent folder for profiledesc work. This should be accepted regardless of how profile description work is ultimately conducted in the WG (aware I have a proposal not yet voted on).


pwin (approved):
first cut for profile description - necessary initial material

Issue #104, opened 2018-02-07
license and rights apply to Dataset as well as Distribution


dr-shorthair:
In DCAT 1.0 the dct properties license and rights are 'recommended for use' on dcat:Distribution, but not on dcat:Dataset. Suggest that licenses and rights can apply to datasets, and be inherited by distributions unless specifically overridden. 

A meta-issue here is that DCAT v1 does not use a formal mechanism to associate properties from external vocabularies (primarily DC) with DCAT classes, just an informal 'recommendation'. Should we consider OWL Restrictions to provide constraints on the use of specific properties in the context of the DCAT classes. See also #103

nicholascar:
perhaps "license and rights apply to Dataset not Distribution". The sense here being that the license etc. applies to the Dataset as delivered by the Distribution, not the Distribution itself. If a Distribution is anything other than a "Distribution", it is an action or method (of distributing the Dataset), not a thing, therefore can't have a license.

davebrowning:
There are examples today in the financial data exchange market of licences and rights/obligations varying according to mechanism and/or 'encoding' of a data set - so the licence would be different if the distribution pointed at a pdf file rather than an rdf or xml file even though the information is the same. [That's distinct from providing access by different means, if you see what I mean - which may again have different licence terms.  This links to issue #55 about distribution services, I think. ]

makxdekkers:
One of the objections against moving rights and licence 'upwards' from Distribution to Dataset is that there is no formal way to impose inheritance (as far as I know). If the properties are moved to Dataset (with the implicit meaning that the permissions and obligations apply to all Distributions), an application that receives DCAT metadata will need to look in two places: first look in the description of the Distribution, if nothing there, look in the description of the Dataset. Associating the information with the Dataset may gain some space (reducing duplication in descriptions of Distributions) but makes processing more complex.

andrea-perego:
Just to note that it may be worth reviewing the rationale behind the choice made by the GLD WG. As far as I can see, the relevant issue is:

https://www.w3.org/2011/gld/track/issues/53

PS: Actually, I think this review should be done programmatically for any revision to DCAT we are considering.

nicholascar:
I have reviewed the logic from the GLD WG - thanks Andrea! I note also the follow-up message from Makx that this was implemented in ADMS as well.

[...]

PR #193, opened 2018-04-06
Text for licences and rights

makxdekkers:
Text for licences and rights as proposed in issue 104  https://github.com/w3c/dxwg/issues/104#issuecomment-378243453


pwin (approved):
Following Action 84.  a much clearer statement following this change

PR #198, opened 2018-04-16
Profiledesc working

rob-metalinkage:
replace prof:resourceType with simeplified dct:conformsTo usage


aisaac:
Wondering what is the process for approving/merging this pull request?
I'm very much in favor of approving it, but I'm not sure who can do this in our profile work.

dr-shorthair:
See https://www.w3.org/2017/dxwg/wiki/GitHub_etiquette#Contributing_to_the_normative_deliverables 
Since Rob submitted, he shouldn't merge. That leaves Nick.

dr-shorthair:
However, it has not yet been established if the profile-description work item is a formal (normative) deliverable, and who the editors are. @kcoyle @pwin this needs to be resolved in plenary perhaps? If not normative, then there are no rules ...

aisaac:
@dr-shorthair I'm quite puzzled indeed. And the wiki page https://www.w3.org/2017/dxwg/wiki/GitHub_etiquette#Contributing_to_the_normative_deliverables does NOT list Lars, Rob and Ruben for another doc that they apparently edit (https://w3c.github.io/dxwg/profiles/). I'm going to try to send an email to the group list, maybe this PR is too specific for the discussion, as you hint.

nicholascar:
This profiledesc work is not yet accepted as DXWG work and will be formally considered at the F2F meeting.

larsgsvensson:
@dr-shorthair [wrote](https://github.com/w3c/dxwg/pull/198#issuecomment-382283526)
> However, it has not yet been established if the profile-description work item is a formal (normative) deliverable, and who the editors are. @kcoyle @pwin this needs to be resolved in plenary perhaps? If not normative, then there are no rules ...

I thought it's part of the __Guidance on publishing application profiles of vocabularies__ document in our [charter](https://www.w3.org/2017/dxwg/charter#deliverables)...

kcoyle:
I'll chat with Peter about this, but I don't think that this has been discussed by the plenary group and I must say that it isn't clear to me how it fits in to DCAT. There is indeed a deliverable regarding profiles and it is separate from DCAT itself. My opinion is that it is quite premature to consider merging this, and the disposition of this is indeed what we should establish at the f2f.

rob-metalinkage:
Updates merged in to reflect conversations. (Cardinalities of properties probably need to be relaxed)

Issue #237, opened 2018-05-17
Example of catalogued service


andrea-perego:
> As per [ACTION-117](https://www.w3.org/2017/dxwg/track/actions/117).

I include herebelow a full example of a metadata record for a service, as it is done in GeoDCAT-AP.

The example include:
1. The catalogue containing the service record
2. The record for the service
3. The corresponding catalogue record
4. The source ISO 19119 record, from which the catalogue record at point (3) has been derived

**NB:** `dcat:contactPoint` is used in the example following the revision in the current DCAT draft, which has relaxed the domain restriction of this property.

````turtle
# The catalog  containing the service record
a:Catalog a dcat:Catalog ;
  dcat:record :EEA-CSW-Endpoint-Record ;
  dcterms:hasPart :EEA-CSW-Endpoint .

# The GeoDCAT-AP catalog record about the EEA CSW public endpoint (a service)
  
:EEA-CSW-Endpoint-Record a dcat:CatalogRecord ;
  foaf:primaryTopic :EEA-CSW-Endpoint .
  dcterms:language <http://publications.europa.eu/resource/authority/language/ENG> ;
  dcterms:modified "2012-05-21"^^xsd:date ;
  dcat:contactPoint :EEA ;
  dcterms:identifier "4be5cb08-e082-426a-9c57-8a53d7cd3f65"^^xsd:string ;
  dcterms:conformsTo <http://data.europa.eu/930>
  dcterms:source :EEA-CSW-Endpoint-SourceRecord ;
  
# The ISO 19119 record from which the GeoDCAT-AP catalog record has been derived

:EEA-CSW-Endpoint-SourceRecord dcterms:conformsTo <http://www.isotc211.org/2005/gmd> .

# The EEA CSW public endpoint

:EEA-CSW-Endpoint a dctype:Service, dcat:Catalog ;
# The resource + service type, using the relevant ISO 19115/19 code lists
  dcterms:type <http://inspire.ec.europa.eu/metadata-codelist/ResourceType/service>, 
    <http://inspire.ec.europa.eu/metadata-codelist/SpatialDataServiceType/discovery> ;
  dcterms:title "European Environment Agency's public catalogue of spatial datasets."@en ;
[...]

Issue #181, opened 2018-03-29
Add mechanism to allow general services to be included in a dcat:Catalog


dr-shorthair:
Per the resolution of #172 it is necessary to create classes and/or properties to allow a general service to be included in a `dcat:Catalog`

dr-shorthair:
Resolved https://www.w3.org/2018/04/25-dxwgdcat-minutes
See https://github.com/w3c/dxwg/wiki/Cataloguing-data-services#chosen-solution

dr-shorthair:
Proposed RDF implementation in https://github.com/w3c/dxwg/blob/dcat-service-simon/dcat/rdf/dcat-service.ttl

dr-shorthair:
Close this as a separate issue - part of #172 and #241

Issue #180, opened 2018-03-29
Add mechanism for including data distribution services in a DCAT Catalogue


dr-shorthair:
Per resolution of #172 the following must be added to the DCAT vocabulary:

1. a class for data distribution services
2. a property to allow a data distribution service to be part of a catalog

dr-shorthair:
The mechanism used to resolve 2. will affect the resolution of #116 .

dr-shorthair:
See proposal here: https://github.com/w3c/dxwg/wiki/Cataloguing-data-services

rob-metalinkage:
IMHO Dataset and DataService look like subclasses of a generic CatalogEntry. If we solve this should also solve case where services are not bound to identified datasets - such as transformation services. Cant really put a counter-proposal until the two facilities needed are available - describing finer-grained service metadata in Distribution, and describing relationships between datasets.   Is a DataService just a subclass of Dataset that has relationships to one or more datasets that have a a service based distribution? Agnostic about final outcome but just feel we are modeling with half the story.

dr-shorthair:
I've tried to sketch a few of the options in the Wiki page 
https://github.com/w3c/dxwg/wiki/Cataloguing-data-services#alternative-views

Of course this is not exhaustive, but we need some concrete proposals in order to move forward.

dr-shorthair:
Resolved https://www.w3.org/2018/04/25-dxwgdcat-minutes 
See https://github.com/w3c/dxwg/wiki/Cataloguing-data-services#chosen-solution

dr-shorthair:
Proposed RDF implementation in https://github.com/w3c/dxwg/blob/dcat-service-simon/dcat/rdf/dcat-service.ttl

dr-shorthair:
Close this as a separate issue - part of #172 and #241

Issue #172, opened 2018-03-22
Clarify scope of DCAT - Datasets, Distributions, Services?


dr-shorthair:
DCAT version 1 supported cataloging of Datasets. Two of the Use Cases describe the need to also support cataloguing of Services: 

- [DCAT Distribution to describe web services - ID6](https://www.w3.org/TR/dcat-ucr/#ID6)
- [Modeling service-based data access - ID18](https://www.w3.org/TR/dcat-ucr/#ID18) 

There is evidence of use of complicated, indirect approaches to resolve this in existing deployments - see   https://github.com/w3c/dxwg/issues/116#issuecomment-374786075 https://github.com/w3c/dxwg/issues/166#issuecomment-373878749 

The following issues have discussed aspects of the problem and links between Datasets, Distributions and Distribution services in the context of a DCAT Catalog:  #56, #116, #124, #145, #166 
Some of these discuss potential solutions before the requirements have been clearly articulated and agreed. 

This issue is to clarify the scope of DCAT. When this has been clarified we can then determine the best way to meet the requirements, through creative use or adaptation of the existing vocabularies, or by adding some classes and properties to DCAT, or some other method.

dr-shorthair:
Possible motions for this week's DCAT meeting:
1. Proposed: DCAT Catalogs may include data services
and if this is agreed
2. Proposed: A class for data services should be added to DCAT backbone, alongside `dcat:Dataset` and `dcat:Catalog`

andrea-perego:
Thanks for drafting this proposal, @dr-shorthair .

I have a couple of comments:

1. I'm not sure `dcat:Catalog` should be made a subclass of `dctype:Service`. [Its definition in DCAT 1.0](https://www.w3.org/TR/vocab-dcat/#class-catalog) says "A data catalog is a curated collection of metadata about datasets." So, although it _may_ correspond to a catalogue service (as done in GeoDCAT-AP), it can simply be a "static" collection, a mechanism for grouping datasets based on some criteria. I'm aware of examples of this use - e.g., the list of datasets produced by a project/activity.
[...]

PR #241, opened 2018-05-25
Extending DCAT for services

dr-shorthair:
This PR implements the model shown in https://github.com/w3c/dxwg/blob/dcat-service-simon/dcat/UML/DCAT-summary.png as agreed in meeting of DCAT sub-group meeting 2018-05-25
https://www.w3.org/2018/05/24-dxwgdcat-minutes.html#ResolutionSummary 

4 classes are added: 
- `Resource`
- `DataService`
- `DistributionService`
- `DiscoveryService`

6 properties are added: 
- `catalog`
- `service`
- `accessService` (discussion at #124)
- `servesDataset`
- `endpointLocation` (discussion at #237)
- `endpointDescription` (discussion at #237)

This extension to DCAT is intended to satisfy requirement #56 and the matters discussed in #172 #180 #181, as illustrated in #237 

The text labels and definitions might need some tweaking as discussed, and also the global (domain/range) and local (Restriction) constraints. 

The intention of this pull request is to get the backbone into the RDF representation and into the draft recommendation document, so that we can refer to it while discussing other issues, and to allow us to refine other aspects _in context_ in due course.


dr-shorthair:
@andrea-perego will you be able to review this?

pwin:
Shouldn't endpointLocation be endpointURL as described by @andrea-perego in #237 ?
"location" suggests geospatial

andrea-perego (commented):
Together with @pwin , I would also prefer `dcat:endpointURL`, as reminds `dcat:accessURL` and makes it explicit the it should take as object a URI/URL and not a class instance.

I would also support the inclusion of a generic "Service" class (either by re-using `dctype:Service` or by defining a new class `dcat:Service`).

Said that, I'm fine with merging this PR and address these points later - we just need to create the relevant issues.

dr-shorthair:
1. I'm OK calling it `endpointURL` _provided_ that for all the properties `*URL`, they are corrected to owl:DatatypeProperty with range xsd:anyURI

2. Service was renamed DataService following the discussion in the F2F - @kcoyle seemed to feel quite strongly about that, and also with @danbri in the room we wanted to be clear that we were talking about web- and data-oriented services not hairdressers.

andrea-perego:
@dr-shorthair said:

> 1. I'm OK calling it `endpointURL` _provided_ that for all the properties `*URL`, they are corrected to owl:DatatypeProperty with range xsd:anyURI

I have to disagree here. It would break existing implementations, and I don't think we have any use cases supporting this revision. Moreover, having them as object properties would enable cross-linking of related resources (as services and distributions). 

Using "URL" in the property name is consistent with similar DCAT properties (`dcat:accessURL` and `dcat:downloadURL`, which are object properties), and DCAT users will recognise the pattern. So, I don't see the need of requiring a term to be a datatype property just because it includes "URL" in its name.

> 2. Service was renamed DataService following the discussion in the F2F - @kcoyle seemed to feel quite strongly about that, and also with @danbri in the room we wanted to be clear that we were talking about web- and data-oriented services not hairdressers. 

I tend to agree that we should not allow any type of service, but I don't think we have currently a clear idea of what should be in or out. 

An option, already discussed, and recently suggested by @pwin [in his mail](http://lists.w3.org/Archives/Public/public-dxwg-wg/2018Jun/0058.html), is to say that the scope is limited to online (Web?) services. On the other hand, we have the example of the Core Public Service vocabulary, which is used to describe a (not necessarily online) service delivered by a public institution, and such descriptions are meant to be available in catalogues. I cannot find a strong case to say that we cannot use DCAT for describing these catalogues.

dr-shorthair:
1. I opened a separate issue to discuss the type and range of *URL properties #249 - let's have the conversation there. In this PR there are no changes to `downloadURL` or `accessURL`. 

2. `DataService` is probably the most conservative option for naming. Can we accept it for now and consider again later if more general services also need to be included?

Note I also moved all the classes and properties that were already marked 'deprecated' in DCAT-2014 into the RDF document [dcat2014.ttl](https://github.com/w3c/dxwg/blob/dcat-service-simon/dcat/rdf/dcat2014.ttl), where we are gathering all the axioms removed in the DCAT-rev process. That allows the dcat.ttl document to reflect the current state without any noise (most of the deprecated classes were related to services ...). They can be reinstated later if necessary. 

I'm concerned that merging becomes difficult if we have too many branches open. This one covers what is likely to be the main refactor that we will make in this DCAT revision. It introduces the class `dcat:Resource` interpolated above `dcat:Dataset`, and also includes `dcat:DataService` and sub-classes, and a small number of new properties - `dcat:service`, `dcat:catalog`, `dcat:endpointURL`, `dcat:endpointDescription`, `dcat:accessService` and  `dcat:servesDataset`. The overall shape of these has been accepted by the team, with the explicit understanding that we can and should tidy up details later. My view is that we need to move this material into the gh-branch now so that we can get community scrutiny of the editor's draft. Nothing is final until we publish a rec in a year or so, but it would be helpful to expose this material.

dr-shorthair:
Just spent another hour resolving conflicts through the various branches and merges. 
I believe I have addressed all @andrea-perego comments now, and this should be a good basis for moving forward. Can we merge this into the gh-pages i.e. _Editor's Draft_. 
I don't think the risks are great to expose this now - all details are still subject to potential revision, but the general shape reflects the agreement in Genova and the following telecon. 
@davebrowning @pwin @agbeltran ?

agbeltran (approved):
I have gone through the changes and they seem ok to me and in line with the discussions. I will merge now - as always, the understanding is that more revisions could arise from further discussions.

PR #243, opened 2018-05-30
Remove one instance of "government"

nichtich:
DCAT is not only for government data catalogs, so "government" should only be used as example. At this section the word should be removed


dr-shorthair:
Thanks @nichtich - that is an obvious update.

Issue #254, opened 2018-06-11
DCAT - LDP (Linked Data Platform) alignment


jakubklimek:
DCAT is a W3C Recommendation, and RDF vocabulary describing mainly collections (of datasets, catalog records, distributions, etc.)
There should be a clear alignment including examples with the [Linked Data Platform 1.0 W3C Recommendation](https://www.w3.org/TR/ldp/) describing modeling of generic containers and their members.

For instance, `dcat:Dataset` could also be (could also have) a `ldp:DirectContainer` with `ldp:hasMemberRelation` `dcat:distribution` and similarly for `dcat:Catalog`, `dcat:CatalogRecord` and maybe others.

Clearing this alignment up would improve interoperability e.g. among potential [Solid](https://solid.mit.edu/) based data catalogs, etc.

dr-shorthair:
Hey @jakubklimek - I admire your ambition to grandly unify W3C's thinking here. 
Mind you, to get this into the normative content there will have to be a stronger use-case than mere (!) consistency ... 

My current expectation is that DCAT will mostly be stand-alone (with a lot of DC and some references to ODRL, and possibly ADMS). The alignments (many!) will probably be 'informative' annexes and graphs. 

I started a couple here: https://github.com/w3c/dxwg/tree/gh-pages/dcat/rdf 
Perhaps you could formulate DCAT-LDP as an RDF graph to be dropped in there?

jakubklimek:
@dr-shorthair The use case is any Solid based data catalog application (admittedly non-existent at the moment).

I am not sure yet that there will be an RDF alignment graph. Maybe a set of examples illustrating the alignment could be enough. That is up for discussion.

Now I think that there should even be examples of interoperability with the recent [Social Web Protocols](https://www.w3.org/TR/social-web-protocols/) so that it is clear 

1. That we know about them and take them into account
2. That they can be used together with DCAT
3. How they can be used together with DCAT to improve interoperability across implementations:

They include:
- [Linked Data Platform 1.0](https://www.w3.org/TR/ldp/) and [Linked Data Platform Best Practices and Guidelines](https://www.w3.org/TR/ldp-bp/) as support for read-write applications using the DCAT data model
- [Linked Data Platform Paging 1.0](https://www.w3.org/TR/ldp-paging/) for catalogs with many records
- [ActivityPub](https://www.w3.org/TR/activitypub/) for interaction with distributed social networks
[...]

PR #257, opened 2018-06-20
Add description of relation to LDP and LDN

jakubklimek:
Added a section with examples and property definitions describing relation of DCAT to LDP and LDN W3C Recommendations. This is in relation to #254.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1773,6 +1773,207 @@ If a ISO 639-1 (two-letter) code is defined for language, then its corresponding
 
 </section>
 
+<section id="other-w3c-recommendations">
+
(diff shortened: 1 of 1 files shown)


davebrowning (approved):
Content of this looks good from a DXWG angle (not that I know much about LDP), and as requested its in its discrete section pending agreement on normative vs informative (So issue #254 stays open)

dr-shorthair:
@davebrowning I note that @jakubklimek has used the HTML class "example" to get the examples nicely formatted in a yellow box. Would be good to do this uniformly through the document? Is this something you can push through for 2PWD?

PR #260, opened 2018-06-22
further PROV alignment, including for mooted WebService class

nicholascar:


dcat/rdf/prov.ttl
```diff
@@ -1,56 +1,81 @@
...
+  rdfs:comment "prov:Agent is modeled as a subclass of foaf:Agent since it is the subset of agents that are involved in prov:Activity individuals" ;
+  rdfs:subClassOf foaf:Agent ;
+.
+<https://www.w3.org/2017/dxwg/wiki/Main_Page>
+  rdf:type foaf:Agent ;
+  rdfs:label "W3C Data Exchange Working Group" ;
+.
+
+dcmitype:Dataset
+  rdfs:subClassOf prov:Entity ;
```
azaroth42: I don't think it's kosher to make this assertion about a resource outside of the control of the WG.

azaroth42:
I don't think it's a good practice to make ontological assertions about resources outside of the control of the WG if the document is to be on the rec track. For example, asserting that classes, other than dcat ones, have new subclasses have very far reaching side effects.  Especially if you take into account the axioms of PROV.

nicholascar:
Then suggest a better mechanism for DCAT + other ont alignment since there is a great desire for such mappings.

We have similar mappings delivered from WG to non-WG models by the Spatial Data on the Web WG recently too.

azaroth42:
How about just human readable documentation as guidelines to manage translation between ontologies, or reducing the ontological implications by using skos:closeMatch or similar rather than subClassing or sameAs? By publishing a normative document (again, assuming that this is rec track!) W3C would be saying that the DCMI's classes have new subclasses, and that conforming systems would take this into account. 

I don't believe that you can adequately provide test coverage for this to get through PR.

dr-shorthair:
1. @nicholascar There is no mooted WebService class - the original (pre-W3C) DCAT has a WebService class which was marked 'deprecated' in DCAT-2014. There is a mooted DataService class in the [current DCAT ED](https://w3c.github.io/dxwg/dcat/#class-dataservice)

2. @azaroth42 I don't understand the problem with sub-classing. Surely that is one of the most common ways to use existing ontologies - i.e. to assert that some concept of interest in a particular application is a subset of one or more classes defined in some more general vocabulary? If we want to use PROV-O capabilities to provide richer descriptions of DCAT things, then we can just say "we consider a dcat:Dataset (etc) to be a kind of prov:Entity, since it has a production history that can be conveniently described using PROV-O ...". Yes, there are risks, but aren't they entirely on the side of the dependent ontology?

dr-shorthair:
1. Alignment needs to be brought up to date with re-factored type hierarchy - `dcat:Resource` is the top
2. The relationships of `dcat:Dataset` with `dctype:Dataset` was discussed in #98 and #64 - the key point is that some of the _siblings_ to `dctype:Dataset` in the DCTYPE set are also plausible values for the `dct:type` of a `dcat:Dataset`. So we should not make a strong assertion that every `dcat:Dataset` is a `dctype:Dataset` . 

See https://github.com/w3c/dxwg/commit/b0b43956e325fa9b38d1acc2340dd4efea4fb30b#diff-7afc9d1f89b949835fd56baf582c948e and https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat.ttl#L287

dr-shorthair:
1. DCAT-PROV mapping has been renamed to [dcat-prov.ttl](https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-prov.ttl) 

2. @azaroth42 wrote "W3C would be saying that the DCMI's classes have new subclasses" - I'm a bit confused about what you are referring to here. The mapping does not mention DCMI classes. 

The intention of publishing a mapping as a separate document/graph is precisely to allow interested parties to only load it if they want to use it, and isolate it from the normative RDF description of the DCAT vocabulary so it is not being forced on anyone. Am I missing something?

dr-shorthair:
Ah - I see that @nicholascar sneaked a reference to DCMI into his proposed change. 

This particular PR is now stale, so I suggest we close it and revisit the PROV alignment in separate issues.

PR #291, opened 2018-06-28
Remove minor nits left behind in copy-and-paste 

dr-shorthair:
See individual commits for details

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -71,11 +71,10 @@
 
 <p>A data catalog conforms to DCAT if:</p>
 <ul>
-	<li> Access to data is organized into datasets, distributions, and data-distribution-services. </li>
-	<li> Access to other resources is organized into suitable sub-classes of catalogued-item. </li>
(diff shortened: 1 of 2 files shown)


agbeltran (commented):
Has the skos import been removed?

dr-shorthair:
@agbeltran another nit - fixed now.

dcat/index.html
```diff
@@ -1330,7 +1329,6 @@ <h3>Class: Catalogued resource</h3>
 <a href="#Property:resource_conformsto">conformsTo</a>,
 <a href="#Property:resource_contactPoint">contact point</a>,
 <a href="#Property:resource_description">description</a>,
-<a href="#Property:resource_distribution">distribution</a>,
```
agbeltran, line 1331: @dr-shorthair just double-checking, the dat:distribution property is removed from here because it only applies to datasets and not catalogued resources, correct?

dr-shorthair: Correct. It was one of the cut-and-paste debris.

agbeltran: ok, thanks

PR #298, opened 2018-07-10
SHOULD/MUST/MAY/NOT

dr-shorthair:
Adjust presentation and use of SHOULD/MUST/MAY to be consistent throughout the document


agbeltran (approved):
I made a small change fixing MUST NOT usage and capitalising DCAT. Now ready to merge.

PR #316, opened 2018-08-22
renaming profile folders

nicholascar:
This renaming is intended to order the profiling working documents in GitHub in line with the Profile Guidance, Profile Negotiation via HTTP & Profile Description Ontology.


nicholascar:
cancelling this for a different structure

PR #318, opened 2018-08-22
updated profile components' naming

nicholascar:
Updating the naming of two folders (profiles -> profile-guidance & profiledesc -> profiledescont) and addition of another folder (profileneg-by-api) to reorder profile guidance work as per generally agreed structure.


kcoyle:
Actually, I see no reason to rename them. In any case, a change of this magnitude should go through the plenary. Can we just leave them as they are?

nicholascar:
The purpose of the propsed renaming was to ensure that the scope of work int he folders was better understood. 'profiledesc' could be about profile descriptions generally so I wanted to change that to 'profiledescont' to ensure it was understood it was a specific thing - an ontology - and not general work. So too with 'profiles' which could either be work about profiles or a set of them. by calling that folder 'profile-guidance' it would be clear it was guidance, not a list of profiles.

kcoyle:
I don't think that anyone has problems with the folder names. You could ask the group. I personally prefer shorter names than longer ones. And we're used to the ones we have.

PR #321, opened 2018-08-22
Notes on data citation and addition of creator property for dataset

agbeltran:
Note that I also re-ordered classes, moving dcat:Resource up in the document.


dr-shorthair:
Suggest adding use of dct:creator to one or more of the examples in Section 5 "Overview"

dr-shorthair (approved):

Issue #300, opened 2018-07-13
Provide guidance discouraging usage of blank nodes


jakubklimek:
Based on the discussion in #253 a need for guidance discouraging usage of blank nodes was identified.

Even though it is one of the principles of Linked Data (to identify everything with an IRI), publishers tend to use blank nodes for various types of entities including some very important ones like `dcat:Distribution`s.

I suggest:

1. Not to use blank nodes in any examples in the specification. A similar effect (shortening) can be achieved by a combination of prefixes and relative IRIs.
2. Recommend explicitly to identify everything with an IRI and support it with a link to [this Linked Data page](https://www.w3.org/DesignIssues/LinkedData.html). This should be placed somewhere in the beginning of the specification.
3. Explain the disadvantages of using blank nodes - impossible to link, different identifiers with each access

riccardoAlbertoni:
@jakubklimek  wrote
> Explain the disadvantages of using blank nodes - impossible to link, different identifiers with each access

I guess the following sentence taken from the book [Leigh Dodds,
Ian Davis; Linked Data Patterns A pattern catalogue for modelling, publishing, and consuming Linked Data, chapter 2](http://patterns.dataincubator.org/book/identifier-patterns.html) might help.

> RDF allows resources to have global identifiers or to be un-named "blank nodes". While blank nodes may offer flexibility for some use cases, in a Linked Data context blank nodes limit our ability to collaboratively annotate data. A blank node cannot be the target of a link and we can't annotate it with new information from new sources. As one of the biggest benefits of the Linked Data approach is that "anyone can say anything anywhere", use of blank nodes undermines some of the advantages we can gain from wide adoption of the RDF model. Even within the closed world of a single application dataset, use of blank nodes can quickly become limiting when integrating new data.

jakubklimek:
@riccardoAlbertoni That is a good reference. I would still look for a stronger discouragement in the specification. Maybe backed by this reference, I would say something like "Since DCAT is meant for description of entities with high linking potential such as not only catalogs, datasets, distributions and services, but also contact information and other useful metadata, the usage of blank nodes is strongly discouraged".

agbeltran:
+1 to adding some text in line with @jakubklimek's suggestion after including the definition of a blank node and citing the article pointed out by @riccardoAlbertoni

davebrowning:
[...]

PR #322, opened 2018-08-23
Remove blank nodes from examples

dr-shorthair:
In response to #300


davebrowning (approved):
All looks good - there was one lingering 'carat' problem (at line 351) which I've presumptively fixed

Issue #320, opened 2018-08-22
Links from DCAT draft to (DCAT 2014) rdf 


davebrowning:
Looking at the editors draft, some (possibly all)  of the internal links in the vocabulary specification pick up old RDF.  Its most obvious where the RDF Property field contains something like `dcat:dataset` which takes you to what's basically a redirect page at https://www.w3.org/ns/dcat that then links to the 2014 ttl or rdf/xml...so you're no longer looking at the updated version.  By comparison, links to definitions in specifications like DCT take you from `dct:hasPart` direct to the ttl (even if the anchor syntax is a bit irrelevant). 

[There's a separate consistency issue - which is what I was trying to address - where we've mixed up links to the rdf with links within the document but that's comparatively minor - I'm fixing that on the back of existing actions]

I think there are two questions:

1. Do we want to persist the 2014 approach of an intermediate page which allows then points independently to the RDF?  That only makes sense to me if we plan to do both ttl and rdf/xml forms - I think we decided not to do that?
2. For Second public draft, what do we want to achieve?  We do have a link to the live version of the ttl in Abstract - which is good, but I don't think its adequate.  [We know nobody reads introductions, but people do read abstracts, I think.  Unfortunately its not always the same people who read the doc.....]

My instinct is that we should fix this so the links work but open to be persuaded...

pwin:
Re: Q1.  I don't see much point in RDF/XML if we have ttl.  I think that a diagram of the examples (just simple circles and arrows) could also be helpful if we are going to do any duplication of the representation of examples.  What do you think?

davebrowning:
While following this up via the published [W3C policy on namespaces](https://www.w3.org/2005/07/13-nsuri), what we have at https://www.w3.org/ns/dcat is an abbreviated namespace document as described in the policy.  Looking for some current practice (for example for [XML](https://www.w3.org/XML/1998/namespace), [XSD](https://www.w3.org/2001/XMLSchema) or [Prov-O](https://www.w3.org/ns/prov#)) the contents of these appear to be fairly flexible though the DCAT offering is a little abbreviated.

For 2PWD purposes, I propose we update the existing DCAT namespace document to add a simple link to the updated dcat ttl file - there is perhaps a broader discussion needed on what we want to do as we get closer to proposed rec etc - now raised as #377

davebrowning:
This issue which was raised in the context of 2PWD is actually resolved.  The longer term/REC approach should be addressed under #377, so I'm closing this one.

Issue #301, opened 2018-07-13
Define and apply rules for definitions in DCAT


makxdekkers:
I would like to propose that we define and apply consistent rules for definitions in DCAT.
Some issues I have encountered (looking at the 11 July 2018 draft at https://w3c.github.io/dxwg/dcat/):

1. Some definitions are circular in the sense that they use the same word in the label and definition. There are cases where there is no chance for people to misunderstand – e.g. the definition of `dcat:mediaType`: “media type .. as defined by IANA” is probably OK, but `dcat:service` as “A service …” is maybe not so useful. Compare the definition of `dcat:DataService` (“A service …”) with `dcat:DataDistributionService` (“A site or end-point … “). The latter is less circular.
My suggestion is to avoid circularity where it can be avoided. Less circularity may also help translators.

2. Some definitions from external vocabularies do not use the same definition as in the source specification, e.g. `dct:license`. 
There I would suggest that the DCAT specification reuse the original definition in the Definition field and add the specifics for use in DCAT in a Usage note.

3. Some definitions of properties use a template “A _some characteristic_ (_further clarification_)” that states what the object of the property is (e.g. `dcat:dataset`, `dcat:mediaType`); others state that it is a link to something (e.g. `dcat:catalog`) or even a link to a description of something (e.g.` dcat:servesDataset`). Furthermore, the property `dcat:distribution` does not say what it is but what it does “Connect a dataset …” A minor note on the use of “The ..” in the definition: this seems to indicate that a `dcat:Catalog` can have only one homepage and use only one KOS for classification, which I think is not necessarily true. 
My suggestion on this point is to standardise on the template “A _some characteristic_ (_further clarification_)”.  

4. I would also suggest that properties that have a URL (Literal) as object should make that explicit, e.g. `dcat:downloadURL`: “The URL of the downloadable file in an given format”.

I think it would be useful to enhance the consistency of the specification before publication of the next Public Working Draft.

davebrowning:
These all look like good ideas.  I'm happy to work this through into the document under my current editorial sweep (in fact, I've already started)

dr-shorthair:
And in particular, we need to ensure that the text in the recommendation document matches the text in the RDF representation.

And per #174 there will need to be a subsequent sweep to align the non-English translations.

davebrowning:
PR #332 (now merged) has aligned the document text with the above rules (E&OE).  RDF (en) presentation pending.

davebrowning:
[...]

PR #332, opened 2018-09-03
Sub-editing/Issue 301

davebrowning:
This set of changes aims to address the textual rules raised in issue #301.  Along the way, there has been some clarifing textual changes ( in particular where the addition of services to catalogs hadn't been acknowledged) and a few broken internal links.  There's a fair few small changes (There was also an indentation problem which has increased the line count....)

Some of these updates need to be reflected in the rdf (turtle) file but since some of the links to that don't work (see #320 ) yet and we need to address how that works for 2PWD, I'd like to ensure the changes are okay then I can cut and paste into the rdf and ensure the links work as needed

Questions/comments:

1. The document still repeats some properties in (e.g.  `title` is in `dcat:Catalog` etc as well as `dcat:Resource`.  In general the definitions are now the same. Are we happy to define the properties once in the super class and make that change now?

2. In general internal links to class definitions don't work in the editors draft - it looks  to me that the respec process fixes that, but I'm following that up to make sure it's done correctly.


dr-shorthair (approved):

dr-shorthair:
@davebrowning I notice that many property descriptions in the rec include text in the 'Range:' section - e.g. https://w3c.github.io/dxwg/dcat/#Property:catalog_language - which has not made it into the RDF representation (i.e. the ttl file). This should probably be included as `skos:scopeNote` elements. In many cases these are on dct properties, which may not be currently described in the RDF representation. But you can add local scope notes which would thus supplement the dct descriptions pulled in from DCMI. 

I have proposed something similar to add the scope notes for `prov:wasGeneratedBy` into the RDF - see lines 790-791 in https://github.com/w3c/dxwg/blob/dcat-issue71-simon/dcat/rdf/dcat.ttl

dr-shorthair:
On this issue: 

> The document still repeats some properties in (e.g. title is in dcat:Catalog etc as well as dcat:Resource. In general the definitions are now the same. Are we happy to define the properties once in the super class and make that change now?

Since the association of many (most?) properties with classes is not axiomatized, but rather is just through normative text in the rec document, I wonder if we should copy down the list of inherited properties, instead of just saying 'In addition to the properties inherited from the super-class ...' - the links could still point back to the original documentation on the super-class, but it might be clearer to implementers what is expected.

davebrowning:
Good point on the range: sections on property descriptions - that's still outstanding (that is, I've fixed the English text in the rdf, but not addressed the ranges.)

On repetition: yes, that might work.  I'm against having inherited property definitions duplicated in their entirety, but listing them in the section as part of/immediately below the Class definition would be helpful.  [Since I've been flipping between the 2014 spec and our work, the usefulness of having all the recommended properties for a class listed together stands out a bit....]

dr-shorthair:
This updated version of the UML shows the inherited properties more clearly: 
https://raw.githubusercontent.com/w3c/dxwg/dcat-issue347-simon/dcat/UML/DCAT-summary-all-attributes.png 
 in #348

davebrowning:
To be a bit more explicit - yes, that's a good step forward

PR #334, opened 2018-09-04
moved stub text to ReSpec Issues

nicholascar:
As requested in last PR. And a few other, small, updates: brings this doc into line with the Profile Guidance doc.


larsgsvensson (approved):
We'll need to clean up the IETF I-D references (currently the work by Ruben, Herbert and me isn't a proper I-D since we haven't submitted it yet...), but IMHO that can be done before going to FPWD.

Issue #71, opened 2018-01-18
Project context [RPCX]


jpullmann:
<section class='requirement' id='RPCX'><h3>Project context [<a href='https://www.w3.org/TR/dcat-ucr#RPCX' target='_blank'>RPCX</a>]</h3><p>Provide a means to define a 'project' as a research, funding or work organzation context of a dataset.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID49' target='_blank'>Dataset business context [ID49]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID31' target='_blank'>Modeling funding sources [ID31]</a> </section>

dr-shorthair:
I have a draft proposal for a small vocabulary for Project, as a subclass of `prov:Activity`. 
See 
https://dr-shorthair.github.io/ont/project/ 

This would support linking a dataset to a project using the `prov:wasGeneratedBy` predicate, as mentioned in #77

andrea-perego:
Thanks, @dr-shorthair . I would also like to contribute some work we did to map DataCite to DCAT-AP, which includes the mapping of what in DataCite is called "Funding Reference". The mapping tables are here: https://ec-jrc.github.io/datacite-to-dcat-ap/

I think this is also one of the possible use cases for the use of qualified and non-qualified forms.
The non-qualified form is the basic case where you want to say that a given dataset has been created by a given project. However, if you also need to say that this is done in a given timeframe of the activity of a project you need to add a node in the graph, between "project" and "dataset", where to attach this information. 

@dr-shorthair , in my understanding the vocabulary you contributed support both cases, right?

dr-shorthair:
Linking through to funding details is part of the proposed Project ontology. Not sure if I got it all right yet, so would be interested in working through other examples.

dr-shorthair:
Suggest removing the following labels: 
profile, semantics, service, usage_control, version
??

dr-shorthair:
Picking up the [second example on #253](https://github.com/w3c/dxwg/issues/253#issuecomment-403276218) which describes a [dataset from CSIRO's DAP](https://data.csiro.au/dap/landingpage?pid=csiro:P366-2003SEPT), the following uses PROV to document the project context for the dataset.  The `PROV-O` property `prov:wasGeneratedBy` points to `dap:P366` which is a `prov:Activity`, which in turn is associated with `dap:ATNF` and used the `dap:Parkes-radio-telescope`. 

```
[...]

PR #338, opened 2018-09-06
Clarify that the activity that generated a dataset might be on-going

dr-shorthair:
Response to https://github.com/w3c/dxwg/issues/71#issuecomment-419075177


dr-shorthair:
See changes rendered in context here 
https://rawgit.com/w3c/dxwg/dcat-issue71-simon/dcat/index.html#examples-dataset-provenance

davebrowning (approved):
This looks like an improvement to the example to me.  It feels proportionate to the use case/need without getting bogged down in the details relevent to PROV-O

PR #339, opened 2018-09-07
action-200

nicholascar:
Dear DCAT editors,

As per [ACTION-200](https://www.w3.org/2017/dxwg/track/actions/200) from the latest Profile Guidance subgroup meeting, please could you include this "Family of Documents" section in the DCAT spec? This follows PROV's use of such a section where a WG produces a large number of outputs. The Profile Guidance and Conneg groups look forward to any suggestions about presentation of this Family of Documents section that you might have.

Thanks,

Nick (and Profile Guidance people)

### DIFF
--- a/dcat/config.js
+++ b/dcat/config.js
@@ -122,6 +122,28 @@ var respecConfig = {
           authors : [ "Leigh Dodds", "Ian Davis" ],
          date: "31 May 2012",
           href : "http://patterns.dataincubator.org/book/"
+      },
+      "PROF-CONNEG": {
+            href: "https://www.w3.org/TR/conneg-by-ap/",
+            title: "Content Negotiation by Profile",
+            date: " 2018-12-31",
+            status: "W3C Editor's Draft"
+      },
+      "PROF-GUIDE": {
+            href: "https://www.w3.org/TR/profile-guidance/",
+            title: "Profile Guidance",
+            date: " 2018-12-31",
+            status: "W3C Editor's Draft"
+      },
+      "PROF-IETF": {
+            authors: [
+                "L. Svensson",
+                "R. Verborgh"
+            ],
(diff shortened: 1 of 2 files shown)


dr-shorthair:
So this is how it would look when rendered: 
https://rawgit.com/w3c/dxwg/dcat-fam-of-docs/dcat/index.html#FamilyOfDocs 

I see the value in providing a document map. However, I'm a little wary of making this a section in the main body. Looking at precedents: 
- OWL2 has a whole separate [overview document](https://www.w3.org/TR/owl2-overview/#Documentation_Roadmap) with the roadmap (reasonably) as a first-class section. 
- [PROV-O](https://www.w3.org/TR/prov-o/) includes a guide to the PROV family of documents in its status section. 

I think the latter might be a better model.

dr-shorthair:
Also - can you confirm that the composition of this set of documents has now been agreed in plenary?

nicholascar:
"Also - can you confirm that the composition of this set of documents has now been agreed in plenary?"

So far, the documents listed are just per-deliverable so yes, these are approved. We may grow the list (e.g. if profileDesc became a document).

dr-shorthair (changes requested):
Suggest moving the description of the doc family into the [Status] section at the top

nicholascar:
> move to status section

I can't mechanically shift the section into status (where I wanted it from the beginning) since it interferes with ReSpec building. If I (or anyone) can work out how to move it in there, I'd be thrilled.

Ignore above: I've managed to do it! Not sure what my previous problem was. Next commit effects this change, see https://rawgit.com/w3c/dxwg/dcat-fam-of-docs/dcat/index.html#FamilyOfDocs

dr-shorthair:
See my subsequent changes. 
Alternatively, you can create another pre-ToC section with `class="introductory"`

dr-shorthair (approved):

davebrowning (approved):
Looks fine to me

PR #345, opened 2018-09-12
Examples of data services

dr-shorthair:
Added some examples of services:
1. a discovery service at EEA - corresponding the example provided by @andrea-perego in #237 
2. a map service at GA

See preview at https://rawgit.com/w3c/dxwg/dcat-issue237-simon/dcat/index.html#data-service-examples 

Maybe add to Milestone 2pwd? 

And still need a fully coupled example of `DataDistributionService` linked to `Datasets`.


agbeltran (changes requested):
It would be good if the document points to the ttl files too.

dr-shorthair:
You mean the ones that I dropped into the ['examples' directory](https://github.com/w3c/dxwg/tree/gh-pages/dcat/examples). 
Perhaps a note next to the one about use of Turtle syntax providing a global pointer to the GitHub directory of DCAT examples? Then maybe also a pointer from each individual example to the document that contains it (actually we don't have them all ...)

agbeltran:
Yes, exactly. As I saw the ttl files in the PR, I think it would be useful to advertise their existence through the document. We could (progressively) add the other ones missing.

davebrowning:
Can we discuss how/what we link in today's meeting?  [I agree with the sentiment, by the way - its more the solution].

dr-shorthair:
I've added a note and some links - strawman. See https://rawgit.com/w3c/dxwg/dcat-issue237-simon/dcat/index.html#data-service-examples (the links in this rendering are broken as the code is in a branch not yet merged). 

But we might want to create a CSS class to make these links stand out (similar to class="note" and class="issue").

agbeltran:
Agree on discussing today - I had already included it on the agenda.

agbeltran (approved):

Issue #251, opened 2018-06-07
DCAT - schema.org alignment (profile)


dr-shorthair:
Prepare a formal alignment of DCAT with schema.org

Also see #65

dr-shorthair:
This is a separate issue to prepare the alignment. 
#65 concerns guidance on use of DCAT in weakly axiomatized environment, which probably has a dependency on the alignment/mapping, but is a distinct concern. 

**Relevant commentary copied over from #65** 

 @andrea-perego commented on 20 Jan • 
Some work has been carried out in the framework of SDW that can turn to be relevant here:

https://www.w3.org/2015/spatial/wiki/ISO_19115_-_DCAT_-_Schema.org_mapping

The specified mappings were based on existing work, including a mapping of DCAT-AP, GeoDCAT-AP, and StatDCAT-AP to Schema.org:

https://ec-jrc.github.io/dcat-ap-to-schema-org/

 @makxdekkers commented on 7 Feb
This would be very useful for tools that could create the relevant schema.org markup for a landing page for the dataset based on the DCAT description.

 @andrea-perego commented on 8 Feb • 
This would be very useful for tools that could create the relevant schema.org markup for a landing page for the dataset based on the DCAT description.

The DCAT-AP to Schema.org mapping exercise I mentioned earlier in this thread includes also SPARQL queries implementing such mappings:

  https://github.com/ec-jrc/dcat-ap-to-schema-org/tree/master/sparql

 @dr-shorthair commented on 1 Mar
I propose adding an RDF alignment document for DCAT-schema.org to enable this discussion

 @dr-shorthair commented on 7 Mar
More mappings from SDWWG and Project Open Data added in #146

 @dr-shorthair commented on 23 Mar
Preliminary DCAT-schema.org alignment here: https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/schema.ttl

dr-shorthair:
RDF file renamed to `dcat-schema.ttl`
Work underway in branch `/dcat-alignments-schema/` - see
https://github.com/w3c/dxwg/blob/dcat-alignments-schema/dcat/rdf/dcat-schema.ttl

dr-shorthair:
| DCAT element | mappings | to schema.org |
|-----------------:|:-----------:|-----------------|
| dc:description | owl:equivalentProperty | schema:description |
| dc:title | owl:equivalentProperty | schema:name |
| dct:description | owl:equivalentProperty | schema:description |
[...]

PR #354, opened 2018-09-17
Dcat alignments schema

dr-shorthair:
Tabulation of the provisional DCAT-schema.org mapping, for potential feedback through 2PWD. 
See preview at https://rawgit.com/w3c/dxwg/dcat-alignments-schema/dcat/index.html#alignments 

In response to #251


agbeltran:
W.r.t Google Dataset Search, we should also point to https://developers.google.com/search/docs/data-types/dataset where it indicates that equivalent DCAT structures to schema.org are also indexed.

agbeltran:
I'm going to do minor edits on the text based on the acknowledgement at the bottom of https://schema.org/Dataset

"This class is based upon W3C DCAT work, and benefits from collaboration around the DCAT, ADMS and VoID vocabularies. See http://www.w3.org/wiki/WebSchemas/Datasets for full details and mappings."

There was a mapping between sdo, dcat, adms and void at https://www.w3.org/wiki/WebSchemas/Datasets

dr-shorthair:
@agbeltran Suggest you make them directly on this branch so we can do a single merge.

larsgsvensson:
Is the mapping provided at https://www.w3.org/wiki/WebSchemas/Datasets really correct? It says that `dcat:Dataset` is equivalent to `void:Dataset`. I always thought that `dcat:Distribution` is equivalent to `void:Dataset` since a `void:Dataset` [is defined thus](https://www.w3.org/TR/void/#dataset): "A dataset is a set of RDF triples that are published, maintained or aggregated by a single provider." `dcat:Dataset`s shouldn't know anything about abstract or concrete syntaxes.

agbeltran:
We won't be claiming that it is correct, but pointing to it as the mapping that was done to extend schema.org for describing datasets and data catalogs.

davebrowning (commented):
That last commit - should it say `<code>schema</code>` rather than `<code>code</code>`?
Apart from that - all looks fine, so I've made that change

PR #355, opened 2018-09-18
A worked example to test the expressivity of ProfDesc

dr-shorthair:
In this worked example I tried to create a description of ADMS as a _profile_ of DCAT. 
I found the following issues:

1. there is no predicate to link from an Implementation Resource Descriptor to the actual artefact - in some of the other examples `pd:resource` is used, but I believe this is overloading that predicate which links from a `Profile` to a `ImplResourceDesc`. The latter is an 'association class' which supports providing information about the nature and role of the artefact in the context of the profile. So I added `pd:artefact`

2. I do not understand what `pd:inheritedFrom` is to be used for.


rob-metalinkage (commented):
pd:artefact makes sense - it allows us to constrain the range of pd:resource.

pd:inheritedFrom (could be renamed) allows "flattened" profiles, where all the relevant resources can be transitively referenced from a single object - we be retain metadata about which base profile the resource relates to.  Another option would be to specialise the resource role so an inherited set of constraints is distinguishable from the specific constraints of the profile.

The use case is for profiles to be simple to write and read - only need to include specific constraints, but there is a way to package the full set. (we could ignore and require some future constraint language to support this)

dr-shorthair:
I've created a new issue #362

dr-shorthair:
Superseded by #363

PR #356, opened 2018-09-18
suggestion of some text about profiles in general to lead into the doc…

pwin:
It is a "profile guidance" doc so I am suggesting we need to start with something about what a profile is *generally* understood as being.


aisaac (approved):

aisaac:
I like these suggestions. I'm keeping them as notes as you've done, but I think they would be great as a general introduction. It gives a sense of depth and craftsmanship that is certainly very appropriate to our case.

Issue #253, opened 2018-06-11
Best practice for a loosely-structured catalog


jakubklimek:
@dr-shorthair raised this in the mailing list:
> I’ve been doing some investigations of some local repositories and catalogues, and have uncovered that in many cases ‘datasets’ are ‘just a bag of files’. There is no distinction made between part/whole, distribution (representation), and other kinds of relationship (e.g. documentation, schema, supporting documents). So while the precision we are aiming for in DCAT is clearly valuable in terms of semantics, it is difficult to implement on these legacy systems. Mostly I see people using the Dataset-distribution-> relationship for everything … which is clearly incorrect in many cases. But I doubt if we are unusual in this.
>
> I’m thinking about how to advise on this, while not actually breaking DCAT.
If we made dcat:distribution a sub-property of dct:relation
dcat:distribution rdfs:subPropertyOf dct:relation .
> 
> then I think we can have a reasonable recommendation to the simple repositories.
> We could tell repositories that use the ‘just a bag of files’ approach to say

```turtle
 :Dataset987 a dcat:Dataset ;
     dct:relation <file1> , <file2> , <file3> , <file4> , <file5> , <file6> , <file7> … .
```
> which would not be inconsistent with a later reclassification to
```turtle
  :Dataset987 a dcat:Dataset ;
              dct:hasPart <file1> , <file2> ;
              dcat:distribution <file3> , <file4> ;
              dct:conformsTo <file5> ;
              dct:requires <file6> ;
              dct:references <file7> .  
```
> If this is not all mad, I will add a new use-case - something like ‘Mapping from simple repository model’ – as justification, and propose this tiny enhancement.

I had a few concerns regarding this proposal:

1. It is not clear to me from the description what exactly the file* IRIs are. If they were actual downloadable files, i.e. something originally linked using `dcat:downloadURL`, I would disagree with the possibility to allow linking them directly from a `dcat:Dataset` record, as this would create mess everywhere where a publisher would be a bit lazy to describe the data properly.
2. Would it be possible to get a few more detailed examples of how this would work?
3. In my experience, data publishers use the `dcat:distribution` in a wrong way mainly due to the lack of support for dataset series, which is being resolved in this DCAT revision. When this support is added, publishers will have the possibility of modeling many use cases correctly.

dr-shorthair:
I should have written 
```
<resource1> , <resource2> , <resource3> , <resource4> , <resource5> , <resource6> , <resource7>
```
[...]

Issue #106, opened 2018-02-07
Provide guidance in DCAT2 on how to extend Distribution


nicholascar:
In DCAT, a Distribution "represents a specific available form of a dataset" and some few properties are given for a Distribution that presuppose certain forms of availability, for example downloadURL indicates the Dataset may be downloaded and byteSize indicates bundles, electronic delivery (as opposed to streaming, Web Services, a printed copy etc.). 

Can DCAT2 generalise a Distribution to cater for more and unknown future forms of Distribution by indicating, in essence, what a distribution should contain vis a vis a Dataset is distributes and how to go about specialising that essence for particulars?

A corollary is to also remove specialised properties, such as dcat:bytesize from distribution.

This is motivated by work within the Research Data Alliance (the Storage Service Definitions WG: https://rd-alliance.org/groups/storage-service-definitions-wg) to characterise data storage and compute scenarios which will have specialised information required for their decision making which could, with guidance, be presented as sepcialised forms of a Distribution.

makxdekkers:
I agree that the definition of Distribution could be made more future-proof. It now supposes too much that a Distribution is a file.
However, I vehemently oppose removing properties in general. In the case of bytesize, it is used in practice (e.g. by around 15,000 Distributions in the European Data Portal).

andrea-perego:
> I agree that the definition of Distribution could be made more future-proof. It now supposes too much that a Distribution is a file.

+1. A clarification of what `dcat:Distribution` is would indeed be beneficial. And, for this purpose, it may be worth checking the relevant issues from the GLD WG:

https://www.w3.org/2011/gld/track/issues/

The word "distribution" appears in the title of quite a few of them. 

> However, I vehemently oppose removing properties in general. In the case of bytesize, it is used in practice (e.g. by around 15,000 Distributions in the European Data Portal).

I totally concur.

davebrowning:
We have a requirement to do that clarification at #52.  I think that needs to subsume this, as well as others such as #53 and #54 (which also touches on 'a distribution isn't just a file'.
I suspect #56 needs to be included as well - what are the 'services' that we expect?

nicholascar:
@makxdekkers: I dont propose to remove the possibility for people to use byteSize, only that particular distribution type properties like this (for a file) be made non-core members of a more general Distribution. A File Distribution could still use byteSize, format etc.

makxdekkers:
[...]

Issue #55, opened 2018-01-18
Distribution schema [RDIS]


jpullmann:
<section class='requirement' id='RDIS'><h3>Distribution schema [<a href='https://www.w3.org/TR/dcat-ucr#RDIS' target='_blank'>RDIS</a>]</h3><p>Define a way to include identification of the schema the described data conforms to</p><p class='note'>This may include rich information via extensions points, URI templates and parameters, dimensions and subsetting operations, dereferenceable identifiers of service behaviour profiles and canonical identifiers of well-known web service interfaces (e.g. OGC - WFS, WMS, OpenDAP, REST apis).</p><p class='note'>Such a description may be provided through identifier of a suitable profile that defines interoperability conditions the distribution conforms to.</p><br/><em>Related requirements</em>: <a href='https://www.w3.org/TR/dcat-ucr#RPFL' target='_blank'>Profiles listing [RPFL]</a> <a href='https://www.w3.org/TR/dcat-ucr#RDSAT' target='_blank'>Dataset aspects [RDSAT]</a> <br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID6' target='_blank'>DCAT Distribution to describe web services [ID6]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID33' target='_blank'>Summarization/Characterization of datasets [ID33]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID17' target='_blank'>Data access restrictions [ID17]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID22' target='_blank'>Template link in metadata [ID22]</a> </section>

makxdekkers:
The European DCAT-AP (https://joinup.ec.europa.eu/release/dcat-ap-v11) specifies the use of dct:conformsTo (http://dublincore.org/documents/dcmi-terms/#terms-conformsTo) with usage note "_This property refers to an established schema to which the described Distribution conforms._" This seems like a solution to the issue as defined in the first line. 
The remainder of the text in the issue description could be part of a usage note.

dr-shorthair:
Notwithstanding little evidence of usage in https://github.com/SEMICeu/DCAT-AP/issues/25 I agree that this looks an eminently practical solution which could be promoted into DCAT. 

Proposal: add dct:conformsTo to the list of properties recommended for use with dcat:Distribution https://w3c.github.io/dxwg/dcat/#class-distribution

andrea-perego:
[...]

Issue #52, opened 2018-01-18
Distribution definition [RDIDF]


jpullmann:
<section class='requirement' id='RDIDF'><h3>Distribution definition [<a href='https://www.w3.org/TR/dcat-ucr#RDIDF' target='_blank'>RDIDF</a>]</h3><p>Revise definition of Distribution. Make clearer what a Distribution is and what it is not. Provide better guidance for data publishers.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID34' target='_blank'>Relationships between Distributions of a Dataset [ID34]</a> </section>

makxdekkers:
I liked the statement in one of the meetings that all Distributions of a Dataset need to be "informationally equivalent".

dr-shorthair:
https://www.w3.org/TR/dcat-ucr/#ID18 provides some history: a taxonomy of distribution types was contemplated, but did not make the cut for DCAT v1. Do we need to look at this again?

dr-shorthair:
In F2F3 @jpullmann suggested that the definition of `Distribution` be clarified to emphasize that distributions are primarily concerned with _representations_ of the dataset - in the REST sense (also see FRBR 'Manifestation'). 

This clarification is consistent with the introduction of `DataDistributionService` as a separate class. See #172

dr-shorthair:
The [definition of dcat:Distribution](https://w3c.github.io/dxwg/dcat/#class-distribution) has been tweaked slightly so that the word 'representation' appears explicitly, and API and feed is removed (DataService should be used for those applications) - 

> A specific representation of a dataset. A dataset might be available in several different forms, and these forms might comprise both different serializations or different schematic arrangements of the same data. Examples of distributions include a CSV file, a netCDF file, or a data-cube

(Compare to [DCAT-2014 where it says](https://www.w3.org/TR/vocab-dcat/#class-distribution) "Represents a specific available form of a dataset. Each dataset might be available in different forms, these forms might represent different formats of the dataset or different endpoints. Examples of distributions include a downloadable CSV file, an API or an RSS feed")

@jpullmann is this enough to mark this issue 'closed'?

makxdekkers:
[...]

PR #357, opened 2018-09-18
Sundry editorial

davebrowning:
This PR includes another general cleanup as email

1. Many which credit the original DCAT 2014 for section content could probably now be squashed. There is pervasive re-writing by now. **There was only one worth removing.  I've left a single instance (the Introduction) to point out we have the same intention as 2014**
2. The note on DCAT scope in the overview could probably be shortened to a couple of sentences . **Don't agree - it depends which parts of the document you read - having this  in the overview is better than having it in Conformance like 2014.  If we do a primer....**
3. There is a bit too much repetition – e.g. there is an almost identical note about dctype:Dataset in both the Overview and against the actual definition of Dataset in chapter 6 - **Done**
4. Issue #253 is closed – remove pink box in https://w3c.github.io/dxwg/dcat/#Class:Dataset **Done**
5. Issue #52 could be closed and pink box removed from https://w3c.github.io/dxwg/dcat/#Class:Distribution **Have added new Note to cover other aspect/clarifications on use of Distributions - if that works then yes, we can close**
6. Issue #55 is closed – remove pink box in https://w3c.github.io/dxwg/dcat/#Class:Distribution **Done**
7. I think issue #106 was overtaken by the resolution of #52? Remove pink box in https://w3c.github.io/dxwg/dcat/#Class:Distribution **Done - change note also tweaked**

I've added some text to point at appendix with change list to hopefully guide the reader.   Also linked the relevant open issue to `dct:rights` as well 'dct:license`, and a couple of minor links.

I **haven't** rationalised the vocabularly definition in this PR.  If there is time I could do that tomorrow.


pwin (approved):
there is some reference to a conflict at lines 189-195, but I cannot see what it causing it

davebrowning:
That looks okay now.  Will delete the extraneous note next time - its not wrong, just superfluous

Issue #350, opened 2018-09-13
Expand note to explain reason for removal of sub-class relationship to dctype:Dataset 


dr-shorthair:
[Feedback from DDI](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Sep/0000.html) asks about relationship between `dcat:Dataset` and `dctype:Dataset` which was relaxed in #98. 

On the one hand the relationship between `dcat:Resource` and the DCTYPE vocabulary is mentioned in [Section 5.3](https://w3c.github.io/dxwg/dcat/#classifying-dataset-types) . 

But it is probably worth expanding the NOTE against the definition of `dcat:Dataset` to indicate _why_ the sub-class axiom was removed.

agbeltran:
The follow up of this discussion is available [here](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Sep/0002.html) and [here](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Sep/0003.html). PR merged, so closing the issue.

PR #358, opened 2018-09-18
Remove superfluous note and stray figure

davebrowning:
Actions item 3 in PR #357 - needless note in vocal overview


agbeltran:
@davebrowning are you saying that it is superflous because is the same as the one appearing in the dcat:Dataset section?

https://rawgit.com/w3c/dxwg/dcat-glitch-fix-daveb/dcat/index.html#Class:Dataset

isn't this the note added as per issue https://github.com/w3c/dxwg/issues/350 and PR https://github.com/w3c/dxwg/pull/351

?

dr-shorthair:
I suggest either removing the first occurrence (in section 5) or cutting it down substantially.

dr-shorthair:
ALso, need to remove Figure 2. 
I moved Figure 1 up to the top of the section, so it looks like the merge got confused.

davebrowning:
I was suggesting that we drop the one in section 5. I think the subclass removal is adequately explained (and indeed more fully explained) in in the vocabulary definition itself (and noted in the change log at the back, which links to the original issue #98).  #350 and PR #351 added the extra detail to the vocab def NOTE.

More generally, I think the vocab overview (section 5) shouldn't go overboard in discussing changes cf 2014, but rather explains what the revised spec aims to achieve.  If that gets broken out into a separate doc/primer then that's fine by me, but we haven't discussed/agreed that approach.  For now, this is the only such NOTE in the vocab overview...

(Let me add the fix to the figures to this, too)

dr-shorthair:
+1

agbeltran:
+1

dr-shorthair (approved):

agbeltran (approved):

PR #359, opened 2018-09-18
Additional scope note for hasPart, …

dr-shorthair:
drawing attention to the specific sub-properties


davebrowning (commented):
Slightly confused on this, but perhaps that's because I'm ignorant on DC.  For example, in index.html, the URL http://purl.org/dc/terms/dataset gives me a file that doesn't mention dataset.  What am I missing?

dr-shorthair:
Right. That's because the names and links are wrong 😕 should be dcat:dataset etc, not dct:dataset. Sorry. 

This is why we have a review phase... Easy enough to fix.

dr-shorthair:
... Though not on a tablet. TTL also needs fixing

dr-shorthair:
Fixed - see https://rawgit.com/w3c/dxwg/dcat-issue81-simon/dcat/index.html#Property:catalog_hasPart

davebrowning (approved):
Excellent

Issue #57, opened 2018-01-18
Quality-related information [RDQIF]


jpullmann:
<section class='requirement' id='RDQIF'><h3>Quality-related information [<a href='https://www.w3.org/TR/dcat-ucr#RDQIF' target='_blank'>RDQIF</a>]</h3><p>Define a way to associate quality-related information with Datasets.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID45' target='_blank'>Annotating data quality [ID45]</a> </section>

makxdekkers:
Include a reference to DQV (https://www.w3.org/TR/vocab-dqv/)?
The European StatDCAT-AP (https://joinup.ec.europa.eu/release/statdcat-ap-v100) specifies the use of property dqv:hasQualityAnnotation (https://www.w3.org/TR/vocab-dqv/#dqv:hasQualityAnnotation) with range dqv:QualityAnnotation (https://www.w3.org/TR/vocab-dqv/#dqv:QualityAnnotation) and usage note "_The annotation requires the provision of information about the motivation of the annotation (oa:motivation), and an explicit link to the resource being annotated (oa:hasTarget) together with either a link to a resource that contains the annotation (oa:hasBody) or text field (oa:bodyText)._"

riccardoAlbertoni:
I have included  two proposals (Proposal 1 and Proposal 2)  in the [wiki page](https://github.com/w3c/dxwg/wiki/Quality-documentation), which are based on the DQV and we might want to discuss and refine in order to meet this requirement.

aisaac:
I am very much in favor of trying Proposal 1. Duplicating DQV in DCAT seems quite a counter-productive effort unless there is a strong reason to do it - which I do not see!

andrea-perego:
+1 from me as well to Proposal 1 - I don't see the need of duplicating / subclassing DQV in DCAT.

riccardoAlbertoni:
>I am very much in favor of trying Proposal 1. Duplicating DQV in DCAT seems quite a counter-productive effort unless there is a strong reason to do it - which I do not see!

Replying to @aisaac:
Neither do I.  I was just provoking to see if others have a different view on this.

agbeltran:
+1 to proposal 1 that reuses DQV

dr-shorthair:
Resolved in DCAT telecon 2018-06-14 https://www.w3.org/2018/06/14-dxwgdcat-minutes.html#x06

PR #360, opened 2018-09-19
dcat :deleting ref to issue #57 which was closed and correcting link t…

riccardoAlbertoni:
…o issue 63


davebrowning (approved):
Good catch on the bad link, and agree that we've addressed #57 in section 7 (and acknowledged the issue it came from.)

Issue #58, opened 2018-01-18
Data quality model [RDQM]


jpullmann:
<section class='requirement' id='RDQM'><h3>Data quality model [<a href='https://www.w3.org/TR/dcat-ucr#RDQM' target='_blank'>RDQM</a>]</h3><p>Identify common modeling patterns for different aspects of data quality based on frequently referenced data quality attributes found in existing standards and practices.</p><p scope='note'>This includes potential use and revision of DQV</p><p>Aspects include:<li>the degree of a dataset's precision (i.e. measure of resolution or variability).</li><li>the degree of a dataset's accuracy (i.e. measure of correctness).</li><li>the degree a dataset conforms to a stated quality standard.</li><li>details of data quality conformance test results.</li></p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID15' target='_blank'>Modeling data precision and accuracy [ID15]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID14' target='_blank'>Data quality modeling patterns [ID14]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID16' target='_blank'>Modeling conformance test results on data quality [ID16]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID21' target='_blank'>Machine actionable link for a mapping client [ID21]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID22' target='_blank'>Template link in metadata [ID22]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID23' target='_blank'>Data Quality Vocabulary (DQV) Wish List left by the DWBP WG [ID23]</a> </section>

riccardoAlbertoni:
Some remarks to start discussing this requirement.

**Remark 1:**
The Data Quality Vocabulary (DQV)[1] already offers common modelling patterns for different aspects of Data Quality. 

For those who are not familiar with DQV,  it relates DCAT datasets and distributions with different types of quality statements including 
  
- dqv:QualityAnnotation, which represents feedback and quality certificates given about the dataset or its distribution.
- dcterms:Standard, which represents a standard the dataset or its distribution conforms to.
- dqv:QualityPolicy, which represents a policy or agreement that is chiefly governed by data quality concerns.
- dqv:QualityMeasurement which represents a metric value providing quantitative or qualitative information about the dataset or distribution.

[...]

PR #361, opened 2018-09-19
dcat: Moving and rephrasing the note about issue 58

riccardoAlbertoni:
Please consider to accept this pull, it rephrases and  moves the note about issue #58 to the beginning of section 7.2, so that, comments and suggestions from the readers are invited.


davebrowning (approved):
Better wording, and a better location

PR #421, opened 2018-09-25
Minor clean-up

dr-shorthair:
- schema: namespace
- correct dc: --> dct: in a couple of places in mapping
- add to list of sub-properties of dct:relation

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2739,6 +2739,11 @@ a:TestingActivity a prov:Activity;
 <td>schema:Organization</td>
 </tr>
 <tr>
+    <td style="text-align: right">foaf:Person</td>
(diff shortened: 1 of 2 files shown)


agbeltran:
I had added a note about the prefix in the alignment section (https://w3c.github.io/dxwg/dcat/) because I assumed (wrongly) that the Namespaces at the top were the relevant namespaces for DCAT and not the document. Should we now remove the reference about the prefix in the alignment section?

agbeltran:
@dr-shorthair see my commit with a few minor fixes and one addition to the mapping

dr-shorthair:
+1

davebrowning (approved):

Issue #419, opened 2018-09-25
Examples in profile guidance


agreiner:
I think the examples should come last, as indicated in section 6. A few could be mentioned in the intro for context, but it seems odd to jump into describing specific ones before we've offered guidance.

kcoyle:
To clarify, by examples are you referring to the ones in section 4 Related Work?

agreiner:
yes

aisaac:
Thanks @agreiner for the comment! And @kcoyle for clarifying. 
Am I right in understanding that this issue is then very close to #366 , as @kcoyle said there that in her mind the 'related work' in section 4 was mostly about giving examples? If yes I'm going to do a cross-reference (and not add a mention of this issue in the doc itself)

Also, in this case I could propose to elaborate on @kcoyle 's suggestion in #366 whereby she suggest to move these 'related work examples' in the introduction. 
I think we won't be able to put a fully fledged description of the examples in the introduction. But what we could do is to have a *brief* description of the 'related work examples' in the intro. We could then compensate for this shortness by coming back to these examples in sections 6 or 7, i.e. as examples of using ProfDesc to create descriptions of profiles and examples of our recommendations being implemented in the wild (even if these implementations are pre-existing to our recommendations, they could still be relevant as examples).

Would this be ok?

agreiner:
sounds good to me! Thanks.

rob-metalinkage:
+1

aisaac:
Considering the reactions on here after cross-linked from #366 , I'm suggest the PR at https://github.com/w3c/dxwg/pull/439 to close the issue.

aisaac:
Closed as @kcoyle has reviewed PR #439 and after positive feedback from @agreiner and @rob-metalinkage . Thanks!

Issue #366, opened 2018-09-20
Location of related work


nicholascar:
This issue was created in the [Profile Guidance document](https://w3c.github.io/dxwg/profiles/) and is listed in it. Once consensus on addressing it is reached here in comments below, the results will be added to the document and the issue closed.

(Antoine:) This section was recommended here in https://github.com/w3c/dxwg/issues/242#issuecomment-408916364. I now have doubts about the it, as it breaks the flow between two crucial (normative) sections on DXWG recommendations. As an alternative, I suggest to try to fit this in (a subsection of) the introduction if it can be short enough. Otherwise we could put it as appendix and refer to it from the introduction.

kcoyle:
Suggests moving related work from body of document to the introduction.

A short treatment of this topic in the introduction could be amplified (if desired) by a separate analysis document or an appendix to the profgui document. There is definitely a certain history of profiles that could help readers understand the context of this work; I'm not sure how far we want to go with that, however. I'm mainly interested in giving readers an idea of the breadth of the topic by showing different styles of profiles and different purposes rather than attempting an extensive discussion of existing profiles.

aisaac:
@kcoyle had put a note which I think relates to the issue:

> I had thought of this as showing examples of TYPES of profiles. There are profiles like DCAT that take place within a community, make primary use of a base metadata schema, but add or remove based on local need. There are profiles like ODRL that have a very formal meaning (legal license) and are in essence cumulative, with the base standard elements assumed and specific rules for over-riding. There are profiles that use only the namespace of the base profile, but are defined subsets (BIBFRA.ME, but any other example of this would be fine - that's the one I know). And then we could perhaps find a profile that has no strong base profile (ADMS perhaps?) but combines elements as needed from a variety of sources and is independent of any existing vocabulary. The idea would be to show a range of profile motivations, designs, and community involvement.

aisaac:
@agreiner 's comment in #419 is very much related to this issue.

rob-metalinkage:
Its a very long document if we start reviewing all the cases where some sort of profile might be implied through recommendations to reuse vocabularies.  This is why we must stick to process and address requirements introduced via Use Cases.

What is relevant here is that some profiles are (sometimes partially) described using constraints languages, and there are a range of such languages.

If we take "specification" as the intention of dct:Standard, then under for example OpenGeospatial Consoritium model, the "conformance target" must be identified and a set of testable requirements identified. (constraints languages introduce the "testable" aspect.

[...]

PR #439, opened 2018-10-03
Shuffle related work examples

aisaac:
Moved examples from their own section to a bit in the intro and a bit in the section on examples of applying best practices. This is to address #366 and #419


kcoyle (approved):
I'm sure that the content of this section will change as we work on the document, but the position looks good to me.

Issue #434, opened 2018-10-03
accessService property of a Distribution


agreiner:
I find it confusing that we define a property of a Distribution to point to a thing called an accessService, which is a DataDistributionService. A DataDistributionService is a service that serves data through an API, no? But a Distribution is now defined not to include APIs, so why does it have an accessService property?

dr-shorthair:
A 'Distribution' describes a file representation or serialization of a Dataset. 
It is available directly from the dcat:downloadURL
It is available from a service endpoint at dcat:accessURL
It is available through a service described in a DCAT catalogue as indicated by dcat:accessService.. 

This is related to the introduction of services to DCAT raised primarily in #172, #180, #181 
There was detailed discussion in May when it was resolved.

agreiner:
I think my confusion is coming from the Note where we say "Under the revised scope, instances of type dcat:Distribution SHOULD be limited to representations of datasets which might be transported as files, and SHOULD NOT be used for data services such as APIs or feeds." Is it the case that a distribution is meant to still be *available* via an API or feed? If so, the fix is relatively easy.  "Data services including APIs and feeds SHOULD be described using instances of type dcat:DataService or dcat:DataDistributionService, which in turn serve dcat:Distributions."

agreiner:
Sorry to bring up an old issue, but it was only resolved in the DCAT subgroup. It's only now that the plenary is being asked to look at the document "with fresh eyes."

agreiner:
Probably that should say "which in turn *may* serve dcat:Distributions

dr-shorthair:
PR #440 has 

Under the revised scope, instances of type dcat:Distribution SHOULD be limited to representations of datasets which might be transported as files, and SHOULD NOT be used for data services such as APIs or feeds. Data services including APIs and feeds SHOULD be described using instances of type dcat:DataService whose sub-class dcat:DataDistributionService MAY serve dcat:Distributions.

davebrowning:
@agreiner - so we can close this now?

agreiner:
yes, I think so.

PR #440, opened 2018-10-04
Improve documentation of dcat:Distribution

dr-shorthair:
to clarify matter raised in #434


agreiner (approved):
This looks good to me for now. We may need to think about the MAY in the future, but I don't think we need to resolve that now.

agbeltran (approved):

Issue #417, opened 2018-09-25
Improving profile guidance intro


agreiner:
In the introduction, there is no mention of profiles before the last sentence, which is quite disconnected from the text immediately above it. There needs to be another paragraph to explain that profiles are the solution to the problem posed above.
I also think that the motivation section belongs as part of the introduction.

rob-metalinkage:
makes sense. Do you want the editors to do another draft or propose changes via a PR?

As the doc gets larger, the intro needs to introduce it - so keeping the motivation separate still feels right to me at this stage.

kcoyle:
Here's an attempt at a new introduction. It uses most of what Phil had written, but adds a paragraph at the beginning. I'm not sure about the part that talks about "best practices" for ontologies - I agree with it but I'm not sure what we should say about it here. In any case, I've left his text intact.

(updated with wording from @agreiner, 05-10-2018)

****
A profile is generally understood as representing a point of view. In information technology, profiles may also support the data needs of specific applications. Profiling is often the work of a community interested in interoperability and data exchange. We define a profile generally as named set of constraints on one or more identified base specifications. 

Communities create and use data standards to ensure interoperability for information exchange. Although members of a community may use the same basic standard schema, it is very common for different subsets within the larger community to need some further specification of the data they create to meet their own needs. To continue to support interoperability of their data with others, these community members need to express the specifics of their implementation of the data schema. Profiles serve this purpose. Profiles enumerate vocabulary terms, cardinality, and validation rules, and can also include descriptions of the rules used by creators to make decisions regarding their data elements.

Good metadata practice begins with the builders of vocabularies and ontologies. Builders of vocabularies and ontologies are encouraged to make their work as broadly applicable as possible so as to maximize future adoption. As a result, vocabularies and ontologies typically define a data model using minimal semantics. For example, DCAT [vocab-dcat-2] defines the concept of a dataset as an abstract entity with distributions and data services as means of accessing data. It is silent on whether a distribution should be in a particular serialization, or set of serializations. It is also silent on how data services should be configured. While it states that the value of dcat:theme should be a SKOS concept, it does not specify a particular SKOS [skos-reference] concept scheme, and so on. Other vocabularies such as Dublin Core Terms [DCTERMS] are equally parsimonious in their prescriptions of how they should be used. This means that data models and methods of working can be applied in different circumstances than those in which the original definition work was carried out and, in that sense, these promote broad interoperability.

[...]

Issue #428, opened 2018-10-02
Strange mention of Guidance in Requirements


aisaac:
Hi @nicholascar 
I guess it's you who edited, this, apologies if you're not...
In the current https://w3c.github.io/dxwg/profiles/#requirements I can read:
> Guidance
This requirement implies that a profile is an identified resource that can be accessed, and hence MUST have a URI that resolves to a representation resource. This resource SHOULD be appropriate for the environment being profiled, but MAY be made available in alternative forms using negotiation.

Is it intended to be like this, with no reference to a github issue?

akuckartz:
Additional question: Is the "MUST" refering to both "have a URI" and "resolves to a representation resource" ?

If yes: Would "have a URL" combine both?

aisaac:
@akuckartz thanks but this issue is only about a super-basic editorial question! Questions like yours should be addressed later on, when we have clarified whether this sentence is really at the place it was intended to be (I don't think it is, as it stands out compared to what's around)

nicholascar:
That text, in grey, was an early attempt to show how Issues (Requirements) may be addressed. I think Rob authored that one but the logic for the answer was taken from discussion.

I have created a commit (to soon be PRed) to remove this now vestigal text from the doc since we are now dealing with Issues directly in GitHub, not in the doc, until final text is agreed to.

Closing Issue with commit https://github.com/w3c/dxwg/commit/febbe0df548cd70243b1e9d75b4f37e5aa1f4a7d

aisaac:
@nicholascar thanks for the answer. This sounds very good but I'm not fond of closing the issue until your commit is deployed at https://w3c.github.io/dxwg/profiles/. Where is it?

nicholascar:
Text removed

PR #446, opened 2018-10-05
Substituted introduction from github discussion #417

kcoyle:


aisaac:
Thanks Karen I think it is good. 

I would limit my feedback here to two comments, as I don't want the discussion to happen here and not in #417 :-)

1. replace "Good metadata practice" by "Good data practice", following up on @agreiner's feedback 

2. use the complete definition from wiki (https://www.w3.org/2017/dxwg/wiki/ProfileContext#Draft_common_definition_for_.22profile.22)

> We define a profile generally as named set of constraints on one or more identified base specifications, including the identification of any implementing subclasses of datatypes, semantic interpretations, vocabularies, options and parameters of those base specifications necessary to accomplish a particular function.

I know it's imperfect, but after so much discussion I think it would be a pity not to use it and then be drawn in the same discussions...

Finally I see there's little from @pwin's suggested text (as you've cited at https://github.com/w3c/dxwg/issues/417#issuecomment-427256227). But indeed I find it hard to integrate with your text. Maybe right at the beginning?

aisaac:
Flagging that @agreiner 's comment may bring even more motivation to take onboard more of Peter's text :-)

rob-metalinkage (changes requested):
"outline of the margin of a thing" is bloated and an oxymoron i feel - suggest "outline of a thing"  

As @aisaac points out - we spent ages agreeing a definition of a profile which meets the Use Cases we have - we really dont have an excuse or time to start again, unless we have strong evidence that this is inconsistent with some requirement, and a proposal to update it.  Otherwise it needs to be introduced verbatim early as possible in the document,.

nicholascar (changes requested):
Minor functional edit: [vocab-dcat-2] must be changed to [[vocab-dcat-2]]

Minor grammatical edit: "...as means of accessing data. It is silent on whether a distribution should be in a particular serialization, or set of serializations." --> "...as means of accessing data; it is silent on whether a distribution should be in a particular serialization, or set of serializations.".

Minor functional edit: "...SKOS concept, it does not specify a particular SKOS [skos-reference] concept scheme..." should be "...SKOS [[skos-reference]] concept, it does not specify a particular SKOS  concept scheme..." i.e. moving SKOS ref to first mention and also using double [[, ]] for ReSpec.

Minor grammar: "Other vocabularies such as Dublin Core Terms [DCTERMS] are..." --> "Other 
vocabularies, such as Dublin Core Terms [DCTERMS], are..."

Minor functional edit: [DCTERMS] --> [[DCTERMS]]

Please just take these minor points as OK and I'll include fixes for them in a follow-on PR in the next few hours.

nicholascar:
I've included the text from this PR in another PR with the minor edits I suggested above so please reject this PR and use that one! Currently stages in Branch [profile-guidance-doc](https://github.com/w3c/dxwg/tree/profile-guidance-doc) with commit https://github.com/w3c/dxwg/commit/2f1ce43f43408c17ca3d78bd01851d0ddd9219ff

aisaac:
Hi @nicholascar your new PR seems to be missing my own suggestions above, unless I'm mis-reading the PR over the PR? Anyway we should perhaps wait until the discussion at #417 has stabilized.

pwin:
"it is silent" really means "there is no mention of", surely?

nicholascar:
@aisaac Just give me another hour or so before I put in the PR and I'll include your suggestions. I'm just handing a bit of a "Issue Dedt" of a lot of issues listed but not placed in the doc yet first!

aisaac:
@nicholascar ok now I understand better (also I guess it answers my comment on #428). This said I would still separate your coming PR on issue debt and the one on the intro, as I think the intro still needs some debating in issue #417, while I'd happily approve some of your other 'debt' changes asap!

aisaac (changes requested):
Sorry I had not understood that my previous comment was not a review. So I guess this does it with a link to the comment: https://github.com/w3c/dxwg/pull/446#issuecomment-427430817
But I believe anyway it is perhaps best to wait until more discussion has happened at #417

kcoyle:
Closing because covered in #451

PR #458, opened 2018-10-09
Cleaned normative statements and references from non-normative sections

dr-shorthair:
Response to #353


agbeltran:
thanks @dr-shorthair - merging into the 2nd PWD branch will cause conflicts, though, as the difference is in the links (in the 2nd PWD branch, the links point to that branch in github)

agbeltran (approved):

PR #459, opened 2018-10-10
Copied lists of inherited properties into class descriptions

dr-shorthair:
See https://rawgit.com/w3c/dxwg/dcat-issue443/dcat/index.html#Class:Catalog for preview


agbeltran (approved):
I reviewed the changes and it seems more natural to me to include first the properties of the more generic class from which each class inherits (e.g. to list dcat:Resource first and dcat:Dataset second). I realise this implies changing the diagram in this respect. 

We can revert it if people don't agree. 

@makxdekkers maybe you can also have a look? Thanks

dr-shorthair:
Actually I can't change that ordering in the diagram. But the order in the text does not have to match exactly.

makxdekkers:
I actually disagree with @agbeltran. It looks more natural to me to go from subclass to superclass to super-superclass. It is done the same way in schema.org, e.g. https://schema.org/Dataset lists properties for Dataset first, then properties inherited from superclass CreativeWork, then the ones from Thing. Also, it seems better to me to keep the text in line with the diagram.
In addition, the lists of properties in the diagram are ordered strictly alphabetically on the namespace prefix/property label, while the properties in the text do not seem to follow any ordering. I would suggest to align the order in the diagram with the list under the class heading and with the sequence of the subsections.

makxdekkers (commented):
I actually disagree with @agbeltran. It looks more natural to me to go from subclass to superclass to super-superclass. It is done the same way in schema.org, e.g. https://schema.org/Dataset lists properties for Dataset first, then properties inherited from superclass CreativeWork, then the ones from Thing. Also, it seems better to me to keep the text in line with the diagram.
In addition, the lists of properties in the diagram are ordered strictly alphabetically on the namespace prefix/property label, while the properties in the text do not seem to follow any ordering. I would suggest to align the order in the diagram with the list under the class heading and with the sequence of the subsections

agbeltran:
thanks @makxdekkers - ok, I didn't have a strong opinion on this and agree that is better that the text matches the diagram, so we can revert my last change and address the issue of alphabetical order.

dr-shorthair:
I generated the diagram using UML tool Enterprise Architect. It provides a lot of control over layout, but not on the order of class attributes (that I've been able to find anyway). So if it is required to adjust the ordering from strict alphabetical a different tool will be needed. :-(

agbeltran:
I'm working on making sure the list of properties matches the order of the diagram, and I was wondering if is there a reason why the diagram doesn't include the properties for ```dcat:DataDistributionService```? @dr-shorthair

dr-shorthair:
I suppressed the class attributes visibility because there are (currently) no new ones on either `dcat:DataDistributionService` or `dcat:DiscoveryService`, so it adds a lot of noise in the diagram without really adding much information. But it is inconsistent so either it should be added, or the fact that they are not shown should be noted in the caption. I can prepare the figure either way.

agbeltran:
Thanks @dr-shorthair - I am inclined to complete the diagram with all the properties even if showing explicitly that no specific property is added.

I've moved back to the initial order of the properties and re-ordered the classes.

https://rawgit.com/w3c/dxwg/dcat-issue443/dcat/index.html

IMO this PR is ready to be merged in if you agree @dr-shorthair @makxdekkers ?

dr-shorthair:
Inherited properties shown on all classes now

agbeltran:
@makxdekkers any other comments on this or can we merge it?

makxdekkers:
@agbeltran Good for me!

PR #474, opened 2018-10-13
Profilesont doc - pubrules

nicholascar:
Passing all pubrule tests
(actually, the validator indicates a single fail but it's manifestly not failing - fail is for lack of inclusion of patent policy, which is included, so I think there's a pubrules tester bug for FPWD of a Note)


nicholascar:
No broken links now

PR #475, opened 2018-10-13
Prof ns

nicholascar:
Using W3C namespace (http://www.w3.org/ns/prof/ - note it's a slach URI) and also updating Roles vocab.


nicholascar:
Cancelling this PR until the major pubrules PR is done

Issue #435, opened 2018-10-03
What is a profile?


kcoyle:
Many of our issues for the profile guidance document hinge on our definition of profile. I'm going to suggest some functions that a profile must/may/should have as a start to that conversation.

- MUST define vocabulary terms that are used in the metadata being profiled.
- Vocabulary terms SHOULD be defined as mandatory, optional, repeatable, not repeatable (alt.: cardinality rules for vocabulary terms SHOULD be defined in the profile)
- Human-readable labels and definitions MUST be included for vocabulary terms
- The profile SHOULD (MUST?) define what values are valid for vocabulary terms
- The profile MAY include input instructions for metadata creators (and to aid users in understanding the deeper meaning of terms)
- The profile MAY include operational validation code
- The profile MUST be published to the relevant metadata-using community
- The profile MAY be published as a human-readable document
- The profile MAY be published as operational code
- The profile MAY be expressed in more than one physical file
- The profile MAY be based on one or more known community standards, for interoperability within that community ("profile of X"), and, when it is, it must indicate that basis.
- The profile MAY combine vocabularies that cross community boundaries without having any strongly identifiable base profile or profiles.

(No, I'm not sure about all of this - just tossing out every idea I can come up with to aid discussion.)

kcoyle:
Instead of the layered view given in the DC [Singapore Framework](http://dublincore.org/documents/singapore-framework/) I've been thinking about a more function-based view, something like:
![profile](https://user-images.githubusercontent.com/1564129/46393652-64293500-c6e6-11e8-9836-f6e571a583de.png)

kcoyle:
Here's more or less what I had in mind for using examples to talk about types of profiles:

Profiles can take a number of forms and can have a variety of relationships to existing vocabularies, standards, and other profiles. Although it is not possible to list all of the types of profiles, some common types are:

- profiles that are subsets of a larger vocabulary. These reduce the vocabulary terms of a broad data standard to a smaller number of terms that are useful for a particular community member or application. An example of this is [BIBFRA.me](http://bibfra.me/) designed for library materials. It defines both a core set of terms as well as profiles for specialized communities such as cataloging of rare materials or early printing trade. In this community, all profiles use only terms from a single vocabulary.
[...]

Issue #242, opened 2018-05-25
profileDesc and the Guidance document


kcoyle:
There are two possible approaches we can take to integrating profileDesc into our second deliverable. First would be a small-ish modification to the outline proposed by Lars. A W3C-structured document would exist as a separate item, not on recommendation track. (I'm not sure what options there are for a non-recommendation ontology, but I believe they do exist.)

* Introduction (non-normative)
* Definitions
    * Profile
    * Schema
    * Formalism
    * ...
* Profile description
    * This is the meat of the document. Important parts are:
        * A profile MUST have URI (or an IRI?) identifying it. The URI SHOULD be an http URI
        * A profile SHOULD have a title
        * A profile SHOULD have a textual description intended for humans
        * A profile's description MAY be expressed using a formal model we define (such as @rob-metalinkage 's proposal)
        * A profile document SHOULD link to schemas implementing the profile it describes
* Examples (non-normative)
    * ODRL is a good example since their profiles are explicitly media-type independent
* _Profile Description Ontology (non-normative)_
    * Introduction with link to PD documentation that exists elsewhere
    * Examples
* Bibliography

In the other option, profileDesc *is* the main normative content of the Guidance document. This could be seen to fit the definition of the Guidance deliverable, which is:
     A definition of what is meant by an application profile and an explanation of one or more methods for publishing and sharing them.
One could read this as meeting this outline:

* Introduction (non-normative)
  * What is a profile
        * A profile MUST have URI (or an IRI?) identifying it. The URI SHOULD be an http URI
        * A profile SHOULD have a title
        * A profile SHOULD have a textual description intended for humans
        * A profile's description MAY be expressed using a formal model we define (such as @rob-metalinkage 's proposal)
        * A profile document SHOULD link to schemas implementing the profile it describes
  * Examples (non-normative)
       * ODRL is a good example since their profiles are explicitly media-type independent
* Definitions
    * Profile
    * Schema
    * Formalism
    * ...
* Profile description ontology (normative)
    * This is the meat of the document
      * Background
      * Design rationale
      * (primary classes, definition of properties, etc. cf. [SKOS document](https://www.w3.org/TR/2009/REC-skos-reference-20090818/))
* Bibliography

agreiner:
So the only thing required for a profile would be a URI?

andrea-perego:
@agreiner said:

> So the only thing required for a profile would be a URI?

I tend to agree that the URI is only mandatory requirement for a profile.

[...]

PR #489, opened 2018-10-24
Change of structure discussed in Oct 24 discussion in Lyon

aisaac:
The changes here have the following motivations:

- splitting the long list of requirements/recommendations (MUST/SHOULD/MAY) from https://github.com/w3c/dxwg/issues/435#issue-366191828 and https://github.com/w3c/dxwg/issues/242#issuecomment-408916364 into various sections where they can be better contextualised and explained (as consistent bits) to readers

- acknowledging that these recommendations did not fit well as intro material and giving more space in the introduction so that the discussion on types and examples of profiles (https://github.com/w3c/dxwg/issues/435#issuecomment-426928897) can 'breathe' better.

- provide finer-grained introduction (and better alignment) to other relevant DXWG work (the Profiles Ontology and Profile Negotiation)

- put a placeholder for a conceptual model to which the rest can be naturally attached,  and for which the Profiles Ontology would provide a straight implementation as a vocabulary.

- exploit the notion of role as first introduced in the Profiles Ontology as an interesting pattern to represent the functions that (expressions/components of) profiles can play, as sketched in #435.

### DIFF
--- a/profiles/index.html
+++ b/profiles/index.html
@@ -142,29 +142,20 @@
       <div class="issue" data-number="418"></div>
     </section>
     
-    <section id="whatisaprofile">
-      <h3>What is a profile?</h3>
-      <p>This section builds on the formal definitions for <a>profile</a> and other terms given above.</p>
-      <div class="issue" data-number="435"></div>
-      <div class="note">(Antoine:) From https://github.com/w3c/dxwg/issues/242#issuecomment-408916364:
-        <ul>
-          <li>A profile SHOULD have a textual description intended for humans</li>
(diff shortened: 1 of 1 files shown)


aisaac:
Changes are visible at https://raw.githack.com/w3c/dxwg/guidance-restructuring-LyonOct24/profiles/index.html

aisaac:
@nicholascar @rob-metalinkage @kcoyle your input in this PR is more than welcome but I think it shouldn't be merged just before the F2F, as it's quite a change and we may want to compare it with the rest of the group at some time in the coming two days.

aisaac:
Approved in Oct 25 F2F

PR #552, opened 2018-11-12
Typos only - minor

kcoyle:
Might be good to include in FPWD but not essential. Very small number of typos.


aisaac:
@nicholascar wrt your label addition: this is about conneg not guidance, isn't it?

larsgsvensson:
@aisaac scripsit:
> this is about conneg not guidance, isn't it?

Yes it is, I've removed the label.

larsgsvensson:
Thanks Karen.

PR #554, opened 2018-11-12
Adding reference to issue 551 in prof-ont document

agbeltran:
This PR adds reference to issue that will address the relationship between the prof-ont and prof-conneg documents. I believe it needs to be added before FPWD publication.

View here: https://raw.githack.com/w3c/dxwg/agb-prof-ont-ref-issue551/profilesont/index.html#motivation

### DIFF
--- a/profilesont/index.html
+++ b/profilesont/index.html
@@ -169,6 +169,7 @@
       </ul>
     </li>
   </ul>
+  <div class="issue" data-number="551"></div>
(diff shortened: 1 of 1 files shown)


rob-metalinkage (commented):
This seems reasonable to me... even though I'm not sure the negotiation part needs to be dependent on the ontology - but it can reference it as a means to achieve the goal of discovery.

PR #577, opened 2018-11-14
folder renaming as per shortcodes

nicholascar:
profiles -> prof-guidance
profilesont -> prof
conneg-by-ap -> prof-conneg


kcoyle (commented):
Nick, do you understand what the conflicts are? Is it something you can resolve, or is it a warning that we can ignore? 

Also, do we have confirmation of these short codes from DSR?

nicholascar:
No word from DSR. Also, the conflicts occur simply because of the multiple edits all flowing in on top of one another. Because of these two things, I'll cancel this PR now and re-issue it when DSR approves and the text edits  pause.

PR #580, opened 2018-11-14
Updated Figure 1

dr-shorthair:
1. add conformsTo as a property of catalog record
2. suppress attribute visibility indicators


davebrowning (approved):
This change looks good, but reviewing it made me wonder about the wording in the vocabulary desription.  Will progress that separately.

PR #590, opened 2018-11-15
Ont fpwd edits

nicholascar:


nicholascar:
I've addressed Andrea's continuing concerns in Commit https://github.com/w3c/dxwg/pull/590/commits/af19ea8d32456fd5a6366d909fa6abb80d3a1036 now listed above.

rob-metalinkage:
line 63 has some bad grammar -  "...them that define rules for implementing it"  is it easiest to fix this and update this PR or accept the PR and then catch this in another?

rob-metalinkage (commented):
ln 509 - Can be use to ind..   change to "used"

Issue #575, opened 2018-11-13
Antoine's conneg doc edits


nicholascar:
Note that there is one matter which I think should stand out: it's
#15, on compatibility with Linked Data conneg.

I hope this helps,

Antoine

=============

1. Abstract
The two last paragraphs ("For details about what a profile is, see the [PROF-GUIDE] [...]" and "For the formal ontology describing profiles, [...]") are clearly not abstract-level material.

2. Introduction (section 1)
"Thus, resources available in different languages" wrongly uses "thus". There's no direct logical implication between Media Type considerations and Language ones.

3. Motivation (3)
To me this section could be merged with the introduction. There's quite a bit of overlap, and I think putting the motivation early in documents always help.

I've re-opened https://github.com/w3c/dxwg/issues/379

4. Related Work content (4)
"What a profile is and how to create one is detailed in the [PROF-GUIDE] document [...]",

"Issue 380: Remove text in favour of full reference to final IETF
doc [...]" and

"Describing the parts of profiles and their relation to other profiles is done within the Profiles Ontology" are not related work material to me. They're rather about explaining our current (pieces of) work, and therefore seem a better fit for the intro.

5. Abstract Model/Context (5.1)
The paragraph that includes "For this reason, other than a directive to maintain independence, no further discussion of negotiation by profile and this relations to other forms of negotiation are given" reads strange in the flow of other paragraphs. It doesn't read like context, in fact.

6. Abstract Model/Context (5.1)
The paragraph starting with "A client requesting the representation of a resource conforming to a profile MUST identify the resource" clearly doesn't read like a piece of context, as it gives normative instructions.

7. Abstract Model/Context (5.1)
The sentence "In this abstract model, we don't assume any specifics about client, server, resource, metadata, request or response" reads strange as definitions have been given earlier about these notions. So there is a form of assumption going on.

8. Abstract Model/Requests and Responses (5.2)
I don't get why there's such a long introduction to 5.2. Not only this is long, but it results in splitting information that would be easier to understand if it was kept together in each of the subsections. For example it's strange to have 5.2.2 not saying what the server should do (this info is only in the intro).

9. Abstract Model/Requests and Responses (5.2)
"a server responds the list of profile tokens it is able to deliver resource representations conforming to and their mapping to profile URIs" is quite hard a sentence to swallow. Maybe adding a bit of punctuation and/or a conjunction would help.

10. Abstract Model/Requests and Response (5.2.1)
[...]

Issue #573, opened 2018-11-13
Rename Resource Descriptor class


nicholascar:
The Resource Descriptor class' name may not convey its intention well. This name needs to be reconsidered with discussion about it here.

agbeltran:
Discussion about this name also happening in https://github.com/w3c/dxwg/issues/404

aisaac:
For the record, for a similar notions, other proposals (my F2F3 diagram [1] and our F2F postits [2]) have used the following words, without making a final decision:
- profile description/specification/expression
- distribution/publication

Implementation may also be an option...

[1] https://docs.google.com/drawings/d/1dHkpwKwUwMgS1RqSCTPO3uOoRiY_qNk0z5bhXJlYi4Y/
[2] currently in the first page of https://docs.google.com/document/d/13hV2tJ6Kg2Hfe7e1BowY5QfCIweH9GxSCFQV1aWtOPg/

smrgeoinfo:
also a related suggestion w3c/dxwg#529, copied here
In the Figure 1. OWL [owl2-overview] diagram of this ontology, the ResourceDescriptor class is basically an associationClass that associates a resource with the Profile in some role with some purpose.  There might be multiple representations of this resource available, and these will have different linking information (URL, Mime Type etc.). The dcat:Distribution class is defined to model exactly this; shouldn't the target of the 'artifact' predicate/role  be dcat:Distribution?  The model would look something like this:
![image](https://user-images.githubusercontent.com/513380/47945975-4a386900-dec3-11e8-86db-c9fbf59c24db.png)

kcoyle:
I'm not convinced that the resource descriptor is necessary. If most supporting artifacts are one to one with a resource descriptor then they can be reduced to a single thing, and that thing is the supporting resource, which has a role and a physicality (MIME type). In the few cases where one provides a resource in more than one physical format (e.g. ShEx in JSON-LD and in RDF) it's not a big deal to describe it twice. Basically, this provides a level of indirection for a small number of cases that can be handled without it. (And BTW I would discourage people from keeping around the same data in multiple formats - that's a maintenance nightmare.)

I really like calling the resource a "supporting resource" and that resource would be a Thing - an artifact. 

I also do not think that we should assume that profile artifacts will be described with DCAT. In fact, we haven't so far seen any proposed metadata for describing profile artifacts beyond _role_ , _dct:format_, and _profileOf_. There's no title, no description, no license ... basically there is no metadata for the content of the artifacts themselves.

[...]

PR #596, opened 2018-11-19
Fixes Antoine's #15

larsgsvensson:
Updated examples two and three to use the 303 redirect pattern. This should fix number 15 in [Antoine's concerns](#575).
Visible in https://raw.githack.com/w3c/dxwg/larsgsvensson-antoine-%2315/conneg-by-ap/index.html#ex-2-using-a-link-header-to-point-to-a-list-of-available-profiles and https://raw.githack.com/w3c/dxwg/larsgsvensson-antoine-%2315/conneg-by-ap/index.html#ex-3-requesting-a-representation-conforming-to-a-specific-profile-using-http-headers

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -408,6 +408,18 @@ Accept: text/turtle
 
 ---
 
+HTTP/1.1 303 See Other
+Location: /a/resource.prof1.ttl
+[more response headers]
+
+===
+
+HEAD /a/resource.prof1.ttl
+Accept: text/turtle
+[more request headers]
+
+---
+
 HTTP/1.1 200 OK
 Content-Type: text/turtle
 Content-Location: http://example.org/a/resource.prof1.ttl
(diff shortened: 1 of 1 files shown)


aisaac (approved):
Thanks @larsgsvensson I think it alleviate my concerns.
This said it triggers a new question: in example 2 couldn't the Link header with the alternative representation appear as soon as the first 303 response from the server?

larsgsvensson:
@aisaac scripsit:
> This said it triggers a new question: in example 2 couldn't the Link header with the alternative representation appear as soon as the first 303 response from the server?

Perhaps it _could_ but I'm not sure how to relate the different representations to one another since the in the first request the `Context IRI` (in the sense of RFC 5988) denotes an abstract resource and not a representation. Do you have a suggestion?

aisaac:
I don't have a suggestion, otherwise I would have created an issue ;-)
More seriously (and perhaps worryingly) this may touch on the discussion at #573, in the sense of what counts as representation and what count as abstract resource. (this was the meta-meta discussion at F2F4 on applying to publications of profiles the recipes of profile negotiation, e.g. the 'SHACL profile of the profile'). Maybe we should just wait and see until this sort of discussion has settled down.

larsgsvensson:
@aisaac scripsit:
> Maybe we should just wait and see until this sort of discussion has settled down.

+1. Let's see what feedback we get.

PR #599, opened 2018-11-20
Change short names and explain scope better

rob-metalinkage:
I couldnt see any definition in the abstract, but I've reworded to make it more explicit about the scope.

### DIFF
--- a/conneg-by-ap/config.js
+++ b/conneg-by-ap/config.js
@@ -1,6 +1,6 @@
 var respecConfig = {
     specStatus: "ED",
-    shortName: "prof-conneg",
+    shortName: "dx-prof-conneg",
     edDraftURI: "https://w3c.github.io/dxwg/conneg-by-ap/",
     previousURI: "https://w3c.github.io/dxwg/conneg-by-ap/",
     testSuiteURI: "https://github.com/CSIRO-enviro-informatics/prof-conneg-jmeter-test-suite",
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -12,7 +12,7 @@
     <p>
       This document describes how Internet clients may negotiate for content provided by servers according to
       <em>profiles</em>. This is distinct from negotiating by Media Type or Language: the <em>profile</em> is expected
-      to specify the content of information returned, which may be a subset of what the responding server contains.
+      to specify the content of information returned, which may be a subset of the information the responding server has about the requested resource. Such profiles may be structured in a specific way to meet interoperability requirements of a community of practice.
     </p>
   </section>
   <section id="sotd">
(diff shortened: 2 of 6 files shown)


kcoyle (changes requested):
The definition of profiles ontology does not match my understanding of the ontology. The definition here says: "describes the general pattern of narrowing the scope of a specification" which to me isn't the case. It describes the components of a profile. It doesn't narrow anything and it isn't general - it is specific to profiles. Here's my view:

"The Profile Ontology allows one to describe a profile as a set of one or more resources that fulfill the functions of the profile such as vocabulary definition, input instructions, validation. The ontology also includes the capability to describe a profile's relationship to other files or profiles, such as 'profile of' another profile or vocabulary."

If possible, other changes included here could be allowed to merge, such as those that add "dx-" to the short names.

rob-metalinkage:
broke the branch. replacing

kcoyle:
Previous text:

The Profiles Ontology is an RDF vocabulary that describes the general pattern of narrowing the scope of a specification with additional, but consistent, constraints. In data exchange situations conformance to such profiles carries additional context about the interoperability expectations of any data. The Profiles Ontology enables profile descriptions to specify the role of resources related to data exchange such as schemas, ontologies, rules about use of controlled vocabularies, validation tools, and guidelines. The ontology may however be used to describe the role of artefacts in any situation where constraints are made on a the usage of more general specifications. 

My text:

The Profiles Ontology is an RDF vocabulary to describe the resources that define and implement a profile. These resources may be human-readable documents (PDFs, textual documents), vocabularies or ontologies (XSD, RDF),  resources specific to validation tools (SHACL, ShEx, Schematron), or any other files or resources that support the profile. Each resource is defined as having a role that defines its function within the profile.

The ontology also provides for the description of relationships between profiles and the standards to which they conform. Standards in this case can be vocabularies or it can be other profiles from which the described profile is a derivation, expansion, or selection. This ontology provides a standardized, machine readable formalism for describing the context of profiles. The basis of the ontology is a specialization of <code><a href="http://www.dublincore.org/documents/dcmi-terms/#terms-Standard">dct:Standard</a></code>, allowing the use of the <code><a href="http://www.dublincore.org/documents/dcmi-terms/#terms-conformsTo">dct:conformsTo</a></code> predicate to specify conformance to a profile.
</end kc text>


I would definitely like @draggett to review - not just this but the entire introduction (just a few paragraphs)

draggett:
@kcoyle writes:
> The Profiles Ontology is an RDF vocabulary to describe the resources that define and implement a profile.

which begs the question of what is the definition of the term "profile" in this context. I would like to see that given immediately before the definition of "profiles ontology". Essentially, we want to minimise the non-everyday terms we use, and never to use them without first defining what we mean by them in this context.

draggett:
Some minor readability tweaks to the second paragraph:
> This ontology also provides for the description of relationships between such profiles and the standards to which they conform. Standards in this case can be vocabularies or it can be other profiles from which the described profile is a derivation, expansion, or selection. This ontology provides a standardized, machine readable formalism for describing the context of profiles. The basis of the ontology is a specialization of dct:Standard, allowing the use of the dct:conformsTo predicate to specify conformance to a profile.

PR #605, opened 2018-11-22
Replaced broken copy edits and short name changes

rob-metalinkage:


nicholascar:
Editorial text changes only

PR #610, opened 2018-11-26
Move alignment stubs to annex

dr-shorthair:
Probably en route out of the rec document - 
- as discussed in recent telecons

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2785,46 +2785,6 @@ a:TestingActivity a prov:Activity;
 
     </section>
 
-    <section id="dcat-dats" class="informative">
-        <h3>DATS</h3>
-        <p class="issue" data-number="149">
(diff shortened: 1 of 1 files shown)


davebrowning (commented):
Good idea - we did talk about this, but without the detail.  Was your intent to insert this as Appendix A after the linked data stuff (Ch 14)?  In which case it needs a h3 header for the appendix title around line 3015.  Or were you suggesting we move Ch 14 into an annex as well?  I decided that I wasn't sure.....

dr-shorthair:
I fixed a couple of headers. 

For the time-being I left the schema.org and prov-o stubs in the body, and moved the domain alignments to the Annex. But in due course I could see shuflfing the sdo and prov-o into an annex and the others out of the document entirely.

davebrowning (approved):
Cool

PR #611, opened 2018-11-26
qualified attribution pattern

dr-shorthair:
New section for qualified attribution


dr-shorthair:
Preview here: https://rawgit.com/w3c/dxwg/dcat-issue79/dcat/index.html#qualified-attribution 

Probably need some richer documentation and perhaps examples.

nicholascar:
Could you step up the qualified pattern backgrounding one notch and mention Linked Data/graph qualified relations in general, on top of the PROV patterning, as per http://patterns.dataincubator.org/book/qualified-relation.html? This will indicate it's a general LD/graph mechanism, not just one used in relation to Agents & works thus allowing its consideration elsewhere.

dr-shorthair:
I have another change in prep for qualified dataset-dataset relations - not yet finished. That will be the time to add the general case for the qualified relations.

dr-shorthair:
I added an entry to the change-log for dcat:hadRole. 
I think this should be good to merge now.

agbeltran:
Added a mention to the new section about "Attribution roles" in the "Change history" part.

agbeltran (approved):

dr-shorthair:
I'm a bit wary of loading up a vocabulary description with too much theory. 
(IMHO the profiles ontology is veering in that direction - no examples until section 8. Really?!)

PR #651 has two consecutive sections using the qualified association pattern https://rawgit.com/w3c/dxwg/dcat-issue79-simon/dcat/index.html#qualified-attribution 
I guess these could be bundled into one, but I think it is best to keep the original requirements as the primary keys, and the solution as an emergent pattern.

Issue #81, opened 2018-01-18
Related datasets [RRDS]


jpullmann:
<section class='requirement' id='RRDS'><h3>Related datasets [<a href='https://www.w3.org/TR/dcat-ucr#RRDS' target='_blank'>RRDS</a>]</h3><p>Ability to represent the different relationships between datasets, including: versions of a dataset, collection of datasets, to describe their inclusion criteria and to define the 'hasPart'/'partOf' relationship, derivation, e.g. processed data that is derived from raw data</p><p class='note'>this requriement to be rolled in here: Update method of Dataset: Indicate the update method of a Dataset description, e.g. whether each new dataset entirely supercedes previous ones (is stand-alone), or whether there is a base dataset with files that effect updates to that base.</p><br/><em>Related requirements</em>: <a href='https://www.w3.org/TR/dcat-ucr#RID8' target='_blank'/> <a href='https://www.w3.org/TR/dcat-ucr#RPIF' target='_blank'>Provenance information [RPIF]</a> <br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID32' target='_blank'>Relationships between Datasets [ID32]</a> <a href='https://www.w3.org/TR/dcat-ucr#ID47' target='_blank'>Define update method [ID47]</a> </section>

dr-shorthair:
Comment received by [email from Nuno Friere](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Apr/0001.html) suggests using `void:rootResource` to point to a manifest.

dr-shorthair:
A general solution for relations from `dcat:Resource` to any other resource is discussed in #253 and a solution provided in #295. 

Guidance on use of more specific properties for relationships between` dcat:Datasets` should continue here. 
In particular considering 
(a) the more specialized DCT predicates `dct:hasPart/dct:isPartOf` , `dct:conformsTo` , `dct:isFormatOf/dct:hasFormat` , `dct:isVersionOf/dct:hasVersion` , `dct:replaces/dct:isReplacedBy` , `dct:references/dct:isReferencedBy` , `dct:requires/dct:isRequiredBy`
(b) the fully general case (possibly using qualified relation pattern and an extensible list of relationship roles) 

Should we add local (guarded) constraints on dcat:Dataset so that the range of these properties in this context is required to be `dcat:Dataset` - see discussion starting here: https://github.com/w3c/dxwg/issues/253#issuecomment-405170975

[...]

PR #613, opened 2018-11-29
Dcat issue79 datasets (don't merge yet)

dr-shorthair:
Started to describe dataset-dataset qualified pattern, in parallel with the dataset-agent pattern I already did. As suspected it raised more issues and more work is required to properly integrate it. 

Also a question of whether these PROV-O-based extensions belong in the main dcat.ttl graph or should be moved over to the dcat-prov alignment graph.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2024,6 +2024,10 @@
           </tbody>
         </table>
 
+        <p class="note">
+            Property added in this revision of DCAT.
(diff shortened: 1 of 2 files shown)


davebrowning:
@dr-shorthair - do you plan that this picks up #81 as well?  Or is that what your reference to :"more work to integrate it"?

Issue #68, opened 2018-01-18
Identifier type [RIDT]


jpullmann:
<section class='requirement' id='RIDT'><h3>Identifier type [<a href='https://www.w3.org/TR/dcat-ucr#RIDT' target='_blank'>RIDT</a>]</h3><p>Indicate type of identifier (e.g.<code>prism:doi, bibo:doi, ISBN</code>etc.).</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID11' target='_blank'>Modeling identifiers and making them actionable [ID11]</a> </section>

andrea-perego:
This is strictly related to #53.

agbeltran:
In DATS, we included an ```identifierSource``` associated with the ```identifier``` to report this information (see https://github.com/biocaddie/WG3-MetadataSpecifications/blob/master/json-schemas/identifier_info_schema.json)

andrea-perego:
> In DATS, we included an identifierSource associated with the identifier to report this information (see https://github.com/biocaddie/WG3-MetadataSpecifications/blob/master/json-schemas/identifier_info_schema.json)

Thanks, @agbeltran . ADMS does something similar, but, following [UN/CEFACT](https://www.w3.org/TR/vocab-adms/#bib-UNCEFACT) it distinguishes between identifier scheme, identifier scheme version, and the agency responsible for the identifier scheme - see:

https://www.w3.org/TR/vocab-adms/#identifier

dr-shorthair:
See https://github.com/w3c/dxwg/issues/53#issuecomment-365791669

agbeltran:
DataCite specifies a set of identifiers types in this XSD: https://schema.datacite.org/meta/kernel-4.1/include/datacite-relatedIdentifierType-v4.xsd

dr-shorthair:
Yeah - DataCite hide a lot of useful stuff inside XSDs. Shame they aren't even tabulated somewhere!

agbeltran:
Closing after merging #614

Issue #67, opened 2018-01-18
Primary and alternative identifier [RIDALT]


jpullmann:
<section class='requirement' id='RIDALT'><h3>Primary and alternative identifier [<a href='https://www.w3.org/TR/dcat-ucr#RIDALT' target='_blank'>RIDALT</a>]</h3><p>Provide means to distinguish the primary and alternative (legacy) identifiers.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID11' target='_blank'>Modeling identifiers and making them actionable [ID11]</a> </section>

andrea-perego:
Related to #53 & #68.

agbeltran:
Following from discussions in #53 and @riccardoAlbertoni's proposal in the [wiki](https://github.com/w3c/dxwg/wiki/DCAT-Identifiers), neither [Dublin Core](http://dublincore.org/documents/dcmi-terms/) nor [ADMS](https://www.w3.org/TR/vocab-adms/#identifier) have terms to represent alternative identifiers. 

In DATS we followed [DataCite](http://schema.datacite.org/meta/kernel-4.1/metadata.xsd) approach by having a representation of: [primary identifier](https://github.com/datatagsuite/schema/blob/master/identifier_info_schema.json), [alternate identifiers](https://github.com/datatagsuite/schema/blob/master/alternate_identifier_info_schema.json) and [related identifiers](https://github.com/datatagsuite/schema/blob/master/related_identifier_info_schema.json).

riccardoAlbertoni:
[Proposal 1 in the wiki](https://github.com/w3c/dxwg/wiki/DCAT-Identifiers#proposal-1)  suggests using 
- dct:identifier for primary id
- adms:identifier for secondary id

as far as I have understood, this was the guideline made by [DCAT-AP to manage duplicates](https://joinup.ec.europa.eu/release/dcat-ap-how-manage-duplicates), though I am not 100% sure this suggestion is still valid in the newest DCAT-AP release.  

@agbeltran Is this acceptable or Do you think we should add new specific terms in DCAT 2?

I guess @andrea-perego and @makxdekkers  might have their own view on the opportunities/issues behind the reuse of such a guideline.

makxdekkers:
That proposal sounds OK to me.

andrea-perego:
+1 from me as well.

agbeltran:
[...]

Issue #53, opened 2018-01-18
Dereferenceable identifiers [RDID]


jpullmann:
<section class='requirement' id='RDID'><h3>Dereferenceable identifiers [<a href='https://www.w3.org/TR/dcat-ucr#RDID' target='_blank'>RDID</a>]</h3><p>Encode identifiers as dereferenceable HTTP URIs</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID11' target='_blank'>Modeling identifiers and making them actionable [ID11]</a> </section>

makxdekkers:
I agree but I don't think there is a need to make any changes in the DCAT specification. It's an automatic consequence of DCAT being an RDF vocabulary.

dr-shorthair:
Yes - that was my impression too. No changes necessary to satisfy this requirement.

andrea-perego:
This issue is strictly related to providing guidance on how to use DCAT to specify identifiers as DOIs, ISBNs, etc. - [see the related use case](https://www.w3.org/TR/dcat-ucr#ID11).
Currently, these IDs are encoded as simple strings, unless they are used as part of the primary resource URI. An option could be to encourage the use of `owl:sameAs` whenever the ID can be resolvable when encoded as  URI (as for DOIs).
So, there may be no need to create new property / classes, but rather to describe how to use the existing ones to address these use cases.

makxdekkers:
It could be part of DCAT guidance, maybe in the usage note of https://www.w3.org/TR/vocab-dcat/#Property:dataset_identifier? In fact, the  current usage note only suggests that the "identifier might be used as part of the URI of the dataset" but it would be good to mention other identifiers in the usage note as well.

andrea-perego:
> It could be part of DCAT guidance, maybe in the usage note of https://www.w3.org/TR/vocab-dcat/#Property:dataset_identifier? In fact, the current usage note only suggests that the "identifier might be used as part of the URI of the dataset" but it would be good to mention other identifiers in the usage note as well.

+1 from me. One of options mentioned in the related use case is to use `dct:identifier` with a datatype denoting the identifier type (DOI, etc.). But these datatypes need to be defined. There's of course also the other option of using specific properties for each type of identifier (`prism:doi`, `bibo:doi`, etc.).
But for specifying multiple identifiers as HTTP URIs we need a property as `owl:sameAs`, which needs to be added to the DCAT spec.

agbeltran:
[...]

PR #614, opened 2018-12-05
Dcat issue53 riccardo

riccardoAlbertoni:
Dear all, 
I have prepared a new  branch adding a section for DCAT identifiers,  which is related to issues #53 #68 #67.

Please don't merge until the group has revised it.

The proposed new section distils [proposal 2 in the wiki](https://github.com/w3c/dxwg/wiki/DCAT-Identifiers#proposal-2-it-revises-and-reassembles-proposal-1) and considers some of the input received during the last dcat call ( [28-11-2018 meeting minutes ](https://www.w3.org/2018/11/28-dxwgdcat-minutes)).  In particular, 

- It does not explicitly mention primary and secondary IDs,
- it deals with IDs in general not only for datasets, 
- it only refers the dcat-ap guidelines about dataset duplications.

I suggest to accept this section as a first attempt to solve the aforementioned issues, and then change/ drop any part the group might want. 

 I'd invite @agbeltran and @makxdekkers, who have made suggestions during the last call, to chime in and change the pull if I have not satisfactorily  embodied their suggestions.

Cheers, 
Riccardo


agbeltran:
This is great @riccardoAlbertoni, thanks. 

Dropping here the link to visualise the new section:

https://rawgit.com/w3c/dxwg/dcat-issue53-riccardo/dcat/index.html#Dereferenceable-identifiers

and the differences with the Editor Draft:

https://services.w3.org/htmldiff?doc1=http%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fdxwg%2Fdcat-issue53-riccardo%2Fdcat%2Findex.html

smrgeoinfo:
thanks @riccardoAlbertoni, looks good to me.

makxdekkers:
The new text states: "_The need to clearly distinguish between primary and legacy identifiers for a dataset has been identified as a requirement to be satisfied in the revision of DCAT_". 
I am not entirely in agreement with that.
I would say that, yes, there is a need for such a distinction in particular cases, i.e. in harvesting environments where descriptions are copied among catalogues. This case is mentioned in the last sentence of the section: "_Depending on the application context, specific guidelines such as "DCAT-AP: How to manage duplicates?" can be adopted for distinguishing authoritative datasets from dataset harvested by third parties catalogs_".
In my mind, the notion of primary and secondary (not **_legacy_**, I think) identifiers depends very much on the application environment. This requirement could be satisfied in a profile (as it is done on the EU DCAT-AP) and not in the base standard.

agbeltran:
@makxdekkers we do have the requirement for primary and alternative identifier (see https://github.com/w3c/dxwg/issues/67) - if we delegate the solution to a profile, it is something we should clarify in the document as a way of addressing that requirement

agbeltran (changes requested):
A comment on the DOI example. Instead of using a DOI that corresponds to a publication, can we use a DOI for a dataset? If the recommendation is to use owl:sameAs for the URI corresponding to the textual DOI, I suggest removing the part of the example that uses skos:notation

agbeltran:
A comment on the DOI example. Instead of using a DOI that corresponds to a publication, can we use a DOI for a dataset? If the recommendation is to use owl:sameAs for the URI corresponding to the textual DOI, I suggest removing the part of the example that uses skos:notation

e.g. we could use https://doi.org/10.5281/zenodo.1486279

we could consider the schema.org version that they expose as a starting point:

```curl -LH "Accept: application/vnd.schemaorg.ld+json" https://doi.org/10.5281/zenodo.1486279```

riccardoAlbertoni:
@agbeltran  I have just changed the example including the zenodo doi instead of ieee doi.

@ agbeltran wrote
> If the recommendation is to use owl:sameAs for the URI corresponding to the textual DOI, I suggest removing the part of the example that uses skos:notation

I suspect that some redudancy might help here.  In my view the owl:sameAs is an additional thing to say, which helps in retrieving more info about IDs.

davebrowning:
Can I nudge this a little - I'd hoped we could discuss last night but we ran out of time.

@agbeltran : looks like @riccardoAlbertoni has addressed your review?

Looking at the thread in #67, it seems to have reached consensus that this contribution addresses the requirement (albeit by pointing out its actually context specific).  The one remaining thing would be to  use the phrase `alternative (legacy) identifiers` to bring it into line with the text in the [requirement](https://github.com/w3c/dxwg/issues/67)

riccardoAlbertoni:
I have added the "use the phrase alternative (legacy) identifiers",  I suggest marking the issues #67 as due-for-closing ( as done by nick for issues in other documents) or if you prefer, closing the issue straight away.
If you are going to close, let me know that I will get rid of issue box in the document, copying its description in the text. 

 In both cases, please mention in the issue thread  the agreed conclusion, for example saying, "the group has agreed that  distinguishing between primary, secondary (alternative) identifiers it is very much application-specific and should be better addressed in application profiles rather than being mandate a general approach."

davebrowning:
Thanks @riccardoAlbertoni, that's very helpful.  I'll ask the question about WG membership as a pre-req for being mentioned.  [Though in my quick survey I did see some examples where the contributors were not on the membership list..... but best to be sure.]

davebrowning:
Thanks @riccardoAlbertoni - I have marked #67 as due for closing with the appropriate (provisional) conclusion.  Suggest we wait for a day or two to give people a chance to respond (ie disagree).

agbeltran (approved):

PR #618, opened 2018-12-11
draft DCAT 2018 to ISO19115-3 and 19139 mapping

smrgeoinfo:
excel spread sheet with proposed mapping from DCAT 2018 elements to ISO 19139 (19115) and 19115-3 (19115-1) mapping


dr-shorthair:
I see no problem merging this - it is merely a non-normative resource to support work that might find its way into the Rec.

dr-shorthair:

andrea-perego:
@dr-shorthair , @smrgeoinfo , it's a bit unclear to me the purpose of this. 

GeoDCAT-AP already defines alignments between DCAT and ISO 19139 (ISO 19115 / ISO 19119). It has been defined by a working group, along with a reference XSLT implementation of the mappings, and it is adopted and implemented at European level. 

So, why should we start re-doing the job?

dr-shorthair:
I saw it as mostly harmless to make it available in the GitHub repo. Also, the spreadsheet does include mappings for most of the new elements introduced in the DCAT revision: `prov:wasAttributedTo`, `dct:hasPart`, `dcat:service`, `dcat:accessService`, `dct:endpointURL`, `dcat:endpointDescription`, `dcat:servesDataset`. 

@smrgeoinfo can you explain the provenance of this set of mappings?

smrgeoinfo:
It's my own work, based on the mappings I've done for IEDA and EarthCube catalogs.  Next step would be to compare with the DCAT-AP XSLT, but I haven't gotten to that yet (it fails if identifiers are not HTTP URIs). It also includes mapping to 19115-3, and addresses some of the variations that are commonly seen in ISO metadata.  Also it's intended to provide guidance for both DCAT to ISO and ISO to DCAT.  

https://github.com/SEMICeu/iso-19139-to-dcat-ap/blob/master/documentation/Mappings.md does not provide xpaths for the ISO elements.  The DCAT-AP work also appears to be very INSPIRE focused.

andrea-perego:
Thanks for the clarification, @smrgeoinfo .

Indeed, the documentation on GeoDCAT-AP mappings does not include XPath expressions. They can actually be derived from the examples in the GeoDCAT-AP specification (Annex B), where the mapping of each of the ISO elements covered by GeoDCAT-AP are explained. Besides the PDF version of the GeoDCAT-AP spec, this can be probably more easily seen in the "live" HTML version:

https://semiceu.github.io/GeoDCAT-AP/drafts/latest/#detailed-usage-notes-and-examples

Said that, the problem is that in many cases the transformation cannot be carried out only with XPath expressions. An example is the value of `@xml:lang` for plain literals, which needs to be inherited by the ISO element for the metadata language. And another one is about the cases where values in ISO metadata needs to be mapped to other ones in the RDF representation (e.g., the resource language, that should be expressed with a URI from a given code list, instead of the ISO code).

About the scope of GeoDCAT-AP, this programmatically covers the union of the metadata elements in INSPIRE and the core profile of ISO 19115:2003 - the full list of covered elements is here:

https://semiceu.github.io/GeoDCAT-AP/drafts/latest/#overview-of-metadata-elements-covered-by-geodcat-ap

So, it is indeed not complete, but I would not say that the mappings are INSPIRE-specific. The only "specific" aspect concerns the code lists used for languages, etc., which (following DCAT-AP) are those maintained by the EU Publications Office and the INSPIRE Registry. These can be easily replaced in the transformation by other reference code lists - as the LoC ones.

It makes perfect sense to me to complement what not included in GeoDCAT-AP, and to look also at ISO 19115-3, but again it would make sense to build upon the mappings already defined in GeoDCAT-AP. BTW, about ISO 19115-3, the GeoDCAT-AP WG did a preliminary analysis of what was changed, which may turn to be useful. It is included in an annex to the GeoDCAT-AP spec:

https://semiceu.github.io/GeoDCAT-AP/drafts/latest/#comparison-between-inspire-and-iso19115-12014

PR #624, opened 2018-12-14
added requirements from UCR document

kcoyle:
This needs a lot more cleanup - I didn't remove much, and I do have some questions about content and placement.


nicholascar:
Good additions - brings the real (agreed to) Reqs into the doc

PR #627, opened 2018-12-15
Nick moved from author to editor

nicholascar:
Nick moved from Author to Editor


aisaac:
Great to see the clarification about editors :-)

PR #628, opened 2018-12-17
cleaned up with requirements

kcoyle:
next steps: create new section headers based on each requirement; compare to DoW best practices to see what else needs to be added


aisaac:
I'm going to merge it but based on the assumption that this is work in progress and you are going to fix the structure (which was already broken by the previous PR) :-)
For the record almost everything is not merged in one big section 2:
https://rawgit.com/w3c/dxwg/kc_branch/profiles/index.html

kcoyle:
Thanks, Antoine. I'll fix the heading levels in the next version. Turns out I can't see that in my local copy, so it may take some iteration.

Issue #649, opened 2019-01-10
Clarify scope of DCAT and potential for extensions


dr-shorthair:
A number of issues (#64 #172 #432 #530) relate to the scope of DCAT. 
Public comments suggest that the extension to support data-services is either too much (https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Oct/0002.html) or not enough (https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Oct/0003.html). 
The matter was discussed in the DCAT team meeting 2019-01-09 and it was [agreed to add text into the Rec in various places](https://www.w3.org/2019/01/09-dxwgdcat-minutes#x08) to 
1. clarify that the scope of DCAT includes Datasets and DataServices
2. note that dcat:Resource provides an extension point for further sub-classes if desired, so that catalogues of other things can use the same DCAT superstructure. 

Doing this should allow several issues to be closed.

Issue #530, opened 2018-11-04
Data catalogs should not treat services as first class citizens


davebrowning:
(Issue created to track comments received on Second PWD of DCAT Recommendation from Clemens Portele - email archived [here](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Oct/0002.html))

The rationale for making services first class citizens in data catalogs is not clear to me. The model of DCAT 1.0 is in this area, in my view, clearer and more useful from an end user perspective. The items registered in a *data* catalog should be restricted to data (that is: datasets - unless you register individual items). As a user I go to a data catalog looking for datasets, not services; once I have found some candidate datasets then I want to understand how to access the data, whether that is a file download, through webpages or an API. This relationship is described in the Data on the Web Best Practices (https://www.w3.org/TR/dwbp/#context), too.

While DCAT 1.0 does not provide sufficient detail or guidance for documenting the different kinds of distributions. A downloadable file is straightforward, but an API could benefit from more information than what DCAT 1.0 supports. From the UCR document I expected that the revision of DCAT would improve Distribution instead of adding Services as additional, separate resources in the mix. The separation between Distribution and Service looks blurred to me, with overlap and redundancies in their properties. Please reconsider this design decision.

In my view, the current direction is repeating some of the mistakes that the spatial data community made in their standards.

agbeltran:
Including here an extract of other comment received from Luca Trani (message available [here](https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Oct/0003.html)) with reference to data services and more clarifications needed:

"In particular, the introduction of additional resources (beyond Dataset) as first-class citizens in a Catalogue was definitely required and we accommodated that by extending DCAT 2014. My view is that in the context of RIs catalogues have an essential role to represent a variety of assets -- having the focus only on Dataset would be too restrictive. Therefore, I am pleased to see that there is a dcat:Resource in this new version. Also, two specialisations of a dcat:Resource are provided: Dataset and DataService. I assume that more might be added to extend the scope of a Catalogue when needed. Am I correct? Do you support composition of resources? Would that be covered by dct:relation?
Also, I would be interested in the approach you are considering to tackle Issue 66 and Issue 71.

[...]

PR #650, opened 2019-01-10
Additional text on scope of DCAT

dr-shorthair:
Dealing with #649 
Potentially also allowing #64, #530 to be resolved:

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -142,7 +142,6 @@
     <ul>
         <li>
           <code><a href="#Class:Catalog">dcat:Catalog</a></code> represents a catalog, which is a dataset in which each individual item is a metadata record describing some resource; the scope of dcat:catalog is collections of metadata about <b>datasets</b> or <b>data services</b>.
-          The resources described in a single catalog are published or curated by a single agent.
         </li>
         <li>
           <code><a href="#Class:Resource">dcat:Resource</a></code> represents an individual item description in a catalog.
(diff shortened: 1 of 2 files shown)


akuckartz (approved):

akuckartz (approved):

akuckartz (approved):

akuckartz (approved):

davebrowning (commented):
Good step forward - the substance seems fine to me.

Some small things - I could make these changes if you agree/like

1. In DCAT scope this says "The resources described in a single catalog are published or curated by a single agent."  I find the first single a bit strange/redundant - isn't the case that any resource has to be published/curated by a single agent, no matter which catalog it's been catalogued in? 
2. The phrase "DCAT profile or application" is used a few times - I'd have said "DCAT or application profile' since these are all cases of profiles.  That would sit neater with the language in the profiles guidance, to my eye?
3.  There are a few references to dcat:Resource that aren't hyperlinked (And some old ones that are marked as `<code>` but not hyperlinked under the CatalogRecord paragraph)

dr-shorthair:
1. This was copied over from the original. I tried to move it to a better place, but did not change the wording. But I am also not sure if it is right or useful. @makxdekkers might remember why it is there. 

2. Yeah - I was trying to capture the idea that there will be DCAT (application-)profiles (which describe a static vocabulary or schema) but also DCAT applications (which use DCAT to build a discipline- or community-specific application). Maybe a global search and replace "DCAT profile or application" with "DCAT application profile or other DCAT application" ? 

3. would be good to fix those ...

davebrowning:
I've addressed my observations (2) & (3) in this branch.

agbeltran (approved):

makxdekkers:
@davebrowning @dr-shorthair On point 1: no I don't remember why it was there, and I agree it's not a useful formulation.

agbeltran:
I removed the sentence 'The resources described in a single catalog are published or curated by a single agent.' and will merge now, as all the comments have been addressed.

PR #657, opened 2019-01-14
Ucr profile requirements

jpullmann:
Status ready for 2.WD publication


aisaac:
This looks good to me but I have not had the time for a complete review. @jpullmann do you think this is needed?

jpullmann:
@aisaac Thanks, I do not expect objections to content, since already shared with the group for review. The static page passed Pubrules and the W3C checker. Will merge and pass over to @draggett

Issue #641, opened 2019-01-07
Remove prof:BaseSpecification


andrea-perego:
This issue has been already raised in https://github.com/w3c/dxwg/issues/404#issuecomment-438253648 , and partially discussed while preparing the ESWC paper.

The [definition](https://w3c.github.io/dxwg/profilesont/#Class:BaseSpecification) of this class is as follows:

> A specification that defines all implementation constraints, without applying constraints on usage of another specification

IMO, this is an abstract scenario (is there any spec not using somehow another one?). And it is also said in the usage note:

> This may not be a useful class: documents of any specification can be regarded as a trivial profile, so applications only need to be concerned with Profile conformance.

Moreover, its use in the PROF spec may lead to confusion. E.g.:
- The definition above implies that `prof:BaseSpecification`, but then it is defined as a subclass of `prof:Profile` (a particular type of profile with an empty set of constraints)
- In [Figure 3](https://w3c.github.io/dxwg/profilesont/#fig-eg-dcat), DCAT is said to be a `prof:BaseSpecification`. Is this correct, considering that DCAT re-uses classes and properties from other vocabularies?

Based on the considerations above, I would be in favour of dropping `prof:BaseSpecification` from PROF.

nicholascar:
There may be use in having a Base Spec class for *relative* base-ness. I know this seems counter intuitive at first but:

From the point of view of DCAT-AP, DCAT may indeed be it's Base Spec. DCAT-AP may wish you to know about DCAT but not really about DC, OWL etc.

kcoyle:
What is the functional use case for base specification? What does it tell you that you need to know and cannot know from the profile itself? Is the need for machines or for human users?

If there is no defined action that results from base specification then I agree with @andrea-perego that we should drop it. However, I also note that there are no properties in the ontology for descriptions, documentation, etc. It would seem to me that a profile should have at least the possibility of human-readable documentation, and that documentation could explain what key vocabularies or profiles the profile is based on, if the perceived need is for human understanding not machine processing. Admittedly, one or more of the resources of a profile may provide documentation, but that is at a resource level, not a profile level. For example, there may be documentation that looks like the current DCAP-AP, but that doesn't explain the relationship between DCAT-AP and the RDF file. Documentation at the inclusive level seems to be an obvious need.

andrea-perego:
@nicholascar said:

> There may be use in having a Base Spec class for relative base-ness. I know this seems counter intuitive at first but:
>
[...]

Issue #275, opened 2018-06-27
Profiles are uniquely named sets of data constraints, such as data elements (e.g. classes, properties, value domains) that describe (meta)data.


nicholascar:
Entered from Google Doc

aisaac:
Could this be reworded as "data" or "(meta)data"? It's perhaps better to make it more general.

kcoyle:
The issue here was the use of "properties" and whether that applies only to RDF. So "or metadata terms" was (awkwardly) added to show that solutions other than RDF are included. Any wording that you can come up with that gets that across would be acceptable to me.

aisaac:
ok I won't add to the currently ongoing work by suggesting new wording now, but this is worth keeping in mind, thanks.

jpullmann:
Please consider rewording in more requirement-oriented , imperative terms of RFC 2119, e.g. 
 "Profiles may/should comprise a named collection of properties or metadata terms, if not specified by RDF."

kcoyle:
I'd prefer:

Profiles should comprise a collection of metadata terms
(The "if not RDF" was added because of the use of "properties" but I think that "metadata terms" covers both.) I think "comprise" is a difficult word and will try to think of something better.

(addding:)
Profiles should be made up of a collection of metadata terms that have been defined in published vocabularies.

This matches the DCMI definition, although as I recall Annette thought that profiles should be able to define new terms.

aisaac:
Actually I would argue against 'comprise' in @jpullmann 's suggestion because it feels strange to say that profile comprises a collection with a name, while the original idea is that the name of the collection is the name of the profile itself.
@kcoyle 's rewording fixes this, but then the name is gone!

Then the fact that the elements of a profile are from published vocabularies is probably relevant but I would prefer to rely on other existing requirements to carry this instead of extending the scope of this one. E.g. we could piggyback on the one that says that profiles are based on existing specifications. Such a requirement is probably a better fit, as profiling a specification is probably nonsensical if that specification has not been published anywhere.

And talking about @agreiner I guess she may argue against restricting the scope to 'metadata' - she has made a comment about this for the intro of the Profile guidance ;-)

kcoyle:
@aisaac I'm open to your suggestions - maybe:

Profiles should made up of a collection of properties or terms, and the collection should have a name. 

The problem is that it is a named collection but it can't be "made up of a named collection" (because it can't be made up of a single thing, a named collection) although it can be that they "are a named collection".

[...]

PR #664, opened 2019-01-15
removal of all references to Base Specification class

nicholascar:
As per Issue <https://github.com/w3c/dxwg/issues/641>


aisaac (changes requested):
I disagee with all references to "base specification" being removed. I am in favour of removing all the references to the class, but it seems this PR goes further.
For example, this PR is changing "base specification" to "standard" in the definition "A named set of constraints on one or more identified base specifications or other profiles". I think that occurrence of "base specification" was alright, and there are probably others.

nicholascar:
Can we put this through and then check against the result that we cover what we need to and improve if not? I'm keen to ensure we purge properly as we have a lot of terminology tangles so that's why I removed occurrences of the phrase "base specification" as well as the actual class. Whether we can cover "base specification" use with "standard" is a good thing to test in another review round.

aisaac:
Hi @nicholascar I would rather avoid any change that would amount to proposing an update to the wording of the agreed definition, and 'base specification' is one of these words. 
First, regarding the matter itself I think 'base specification' is a matter of the role played by another specification rather than a strong statement about the type of resource and I think this is appropriate.
Second, on the meta-level, it avoids another round of terminology discussion. If just because if we remove "specification" and keep only "standard", then we will probably have to re-open the discussion on our understanding of "standard". And honestly the experience of the past week on #275 makes me shiver at the idea of opening an another front.

If you are uncomfortable deciding which "base specification" should be kept, then I'm happy to go through the document and make a proposal.

nicholascar:
Sorry, still not comfortable with the partial retention of the phrase "base specification". The phrase in text is inevitably linked in the mind of the reader to the Base Specification class definition. If we remove the class, we would, at the very least, need to re-examine all uses of "base specification" but, my preference, is to just remove the term altogether and then, on a second reading of the spec without both Base Specification and "base specification" determine appropriate wording.

Can we just use "dct:Standard" as a replacement for "base specification" initially, as per the PR and the reevaluate in a second phase? This is, in my opinion, a cleaner progression.

aisaac:
I'm sorry @nicholascar but I'm not buying the first part of this. If we remove the class BaseSpecification, then how can the reader connect any occurrence of the phrase 'base specification' to a class they can't see?

I think replacing by Standard everywhere is going to resume again all the discussions on standards, and we can't afford this. Again I'm happy making a proposal for changed text. I could even try to replace 'base specification' by 'grounding specification', or 'specification that base the profile' whatever keeps 'specification' in (and the fact that this is a specification the profile is a profile of)

aisaac:
@nicholascar 
In fact I'm not so much against your suggestion of 'remove the term altogether and then, on a second reading of the spec without both Base Specification and "base specification" determine appropriate wording'. It's just that I'm really uncomfortable using "standard" as replacement now, because this will open discussions that will just make us forget what we wanted to do first, and thus lose time.

I'm tempted to suggest a horrible trick that probably only editors can decide on :-)
How about replacing all the remaining occurrences of "base specification" by an artificial token like "[base specification/standard]" (with the brackets) or even "BSS"? This would mark the places we need to discuss in a more neutral way.

nicholascar:
@aisaac: Rob & I discussed this following the meeting just gone. We've agreed to reinstate the text "base specification" as you originally asked for so I'll amend this PR with an extra commit just now to do that.

nicholascar:
textual "base specification" re-added, I think Rob's good to go approving this PR which now addresses @aisaac concerns! Thanks all, I concede that this is now better with those words left in.

aisaac:
Great, thanks, @nicholascar !

PR #669, opened 2019-01-16
Editorial revisions

andrea-perego:
- Fixed typo ("statementss") as per @kcoyle 's email: http://lists.w3.org/Archives/Public/public-dxwg-wg/2019Jan/0195.html
- Added CODE HTML tags to qualified names of classes and properties used in the running text

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -45,7 +45,7 @@
     <p>This revised version of DCAT was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[DCAT-UCR]] submitted on the basis of experience with the DCAT vocabulary from the time of the original version, and new applications not originally considered. A summary of the changes from  [[VOCAB-DCAT-20140116]] can be found at <a href="#changes">Change History</a></p>
 
     <h3 id="external_terms">External terms</h3>
-    <p>DCAT incorporates terms from pre-existing vocabularies where stable terms with appropriate meanings could be found, such as foaf:homepage and dct:title.  Informal summary definitions of the externally-defined terms are included here for convenience, while authoritative definitions are available in the normative references.  Changes to definitions in the references, if any, supersede the summaries given in this specification.  Note that conformance to DCAT (Section 4) concerns usage of only the terms in the DCAT namespace itself, so possible changes to the external definitions will not affect the conformance of DCAT implementations.</p>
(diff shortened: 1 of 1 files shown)


agbeltran:
thanks @andrea-perego - I was checking the changes in [rawgit](https://rawgit.com/w3c/dxwg/andrea-perego-dcat-ed-rev/dcat/index.html) and in previous commits, we actually replaced the CODE HTML instances for URLs pointing to the term definitions everywhere in previous editorial changes.  Pinging @davebrowning and @dr-shorthair to double-check.

dr-shorthair:
There was some inconsistency, but I thought @davebrowning had cleaned it up. 

Should the mentions in running text be distinguished using a typeface (\<code\>) or by making them into links? 

And if the latter, then we need to be consistent. Here's my proposal: 

- within chapter 6, the linked text in the slot next to **RDF Class:** or **RDF Property** heading in each table should point to the W3C /ns/ or to the external reference (FOAF and DC)
- all other cross-references within the document, including within chapter 6, should point to the relevant section/fragment within Chapter 6

davebrowning:
Last week I "fixed" the missing links in the sections that were being changed at the time (and added  `<code>` tags where I was changing something, but not a general sweep of across the document.   I suspected then that we'd got somewhat inconsistent and needed to look across the whole but didn't get a chance to look at the rest of the doc on both `<code>` and link usage.  Thanks for doing that @andrea-perego !  Adding the `<code>` markups do make them stand out.

 The 2014 spec also used /ns/ references in **Domain:** and **Range** entries, but I think that's not particularly helpful.  (**See Also** consistently has intra-document links).  On a quick look at the editors draft there are probably a few other stray references to the /ns/ in other parts  of the text, I think (such as some of the intro paragraphs for each class.) My question is - are links to either /ns/ (essentially the ttl file) ever really useful ?  It only ever returns the whole file (for me, anyway) whereas a link into a _relevant_ fragment of a doc is much more useful.  For me, I'd rather we used useful links where we can find them.  What do other people think?  [For what it's worth I see multiple different behaviours across recommendations]

To keep it simple though, I agree with @dr-shorthair's proposal that we use internal document links in all cases except the headline **RDF Class:** or **RDF Property:** texts in the subsections of Chapter 6, mainly because there's little else to link it too....  (And of course, mark them as `<code>` when appropriate) 

I'd propose we merge this branch, and then do a sweep to add the links.  I'm happy to do that, though it'll be next week before i have the time - I'll raise a (presumptive) issue so we don't forget...

Issue #670, opened 2019-01-17
'smushing'


dr-shorthair:
Respond to comment from @smrgeoinfo - https://lists.w3.org/Archives/Public/public-dxwg-comments/2018Nov/0007.html

This text was copied from DCAT-2014: 

> foaf:homepage is an inverse functional property (IFP) which means that it SHOULD be unique and precisely identify the catalog. This allows smushing various descriptions of the catalog when different URIs are used.

The point is that the `foaf:homepage` is the _canonical_ webpage for a Catalog. This is implied by the fact that `foaf:homepage` [is defined (in FOAF)](http://xmlns.com/foaf/spec/#term_homepage) as an **inverse-functional-property**. The 'smushing' text is just a commentary or interpretation of the implication of IFP in the DCAT context, as there might be more than one web-page for a `dcat:Catalog` (or any other `dcat:Resource` for that matter). 

I propose adjusting the usage note to the following: 

> foaf:homepage is an inverse functional property (IFP) which means that it SHOULD be unique and precisely identify the web-page for the catalog. This property indicates the _canonical_ web-page, which might be helpful in cases where there is more than one web-page about the catalog.

akuckartz:
SHOULD or MUST ?

smrgeoinfo:
If the constraint is MUST, is there any way to test conformance?

I like the proposed language.

agbeltran:
@smrgeoinfo a reasoner will infer that if there are two subjects for the inverse functional property, they are the same and raise a conflict if that is not possible due to other axioms

PR #672, opened 2019-01-18
DCAT issue #670 - Eliminate 'smushing'.

dr-shorthair:
Responding to #670


agbeltran:
While the text now says MUST, the turtle file still reads SHOULD. I will make that change and then merge.

agbeltran (approved):

PR #682, opened 2019-01-23
Align document with RDF content

davebrowning:
In looking across recent merges, a couple of inconsistencies had crept in between dcat.ttl and the document.


davebrowning:
We also appear to have two sections in the document defining the `dcat:hadRole` property.  (See [here](https://rawgit.com/w3c/dxwg/dcat-minor-rdf-doc-alignments-daveb/dcat/index.html#Property:hadRole) and the preceding 6.14.1.1)

Should these usage notes be coalesced or does one take priority?

dr-shorthair:
I think this is good now.

dr-shorthair (approved):

PR #694, opened 2019-01-25
Removed entries now in Specref

andrea-perego:
See https://github.com/tobie/specref/pull/501


nicholascar:
Thanks Andrea; I will apply this update to the Ontology & Conneg docs too as they also contain those refs.

PR #705, opened 2019-01-28
Polish translation

domel:
Polish translation of Profile Descriptions Ontology


deniak:

deniak:
Marked as  substantive for IPR from ash-nazg.

rob-metalinkage:
definitions all updated so sadly this will need to be revisited.

PR #708, opened 2019-01-29
Paper reviews

nicholascar:
Multiple edits made as a result of following ESWC reviews of the Profiles Ontology paper.


rob-metalinkage:
These are small editorial fixes. No substantive changes.

Issue #710, opened 2019-01-29
 Rephrase/Extend the opening of DCAT section about dereferenceable identifiers  


riccardoAlbertoni:
in the [mail](https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Jan/0195.html), @kcoyle  referring to the section on 'De-referenceable identifier' wrote

>I don't think this is the best wording to open this section: "DCAT
should rely on persistent HTTP URIs,..." I suspect what is mean is that
when identifiers are used in DCAT they should be HTTP URIs. But that
also is unclear because the DCAT RDF vocabulary by its nature must use
HTTP URIs for all "things". So something else is needed to explain what
the optional situations are where a choice can be made. I'm afraid it
wasn't clear to me what situations the section was referring to.

This is related to #675

riccardoAlbertoni:
Taking inspiration from the [use case 11](https://www.w3.org/TR/dcat-ucr/#ID11) and the description   provided by @andrea-perego  from which the identifier requirements are originated,  I have added the following few sentences at the beginning of the identifier section.  (see the branch https://github.com/w3c/dxwg/tree/dcat-issue710-riccardo)
 
>"The scientific and data provider communities use a number of different identifiers for publications, authors and data.

>DCAT primarily relies on persistent HTTP URIs as an effective way of making identifiers actionable. Notably, quite a few identifier schemes can be encoded as dereferenceable HTTP URIs, and some of them are also returning machine-readable metadata (e.g., DOIs, ORCIDs). Regardless, data providers still might need to refer to legacy identifiers, non-HTTP dereferenceable identifiers, locally minted or third-party-provided identifiers. In these cases, [DC11] and [VOCAB-ADMS] can be of use."

@kcoyle does this address your concers?

riccardoAlbertoni:
@kcoyle  approved my pull request and @davebrowning has just merged it. I suggest the DCAT editors to label this issue as "due for closing" or  even closing it straight away.

PR #711, opened 2019-01-29
adding an opening sentence for the section on Dereferenceable Identifiers

riccardoAlbertoni:
related to issue #710


kcoyle (approved):
this is much improved. I think I would combine them into a single paragraph but that's a minor editorial change and shouldn't hold this up.

davebrowning:

Issue #700, opened 2019-01-27
UCR: Re-phrase description of UC26


andrea-perego:
Ref: https://w3c.github.io/dxwg/ucr/#ID26

The UC description currently reads as follows:

> Considering DCAT a high-level model for data exchange agree on significant aspects missing so far and define extension points (typically properties) for re-use and integration of existing standards, application profiles etc. The reference listing of aspects deemed relevant is based on evaluation of DXWG use cases and [ISO-19115-1]:

@jpullmann , the first sentence sounds wrong (something missing?), and should be re-phrased. A minimal revision could be (please note also the added commas):

> Considering DCAT a high-level model for data exchange,  **it would be important to** agree on significant aspects missing so far and define extension points (typically properties) for re-use and integration of existing standards, application profiles, etc.

jpullmann:
@andrea-perego  This sentence reads like a requirement (imperative), will update in suggested way, thanks!

andrea-perego:
Thanks, @jpullmann !

I wonder whether we should add a milestone for the 3rd PWD of the UCR document...

jpullmann:
resolved by commit https://github.com/w3c/dxwg/commit/4aa3a09b5530be4df18be3566e82754a00a2f7d7

Issue #683, opened 2019-01-23
UCR: Editorial issues


andrea-perego:
While reviewing the current version of the UCR document, I saw some editorial issues that needs to be addressed. I include here the ones I found:

1. The are still in some places the name of the person who added a comment - see, e.g., [Section 5.32](https://www.w3.org/TR/dcat-ucr/#ID32):

    > (Makx Dekkers) Specific cases of relationships that I have come across:
    >
    > - a dataset that contains multi-annual budget data (e.g. for a multi-annual programme) but also contains the data for individual years -- this could be as a spreadsheet with worksheets for each year and a sheet with the sum for the whole period
    > - two datasets that contain the same data but differ in other aspects than format, for example currency, measurement units, resolution, projection -- if we adopt an understanding that distributions of a dataset may only differ in format

    I guess this needs to be fixed

2. The cited standards and specification are sometimes cited other times directly linked to. The result is that the list of references is not comprehensive

3. All references are said to be normative - see https://w3c.github.io/dxwg/ucr/#references. Is this correct?

I've started fixing points (1) & (2), and I'll create shortly a PR with a proposed revision

andrea-perego:
See also #700 & #701

andrea-perego:
I partially addressed the issues listed here via PR #684 . 

I include below an update of what done so far:

- > 2 The cited standards and specification are sometimes cited other times directly linked to. The result is that the list of references is not comprehensive

  This is more or less fixed in PR #684, although I may have missed some

- Issue #701 : Fixed

- > 1 The are still in some places the name of the person who added a comment [...]

  I marked as EDITOR's NOTEs all these, plus those NOTEs that seemed more comments for the WG rather than explanations for the reader. This way they could be easily identified and reviewed, to see if any action needs to be taken

- > 3 All references are said to be normative - see https://w3c.github.io/dxwg/ucr/#references. Is this correct?

  Yet to be fixed.

- Issue #700 : Yet to be fixed

In addition to this, I made further editorial changes, all summarised in #684 

Unless there's any objection, I would propose merging PR #684 and open separate issues for points (1) and (3) above.

@w3c/dxwg-editors  , WDYT?

aisaac:
@andrea-perego thanks a lot for the review! I agree with creating issues for (1) and (3) and keeping names with notes for the moment.

jpullmann:
@andrea-perego thanks for a careful review!
[...]

PR #712, opened 2019-01-29
Minor editorial changes

jpullmann:
As suuggested by @andrea-perego following updates were made: 

- Rephrased UC26 as of issue #700 
- Spec references were enclosed into an "informative" context  as of issue #683


aisaac (changes requested):
Thanks @jpullmann .
I'm assuming that this fixes the issue of undesired references without waiting for a possible change in reSpec. Maybe it's not the ideal solution, but it seems actually alright.
On the minor changes, there's now a problem with a sentence "This information is  Lars G. Svensson and Phil Archer.normally"

andrea-perego (approved):
I made a commit addressing @aisaac 's comment, plus a minor fix.

BTW, I raised the issue about normative/informative refs in ReSpec, and a fix in under-way - see https://github.com/w3c/respec/issues/1754#issuecomment-458749170

After that fix will be implemented, we won't need `class="informative"` any longer.

jpullmann:
thanks @aisaac for the issue report and @andrea-perego for handling!

Issue #718, opened 2019-01-30
Remove inherited properties from the dcat:Catalog section


agbeltran:
Currently, the [dcat:Catalog](https://w3c.github.io/dxwg/dcat/#Class:Catalog) section includes descriptions of the properties inherited from ```dcat:Resource``` contrary to what we do on the other classes. So, this issue to address the removal of those.

agbeltran:
Better to address this after #717 has been merged.

agbeltran:
related to #704

PR #724, opened 2019-01-31
Consolidating property definitions for catalog

### LINKED ISSUE #704: Repetition in vocabulary section
In rewiring the internal document links, I noticed we've been a little inconsistent within Ch 6 with repetition of the inherited properties for `dcat:Catalog` from its super-classes `dcat:Resource` and `dcat:Dataset` which we don't repeat in the parallel sections for `dcat:Dataset` and `dcat:DataService`.  It would be easy to remove the duplicates most of the definitions/use e.g. `title` , `description` etc are identical though there are some usage notes that may require rewording to make sense such the various forms of the `language` property.

I think this is worth doing  but wanted to ask if it considered and ruled out?  (As a reader, duplicating in some places and not others without a visible rationale makes me worried I've missed something).

davebrowning:
will close #704 #718.

Would appreciate a careful review of definitions and usage notes for `dcat:Resource` carefully, since I've had to wordsmith the Catalogued Resource a little to retain clarity ( I hope). `language` (at Catalog level, at Dataset level, potentially recorded as distribution level), and well as update dates etc were the main cases.

Rendered at https://rawgit.com/w3c/dxwg/dcat-issues718-704-daveb/dcat/index.html, with htmldiff [here](https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fdxwg%2Fdcat-issues718-704-daveb%2Fdcat%2Findex.html)

I haven't found a good home for some of the statements such as  the very clear "A legal document under which the **catalog** is made available and **not** the **datasets** or **services**" from 2014 vocab  for `dcat:license` for `dcat:Catalog`.


dr-shorthair (approved):
I did a little more wordsmithing on resource_language
Otherwise OK

PR #729, opened 2019-02-04
Removed issue 371 as it is already closed.

aisaac:


nicholascar:
Actioning this as it's a simple removal of an Issue we've agreed is closed.

Issue #317, opened 2018-08-22
Change domain or create superclass of dcat:Distribution


rob-metalinkage:
dcat:distribution currently has domain dcat:Dataset

dcat:Resource needs the equivalent property, so either create a a superclass of Distribution and a corresponding superproperty of dcat:distribution or relax the domain of dcat:distribution to be dcat:Resource

(i favour the latter as ist not obvious whats different about the superclass...)

see #110

dr-shorthair:
I assume this is to assist in managing representations of standards/profiles? 

**Dataset --> Distribution** is a well-known terminology in the data space, but elsewhere **Resource --> Representation** is more typical. So I'd favour the superclass/super-property route. How about 
```
dcat:hasRepresentation a owl:ObjectProperty ;
    rdfs:domain dcat:Resource ; 
    rdfs:range dcat:Representation .

dcat:distribution rdfs:subPropertyOf dcat:hasRepresentation .

dcat:Representation a owl:Class . 

dcat:Distribution rdfs:subClassOf dcat:Representation .
```

jakubklimek:
> dcat:Resource needs the equivalent property

@rob-metalinkage Why is this? `dcat:Resource` is either `dcat:Dataset`, then it has `dcat:distribution`s, or `dcat:DataService`, then it does not need a `dcat:distribution`. I always pictured `dcat:Resource` as an abstract class. 

Are there any examples where you actually have an instance of `dcat:Resource` which is neither a dataset nor a data service and it needs to have a representation? And does this instance belong to DCAT catalogs?

dr-shorthair:
I understand that the concern is to be able to add Profiles (and other technical standards) to a Catalog. Is a Profile a subclass of Dataset, or is it another kind of Resource?

rob-metalinkage:
I dont think a profile is either a Dataset or a Service  - so its a direct subclass of Resource, hence the need to have the equivalent of a Distribution, and the suggestion to model this consistently with the dataset/Distribution relationship. I think Services and service descriptions will need to be treated the same too - e.g. a OAS API document for a service API.

rob-metalinkage:
@jakubklimek : In the Use Cases we have documented exactly this practice, where DCAT-AP profiles have been catalogues as pseudo-Datasets.

see #238

kcoyle:
[...]

Issue #127, opened 2018-02-15
dcat:mediaType - check constraints


dr-shorthair:
In DCAT v1 the property `dcat:mediaType` is axiomatized 
```
dcat:mediaType
  rdf:type owl:ObjectProperty ;
  rdfs:domain dcat:Distribution ;
  rdfs:range dcterms:MediaTypeOrExtent ;
  rdfs:subPropertyOf dcterms:format ;
.
```

1. Verify that the range is appropriate and necessary
2. Verify that the domain is appropriate and necessary (see #110)
3. Consider whether any guarded constraints (using owl:Restriction) should be introduced (see #105)

makxdekkers:
Dropping the domain axiom makes dcat:mediaType identical to dct:format of which it is a subproperty. However, the usage note for dcat:mediaType restricts its use to IANA media types, so maybe the range need to be tightened to a new class dcat:IANAMediaType?

akuckartz:
I always found the properties for media types and formats and especially the range "dcterms:MediaTypeOrExtent" confusing. I am in favor of improving that situation.

akuckartz:
Maybe this can be helpful:

Ontology for Media Resources 1.0
W3C Recommendation 09 February 2012
http://www.w3.org/TR/mediaont-10/

dr-shorthair:
The domain axiom appears relevant - unless it is required to sometimes say something about a dcat:CatalogRecord.

dr-shorthair:
Proposal:
1. leave domain alone
2. change range to dcat:IANAMediaType or dcat:MIMEtype

nicholascar:
+1 to changing to dcat:MIMEType (capital ‘T’?) as this will be better understood by web devs than dcat:IANAMediaType

makxdekkers:
@dr-shorthair Although I admit that I am the one who suggested the range "to be tightened to a new class `dcat:IANAMediaType`", on second thought, I think this would not be a good idea as it could break existing implementations. If someone had implemented DCAT using `dcat:mediaType` and used something that is not an IANA media type as its object but still a valid `dct:MediaTypeOrExtent`, that implementation would violate the spirit but not the letter of the specification. It could only be accused of not respecting the SHOULD in the [Usage Note](https://www.w3.org/TR/vocab-dcat/#Property:distribution_media_type), which is a very weak axiom, if you could call it an axiom at all. However, after the suggested change, that implementation would violate the letter of the rule. I don't think we want that.
[...]

Issue #126, opened 2018-02-15
dcat:downloadURL  - check constraints


dr-shorthair:
In DCAT v1 the property `dcat:downloadURL` is axiomatized 
```
dcat:downloadURL
  rdf:type owl:ObjectProperty ;
  rdfs:domain dcat:Distribution ;
  rdfs:range rdfs:Resource ;
.
```

1. Verify that the range is appropriate and necessary
2. Verify that the domain is appropriate and necessary (see #110)
3. Consider whether any guarded constraints (using owl:Restriction) should be introduced (see #105)

dr-shorthair:
Domain OK. 
Range axiom is redundant, given that it is `rdf:type owl:ObjectProperty`

dr-shorthair:
Resolved: no changes needed to axiomatization. But check the scopeNotes
See https://www.w3.org/2018/06/14-dxwgdcat-minutes.html#x09

PR #730, opened 2019-02-04
Richer typing of relations in bag-of-files case

dr-shorthair:
Minor changes + a new example, discussed in #317

### DIFF
--- a/dcat/examples/csiro-stratchart-schema.jsonld
+++ b/dcat/examples/csiro-stratchart-schema.jsonld
@@ -1,49 +1,49 @@
 {
   "@graph" : [ {
     "@id" : "_:b0",
-    "@type" : "s:EntryPoint",
+    "@type" : "sdo::EntryPoint",
     "actionApplication" : "_:b1",
     "comment" : "Service that supports queries to obtain RDF representations of subsets of the data",
     "name" : "International Chronostratigraphic Chart hosted at Research Vocabularies Australia",
     "url" : "https://vocabs.ands.org.au/viewById/196"
   }, {
     "@id" : "_:b1",
-    "@type" : "s:SoftwareApplication",
+    "@type" : "sdo::SoftwareApplication",
     "conformsTo" : "https://www.w3.org/TR/sparql11-query/",
     "url" : "http://vocabs.ands.org.au/repository/api/sparql/csiro_international-chronostratigraphic-chart_2017"
   }, {
     "@id" : "_:b10",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo::DataDownload",
     "comment" : "Webpage describing the ontology used for the data",
     "contentUrl" : "http://resource.geosciml.org/ontology/timescale/gts.html",
     "encodingFormat" : "mime:text/html"
   }, {
     "@id" : "_:b2",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo::DataDownload",
     "comment" : "TTL serialization of the RDF representation of the entire dataset",
     "encodingFormat" : "mime:text/turtle",
     "identifier" : "isc2017.ttl"
   }, {
     "@id" : "_:b3",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo::DataDownload",
     "comment" : "RDF/XML serialization of the RDF representation of the entire dataset",
     "encodingFormat" : "mime:application/rdf+xml",
     "identifier" : "isc2017.rdf"
   }, {
     "@id" : "_:b4",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo::DataDownload",
     "conformsTo" : "https://www.w3.org/TR/rdf-schema/",
     "comment" : "RDF representation of the data",
     "accessService" : "_:b0"
   }, {
     "@id" : "_:b5",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo::DataDownload",
     "comment" : "JSON-LD serialization of the RDF representation of the entire dataset",
     "encodingFormat" : "mime:application/ld+json",
     "identifier" : "isc2017.jsonld"
   }, {
     "@id" : "_:b6",
-    "@type" : "s:DataDownload",
+    "@type" : "sdo:DataDownload",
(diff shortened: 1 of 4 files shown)


agbeltran:
@makxdekkers - pinging you again about this, it would be great to have your review about this

makxdekkers:
Mostly OK, but there are a number of properties in the `dct:relation` with the `adms:Asset`, namely: `dcat:downloadURL`, `dcat:mediaType`, `dct:conformsTo` and `dct:description` that seem to be properties of the `adms:AssetDistribution` and not of the `adms:Asset`.

So I think there should be a `dcat:distribution` property under the `adms:Asset` with:
`dcat:distribution [ a adms:AssetDistribution ; dcat:downloadURL etc`

dr-shorthair:
OK - I've added two distributions of the Ontology - in Turtle and RDF/XML.

dr-shorthair:
See https://rawgit.com/w3c/dxwg/dcat-issue317-simon/dcat/index.html#bag-of-files 
and https://github.com/w3c/dxwg/blob/dcat-issue317-simon/dcat/examples/csiro-stratchart.ttl

makxdekkers:
Thanks @dr-shorthair. Looking more closely at the two sections with links to the images, I see that you describe them with `dcat:downloadURL` and `dcat:mediatype`, but both those properies have domain dcat:Distribution so it is being implied that the images are` dcat:Distribution`s. It might be better to describe them without DCAT-specific properties but use the equivalent DCMI terms instead. 
Something like:

`dct:relation <http://stratigraphy.org/ICSchart/ChronostratChart2017-02.jpg> ;`
...
```
<http://stratigraphy.org/ICSchart/ChronostratChart2017-02.jpg> [
      rdf:type foaf:Document ;
      dct:type <http://purl.org/dc/dcmitype/Image> ;
      dct:format <https://www.iana.org/assignments/media-types/img/jpeg> ;
      dct:description "Coloured image representation of the International Chronostratigraphic Chart" ;
      dct:issued "2017-02-01"^^xsd:date ;
      dct:title "International Chronostratigraphic Chart" ;
    ] ;
```

dr-shorthair:
Else relax the domain of `dcat:downloadURL` and `dcat:mediatype` so use of these properties does not necessarily entail that the resource is a `dcat:Distribution` ?

dr-shorthair:
I note that this was potentially up for consideration in #126 and #127 which were closed a long time ago. 
I suggest re-opening both in conjunction with this discussion.

makxdekkers:
@dr-shorthair Is this really necessary? You seem to be suggesting we change the domains of two DCAT properties in order to support an example.
Is there anything wrong with my counterproposal to describe the images with 'regular' DCMI terms?

dr-shorthair:
OK, I was just copying the pattern of `dcat:distribution` down to the `dct:relation` cases. 
But I guess it does not apply to resources that are not in a DCAT catalog (e.g. `foaf:Document`). 
So where would the descriptions of such related resources, which are neither `dcat:Resource` or `dcat:Distribution`, be available from? 

My goal here was to show how a simple pattern for how some additional information about the related resources could be included in the graph, but now I'm not sure it is possible.

makxdekkers:
You ask "_So where would the descriptions of the related resources that are neither `dcat:Dataset` or `dcat:Distribution` be available from?_"
This is always an issue: we would like to rely on the owner of 'external' resources to describe their resources with LOD statements. In the case of http://stratigraphy.org/ICSchart/ChronostratChart2017-02.jpg, you'd hope that stratigraphy.org provide the set of statements that describe the image. Unfortunately, many resources on the Web do not have that kind of information available, but that is how LOD works (or doesn't work in some cases).

smrgeoinfo:
What about using schema.org DigitalDocument type, which allows a URL to explicitly state where to get the resource representation; otherwise it appears one must assume that the URI for dct:relation object is resolvable (which ideally would be true, but not necessarily). 

```
<http://stratigraphy.org/ICSchart/ChronostratChart2017-02> [
      rdf:type sdo:DigitalDocument;
      sdo:url  "http://stratigraphy.org/ICSchart/ChronostratChart2017-02.jpg";
      dct:type <http://purl.org/dc/dcmitype/Image> ;
      dct:format <https://www.iana.org/assignments/media-types/img/jpeg> ;
      dct:description "Coloured image representation of the International Chronostratigraphic Chart" ;
      dct:issued "2017-02-01"^^xsd:date ;
      dct:title "International Chronostratigraphic Chart" ;
    ] ;
```

makxdekkers:
@smrgeoinfo That is another way of doing it. However, you're still creating a description for an 'external' resource that you don't own. Doing so creates a risk as you cannot control what the owner does with its resource. For example, stratigraphy.org could tomorrow reissue the image and from that time the  `dct:issued` in the description you have provided will be incorrect.

dr-shorthair:
@makxdekkers I have revised the PR following your suggestion. 

See https://rawgit.com/w3c/dxwg/dcat-issue317-simon/dcat/index.html#bag-of-files
and https://github.com/w3c/dxwg/blob/dcat-issue317-simon/dcat/examples/csiro-stratchart.ttl

makxdekkers (approved):
OK for me. Thanks!

dr-shorthair:
FWIW I also made a schema.org variant of the same example 

https://github.com/w3c/dxwg/blob/dcat-issue317-simon/dcat/examples/csiro-stratchart.schema.ttl

It is almost complete - the only props I couldn't match are `dct:conformsTo` and `dcat:accessService`.

agbeltran:
There is some discussion about including something like ```dct:conformsTo``` in schema.org (see https://github.com/schemaorg/schemaorg/issues/1516) but not yet available AFAIK.

agbeltran:
We do have the prefixes ```dctype``` and ```adms``` defined in the document. Should we use them too in this example?

Also, section D.1 (https://rawgit.com/w3c/dxwg/dcat-issue317-simon/dcat/index.html#bag-of-files for this PR), presents several representations by adding more details to the same example, which is nice. While this is clear by looking at the IDs, etc, it might be useful to spell it out in the presentation.

dr-shorthair:
I've inserted some `rdfs:comment` to explain the role and type of the different related resources

Issue #644, opened 2019-01-09
Specify the semantics of prof:isTransitiveProfileOf wrt prof:isProfileOf


aisaac:
(This came to mind while writing the ESWC paper)
One note at https://w3c.github.io/dxwg/profilesont/#Property:isTransitiveProfileOf says

> While this ontology provides this `prof:isProfileOf` & `prof:isTransitiveProfileOf` pair of properties, it does not specify how a particular implementation of a Profile that is related to another Profile or Standard by `prof:isTransitiveProfileOf` should implement specific inferences. 

That talk about lack of inference could be confusing. As a matter of fact if PROF is to follow the SKOS pattern for `skos:broader` and `skos:broaderTransitive`, the idea is that `prof:isTransityProfileOf` should be a transitive property and it should be a super-property of `prof:isProfile`. It isn't either, at the moment.

andrea-perego:
+1 to @aisaac . See also my more general concerns on this property in https://github.com/w3c/dxwg/issues/643#issuecomment-452643079

rob-metalinkage:
proposed action: declare prof:isProfile as a subproperty of prof:isTransitiveProfileOf

aisaac:
+1 to @rob-metalinkage 's proposed action. I would still push for trying to remove from the notes any occurrence of the word 'inference' that could (after a first, shallow look) contradict with the goal of applying the formal semantics given to the property (ie. applying an inference engine to infer the transitive closure of the property).

nicholascar:
Addressed by PR https://github.com/w3c/dxwg/pull/734

kcoyle:
@nicholascar I'm not convinced that this PR is a response to the question posed here. As I read it, the PR says: 

```
# A profile that is within a hierarchy of profiles may wish to indicate it profiles
# things "further up the chain". To do this, prof:isTransitiveProfileOf can be used
# to indicate anything the profile is related to by a series of one or more
# prof:isProfileOf properties.
```

but that's pretty vague: "things further up the chain". What does "profile" mean as a verb? Is anything a "thing" to be profiled?

Perhaps the diff doesn't show enough context and this snippet out of context isn't sufficient. 

Also, there is still at least one reference to "inference" that should be reviewed by @aisaac .

```
# Now, according to the semantics of prof:isTransitiveProfileOf, using the
# prof:isProfileOf statements above, one can infer the following additional
# statements:
```

[...]

Issue #643, opened 2019-01-09
prof:isTransitiveProfileOf needs more convincing case and/or example 


aisaac:
As in the title...
This came to mind while writing the ESWC paper.
The property is not super-clear in https://w3c.github.io/dxwg/profilesont/ and there's no example for it.

andrea-perego:
+1 to @aisaac . 

I am actually unsure how `prof:isTransitiveProfileOf` can be safely implemented. E.g., this would require at least all profiles in the "chain" to be under the same governance and design rules. And also in this case profile transitiveness may not be granted.

Do we have a use case for this property?

kcoyle:
w3c/dx-prof#2 goes into this as well, even though the immediate thrust was its name. And this links to many of the comments by @agreiner about the dangers of inheritance.

rob-metalinkage:
This predicate was included because @agreiner argued we needed a mechanism to have a "flat representation" - not that inheritance wasnt a fundamental requirement or had any specific "dangers" - its a convenience we dont specify how should be used.

proposed action: include in examples.

aisaac:
The proposal to add an example sounds of course good to me :-) 
The beginning of explanation for the motivation sounds quite appropriate too.

aisaac:
@nicholascar it conveys useful information, thanks!

But it may have one flaw:
> 
> The ISO thing that the NZ Profile profiles is actually a Web Ontology Language
> OWL) version of the original ISO addressing standard 
sounds really loose. Can "the ISO thing" be something like "ISO19160-1"? 
And here's where the flaw could be: in this case, is `<http://linked.data.gov.au/def/iso19160-1-address>` really a profile of ISO19160-1:2015? I think that one could see in it a resource descriptor (in OWL) for that base ISO standard, not a fully-fledged profile.

Also, in terms of formulation, my expectation would be that an example for a transitive property should not hint that data publishers have to materialize the statements of that property. In SKOS we were rather clear that the general expectation is that data publishers assert `skos:broader` statements and that `skos:broaderTransitive` are then infered, possibly by others, mainly for data consumption scenarios (i.e., querying). 
Of course this is not a general rule. Still, the intro and formulation for your example could be changed.

Here's the order of things that I would suggest:

1. The example without the transitive statements.

2. A sentence that introduces the inference/semantics. Like "From the `prof:isProfileOf` statements in this example, and according to the semantics of `prof:isTransitiveProfileOf`, one can infer the following statements:"

3. The transitive statements infered.

[...]

PR #734, opened 2019-02-06
Issue 644

nicholascar:
Addresses Issues https://github.com/w3c/dxwg/issues/643 & https://github.com/w3c/dxwg/issues/644: "prof:isTransitiveProfileOf needs more convincing case and/or example" & "Specify the semantics of prof:isTransitiveProfileOf wrt prof:isProfileOf" in separate commits.


aisaac:
@nicholascar would it be possible to split this PR? I believe that the commit for isTransitiveProfileOf semantics is fine, and we could close issue #644 , but in the current state I think the example is not right enough for closing #643 . Separating them would allow the latter to not block the former!

aisaac:
@nicholascar @rob-metalinkage @kcoyle the discussion on PROF examples yesterday and on #745 makes me realize that we may block something useful if this PR is not merged. In that case it can be merged, but by all means it shouldn't mean that the corresponding issues can be closed now!
(even though I still think that #644 could be closed, and that splitting this PR would have been ideal...)

nicholascar:
Thanks @aisaac, we’ll get this merged, leave the issues open and in the future I’ll try and make more granular PRs that try and tackle single smaller things each. That should make approving bits easier and reduce the chance of things being held up!

PR #735, opened 2019-02-06
Change to dcat:Distribution definition and note

agbeltran:
As per discussions in DCAT sprint: https://www.w3.org/2019/02/05-dxwgdcat-minutes

Preview: https://rawgit.com/w3c/dxwg/agb-distribution-definition/dcat/index.html#Class:Distribution

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1627,6 +1627,8 @@ $(document).ready( function() {
 
     </section> <!-- end class Dataset -->
 
+
+    <!-- class Distribution -->
     <section id="Class:Distribution">
         <h3>Class: Distribution</h3>
         <p>The following properties are recommended for use on this class:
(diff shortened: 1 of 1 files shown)


dr-shorthair (commented):
Should we also mention the idea of "informationally equivalent". I suggest adding this to the `Note:` 

"In some cases all distributions of a dataset will be fully informationally equivalent, in the sense that lossless transformations between the representations are possible. An example would be different serializations of an RDF graph using RDF/XML, Turtle, N3, JSON-LD. However, in other cases the distributions might have different levels of fidelity to the underlying data. For example, a graphical representation alongside a CSV file. The question of whether different representations can be understood to be distributions of the same dataset is use-case specific, so the judgement is the responsibility of the provider."

agbeltran:
thanks @dr-shorthair - it's been added

agbeltran:
I'll also remove the previous note that read:

"The definition text of dcat:Distribution has been revised to clarify that distributions are primarily representations of datasets. As such, all distributions of a given dataset should be informationally equivalent. "

dr-shorthair (approved):

PR #745, opened 2019-02-08
Prof examples

nicholascar:
All examples separated into individual files,  all RDF properties updated to match ontology, diagrams delivered as SVG, colour coding used instead of rdf:type relations.


andrea-perego (commented):
@nicholascar , it is not clear to me why the examples have been moved to separate files. I thought the intention was to maintain them separately from the spec, and loading them dynamically via `@data-include` (see https://github.com/w3c/respec/wiki/data--include), but I saw that the examples are already included in the spec.

kcoyle:
"... Profiles Ontology which is, itself, a profile of the Dublin Core Terms vocabulary."

This use of "profile of" has been questioned by Tom Baker, who is the main expert on the Dublin Core Terms, and should not be included here until it has been resolved.

nicholascar:
@andrea-perego I think this is just a not-sorted-out-yet problem. We've always had both example files in example/ and also some examples in the spec itself. Simon and others have recently hassled me about putting examples right up front in the spec.

Since I have to keep updating examples when the spec changes, I like the idea of only one place to go to update them all so perhaps we can aim to put all examples in examples/ and then do as you suggest to load them dynamically. 

I guess I've begun a process of placing all examples in examples/ already but I'll make it a goal to have them all in there and included across to the spec for the 3PWD.

nicholascar:
@kcoyle the use of isProfileOf of Dublin Core is correct *as PROF is constructed now*. I agree that there’s a need, highlighted by Tom Baker, to define Standards/Profiles v. Namespaces and examples like these help with that as they put the vocab as it currently is to the test.

Some of the discussion at the plenary meeting I missed this morning would have benefitted from up-to-date PROF examples, even if that was to indicate where everyone thinks they are wrong!

So please allow this PR through so we can iterate on top of it. If it is held back - it updates all examples so much more than just PROF’s profile - we stall progress.

For me, the process of improving models is theory followed by test implementations (examples) and then iterations on top of those.

nicholascar:
The conflicts this PR now has are due to my other PR to revamp PROF from Ontology to Vocab having been metered.

Should be easy to sort out this merge preserving the Vocabulary name.

kcoyle:
Go ahead and merge and we'll sort out the "profile of" question later.

Issue #482, opened 2018-10-22
Distribution composed of more than one file, but not packaged


dr-shorthair:
A Distribution may be composed of multiple files which cannot be used independently, such as a shapefile and its attendant sidecars (index and database files). These might not be packaged into a single distributable artefact, such as a tar or zip archive (see #54 and #259) . So a dataset's distribution, while a single entity, is composed of multiple artefacts. We need to show patterns about how these will appear in a catalog.

agbeltran:
One particular case for the multiple files of a distribution is that of the checksum files. This is something we want to include in DATS (see https://github.com/datatagsuite/schema/issues/11) and it would be good to have specific vocabulary to refer to them.

agreiner:
I have a use case where we have log files that we release as datasets. One "log" could contain hundreds of raw files, making it impractical to treat each as a separate dataset. They are often grouped in multiple layers of directories, such as a directory per cabinet, and within that a directory per chassis, and within that a directory per slot, etc. We've been releasing them as a single gzip, but they are getting so big that it's impractical to use a simple http download. (We offer Globus instead.)

makxdekkers:
This seem to be a completely new requirement. And, given the resource and time constraints, I think it will be hard to incorporate a solution in the current revision of DCAT. I also think that it might be difficult to define a single approach/vocabulary that covers all possible relationships -- even the three examples from @dr-shorthair, @agbeltran and @agreiner seem to require different solutions.
In the current revision, you could do something with `dcat:accessURL` which you could use to point to a landing page where you link to various files and explain their relationships. The use of `dcat:downloadURL` is only possible for a single file according to the definition: "_The URL of **the downloadable file** in a given format_". 
Should we leave this for future work?

dr-shorthair:
@agreiner I wonder if this is just another case of part-whole, as discussed in #411 (proposed new property `dcat:componentDistribution`) ? 

@makxdekkers Yes - I tend to agree - our docket is quite full. However, perhaps we can recommend the solution agreed for #256 (i.e. use `dct:relation` ) and leave stronger semantics to another day.

agreiner:
[...]

Issue #259, opened 2018-06-22
How to express distributions provided as compressed files


nichtich:
The `dcat:downloadURL` of a `dcat:Distribution` can point to compressed files (`.zip`, `.gz`...). What is the data format in this case? Stating that it is a ZIP-File will not help much: the more interesting information is what's in the archive. For `dcat:mediaType` the `+gzip` [MIME TYPE suffix](https://www.iana.org/assignments/media-type-structured-suffix/media-type-structured-suffix.xhtml) can be added but what to put into `dct:format` - an identifier of the archive format or of its content (I'd prefer the latter)?

nicholascar:
I think you're asking a bit much here! If there are a bunch of different things all zipped together (say PDF files, CSV and images), then what are you going to do if you want to communicate the content, provide a list of media types like this: `application/pdf,text/csv,image/png`? How is this helpful for machine purposes? I written inventory of dataset content would do for a person (in the `Distribution`'s`dct:description`).

`dcat:mediaType` should be used with a Media Type to indicate the compression types unless the compression type itself doens't have an IANA Media Type (a new BZip format or something) in which case `dct:format` should be used.

akuckartz:
See also: #54

nicholascar:
@akuckartz thanks for the reminder about [#54](https://github.com/w3c/dxwg/issues/54) which I recall now that you mention it!

Is this issue a duplicate? I think it is. Or, at least, the scope of [#54](https://github.com/w3c/dxwg/issues/54) covers this issue. I move to close this issue in favour of dealign with compression issues in [#54](https://github.com/w3c/dxwg/issues/54).

jakubklimek:
@nicholascar I actually these there are 2 separate issues and their combination, but even in #54 they are a bit mixed up. 
[...]

Issue #54, opened 2018-01-18
Distribution package [RDIP]


jpullmann:
<section class='requirement' id='RDIP'><h3>Distribution package [<a href='https://www.w3.org/TR/dcat-ucr#RDIP' target='_blank'>RDIP</a>]</h3><p>Define way to specify content of packaged files in a Distribution. For example, a set of files may be organised in an archive format and then compressed, but dct:hasFormat property only indicates the encoding type of the outer layer of packaging.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID1' target='_blank'>DCAT packaged distributions [ID1]</a> </section>

makxdekkers:
There is an RFC (6839, Additional Media Type Structured Syntax Suffixes, https://www.rfc-editor.org/rfc/rfc6839.txt) that defines the '+zip' suffix that can be appended to any IANA defined media type. This takes care of a common way of packaging distribution files, although it is not a full solution to the issue raised.
A more general solution (but still only covering one level of embedding) is the one adopted by ADMS (https://www.w3.org/TR/vocab-adms/). ADMS contains a property representationTechnique (https://www.w3.org/TR/vocab-adms/#adms-representationtechnique) that points to "_More information about the format in which an Asset Distribution is released. This is different from the file format as, for example, a ZIP file (file format) could contain an XML schema (representation technique)_."

kcoyle:
Incorporated into DCAT WD.

PR #746, opened 2019-02-09
Packaging and compression of files in a distribution

makxdekkers:
I added text for the packaging/compression of files in the distribution. I simplified the proposal by only adding one property: `dcat:wrapFormat`.

I could have created four properties `dcat:compressionFormat`, `dcat:compressionMediaType`, `dcat:packagingFormat` and `dcat:packagingMediaType`, but I thought that would make the solution much too complex. I would think that it would be fairly trivial for the system to decompress or unpack using a single property, and the duplication between mediaType and format seemed unnecessary.


agbeltran:
Preview here: https://rawgit.com/w3c/dxwg/makxdekkers-patch-1/dcat/index.html

smrgeoinfo:
Why is the range dct:MediaTypeOrExtent instead of dct:MediaType?

jakubklimek:
I still think that 2 or 4 properties would be cleaner than just the one.

1. The separation between mediaType and format is due to the incompleteness of IANA Media Types list (e.g. does not contain Media Type for `.tar` files) and, therefore, usage of, e.g. File Type EU MDR NAL in DCAT-AP in addition to it. Since we have both properties describing the main content of the distribution, why should the packaging and compression be described by only one of those?
2. Could we go back to #259 and try to represent the various use cases (compressed file, packaged files, compressed and packaged files) using just the wrapFormat property and see how the files can be automatically processed based on that description? I think this will only work for simply compressed file, e.g. `.csv.gz`. We will be unable to determine that the file is actually an archive of multiple files, e.g. `.zip` containing multiple `.csv` files, and how to unpack it (e.g. compared to `.tar.gz` containing multiple `.csv` files). That is why I would rather have a robust solution than a less complex one. I would say that the three use cases are pretty common and it would be nice to be able to properly describe them.

makxdekkers:
I am happy to consider the more complex solution with multiple properties, if that is what the group wants. In response to @jakubklimek I would say that 'cleaner' is not the main objective here -- I would say that a more complex solution should be chosen if it is essential and a simpler solution does not work.
Furthermore, I think that even the approach with four properties can only distinguish simple cases; you can only indicate the outermost and the innermost formats. For example, there is no way to know whether a RAR archive contains multiple ZIP files that contain CSV and XLS files.

makxdekkers:
@smrgeoinfo I made the range `dct:MediaTypeOrExtent` instead of `dct:MediaType` to align with the range of `dcat:mediaType` and `dct:format`, but you are right, it could easily be `dct:MediaType`.

smrgeoinfo:
I was assuming that for a packaged distribution, one could use dct:conformsTo to specify the packaging convention (e.g. bagit, OASIS OpenDocument Package Format...). In those schemes there is a manifest and metadata document in the package that details the internal structure and file formats.

jakubklimek:
@smrgeoinfo That may apply to distributions complying with the package formats you mention. However, we could have a more loose package. If we had for instance a tar or zip file with 100 CSV files, one per day, all with the same schema, this schema (CSV on the Web JSON-LD descriptor) would go to `dct:conformsTo` and we need a separate property (or properties) to describe the package/compression.

jakubklimek:
@makxdekkers You are right about the simple cases. However, I would say that double compression and similar techniques are often enough just bad practice. I often see compressed single files and packaged (and compressed) sets of files with the same schema where it makes sense (splitting larger files into smaller ones or just reducing size of file on disk). However, when I come across a zip file containing many other zip files, it is often because the publisher does not know what they are doing. The question seems to be where do we draw the line.

makxdekkers:
Can I ask the members of the WG to vote for one of the two alternatives:

1. use a single property, `dcat:wrapFormat`, for both compression and packaging, as in the pull request
2. use two properties, e.g. `dcat:compressFormat` and `dcat:packFormat`, to distinguish between the two

Thanks.

larsgsvensson:
@makxdekkers How does the vote take place? My suggestion would be one comment for each of the proposals (single property vs two properties) and then use thumbs-up for the vote.

makxdekkers:
Proposal 1:  use a single property, `dcat:wrapFormat`, for both compression and packaging, as in the pull request 
Please give thumbs-up if you agree.

makxdekkers:
Proposal 2: use two properties, e.g. `dcat:compressFormat` and `dcat:packFormat`, to distinguish between the two
Please give thumbs-up if you agree.

dr-shorthair (changes requested):
1. These new properties also need to be expressed in the RDF file (i.e. dcat.ttl) 
2. Also add an entry in the change-log (Annex E)

agbeltran:
Linking here the related issues: #54, #259, #482

We should also add examples to cover the different use cases.

dcat/index.html
```diff
@@ -62,12 +62,12 @@
...
-        and the list of all the changes since the previous version of DCAT in the <a href="#changes">Change History</a> section.
+        and the list of all the changes since the previous version of DCAT in the <a href="#changes">compression</a> section.
     </p>
 
     <h3 id="dcat_history">DCAT history</h3>
     <p>The original DCAT vocabulary was developed and <a href="http://vocab.deri.ie/dcat">hosted</a> at the Digital Enterprise Research Institute (DERI), then refined by the <a href="http://www.w3.org/egov/">eGov Interest Group</a>, and finally standardized in 2014 [[?VOCAB-DCAT-20140116]] by the <a href="http://www.w3.org/2011/gld/">Government Linked Data (GLD)</a> Working Group.</p>
-    <p>This revised version of DCAT was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[?DCAT-UCR]] submitted on the basis of experience with the DCAT vocabulary from the time of the original version, and new applications not originally considered. A summary of the changes from  [[?VOCAB-DCAT-20140116]] can be found at <a href="#changes">Change History</a></p>
```
agbeltran: This needs to be rolled back

dcat/index.html
```diff
@@ -4073,6 +4107,9 @@ <h3>Changes since the W3C Recommendation of 16 January 2014</h3>
                 Each relevant class section was extended with properties to deal with licensing, acccess rights, other rights and policies associated with the classes, as
                 recommended in the  <a href="#license-rights">License and rights statements</a> section.
             </li>
+            <li>
```
agbeltran: This needs to be removed: it is repeated because I merged the two branches that had similar changes

agbeltran: done in 5b88e8a

dcat/index.html
agbeltran: done in 67d9e3b

agbeltran:
@makxdekkers I merged [the other branch](https://github.com/w3c/dxwg/tree/dcat-makx-packaging) into this PR

@dr-shorthair if you could have a final check and merge, we resolved this on the call today: https://www.w3.org/2019/02/20-dxwgdcat-minutes.html#x13

agbeltran:
the remaining work on this is to add the examples

dr-shorthair (approved):

dr-shorthair:
Merging for now, noting that instance examples also needed. 
Working on the assumption that the discussion thread above is sufficient endorsement by the team.

dr-shorthair:
Hmmm. This merge was probably premature. I don't think we have a decision recorded to support it. My bad. And not competent to unwind ...

agbeltran:
We did decide about this on the call yesterday: https://www.w3.org/2019/02/20-dxwgdcat-minutes.html#x13

dr-shorthair:
Phew. It was resolved after I left the meeting and before I read the minutes :-) No reverts needed.

Issue #751, opened 2019-02-12
Inconsistent definition for dcat:accessService


dr-shorthair:
There is an inconsistency in the definition of `dcat:accessService`. 

The [definition of `accessURL`](https://w3c.github.io/dxwg/dcat/#Property:distribution_accessurl) notes that 
>` dcat:accessURL` generally matches the property-chain `dcat:accessService/dcat:endpointURL`. 

However, the [definition of `accessService`](https://w3c.github.io/dxwg/dcat/#Property:distribution_accessservice) immediately following says 
> Sub-property of: `dcat:accessURL`

These can't be simultaneously true. The sub-property axiom is wrong. It is probably cruft from earlier drafts.

PR #752, opened 2019-02-12
DCAT issue 751 - accessService is not sub-property of accessURL

dr-shorthair:
Addressing #751

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1827,7 +1827,7 @@ $(document).ready( function() {
             </table>
 
             <p>
-                <code>dcat:accessURL</code> generally matches the property-chain <code>dcat:accessService</code>/<code>dcat:endpointURL</code>. In the RDF representation of DCAT this is axiomatized as an OWL property-chain axiom.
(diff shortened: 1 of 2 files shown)


agbeltran (changes requested):
I suggest we change

"dcat:accessURL generally matches the property-chain dcat:accessService/dcat:endpointURL. In the RDF representation of DCAT this is axiomatized as an OWL property-chain axiom."

for 

"dcat:accessURL matches the property-chain dcat:accessService/dcat:endpointURL. In the RDF representation of DCAT this is axiomatized as an OWL property-chain axiom."

as the axiom is included, we can safely remove 'generally'?

I'd also use the Manchester OWL Syntax for the property chain 'dcat:accessService o dcat:endpointURL'

This property chain implies some domains/ranges that are not explicitly stated - do we want to do include them?

agbeltran:
@dr-shorthair let me know what you think about the suggestions in the review - I could implement the changes

dr-shorthair:
Thanks @agbeltran. Yes I agree with your suggestions. Should indicate what the notation used is. 

(Would be nice to have all the example available in alternative syntaxes. We all like Turtle, but the market is looking for JSON. Like here: https://www.w3.org/TR/owl2-primer/#OWL_Syntaxes )

agbeltran (approved):

agbeltran:
Addressed the very minor text change - we can address the different syntaxes in a separate issue.

PR #768, opened 2019-02-19
Profiles Ontology -> Profiles Vocabulary

nicholascar:
As per vote via email https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Feb/0259.html


rob-metalinkage (approved):
looks ok -

Issue #84, opened 2018-01-18
Summary statistics [RSS]


jpullmann:
<section class='requirement' id='RSS'><h3>Summary statistics [<a href='https://www.w3.org/TR/dcat-ucr#RSS' target='_blank'>RSS</a>]</h3><p>Express summary statistics and descriptive metrics to characterize a Dataset.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID33' target='_blank'>Summarization/Characterization of datasets [ID33]</a> </section>

makxdekkers:
[DQV](https://www.w3.org/TR/vocab-dqv/) should be able to meet these requirements.

dr-shorthair:
Do we also need statistics on distributions? This requirement is suggested by the comment submitted by Daniel Pop [1]. Of course it also depends on how we resolve the matter of 'information equivalence' of different distributions. 

[1] https://lists.w3.org/Archives/Public/public-dxwg-comments/2019Jan/0013.html

andrea-perego:
I think we shouldn't prevent this - as done for other information. The question is how to put this option in the spec.

I'm a bit reluctant to explicitly add properties in class definitions where we don't have real-world use cases and/or implementation evidence. So, this could be included in the "guidance" part of the spec.

About "information equivalence", (again) -1 to it. This ends up to be a matter of the "granularity" of the notion of dataset, which is mainly a data provider choice (possibly also based on the requirements of the intended users).

makxdekkers:
@dr-shorthair I am not quite sure how you derive a requirement for statistics on datasets? If there is a need for it, maybe we could refer to DQV or Data Cube?
In my mind, Daniel's point (1) could be resolved by modelling the real-time data stream as a `dcat:DataService` and modelling the CSVs as separate datasets.

dr-shorthair:
The statistic that Daniel mentioned is the frequency or spacing of members in a time series, where various distributions might have fixed spacing that is different (usually coarser) than what is available from the underlying dataset. I was on the point of creating an explicit issue for this aspect alone, but since this is an aspect of dataset statistics I thought it would be best to open the discussion here first.

dr-shorthair:
@makxdekkers I did not derive a new requirement for dataset statistics - this was one of the original requirements taken from UCR. 

[...]

PR #776, opened 2019-02-27
DCAT - temporal resolution

dr-shorthair:
Add `dcat:temporalResolution` as discussed in #84


agbeltran:
Pre-view: https://rawgit.com/w3c/dxwg/dcat-issue84-tres-simon/dcat/index.html#Property:dataset_temporalresolution

agbeltran:
I think it would be helpful to expand the description below [the example](https://rawgit.com/w3c/dxwg/dcat-issue84-tres-simon/dcat/index.html#ex-dataset) to explain what we are representing with ```dct:temporal``` and ```dcat:temporalResolution```.

In [the section describing the property](https://rawgit.com/w3c/dxwg/dcat-issue84-tres-simon/dcat/index.html#Property:dataset_temporalresolution), there is a type ```dct:temporalResolution``` instead of ```dcat:temporalResolution```.

agbeltran:
Following the discussion in #84, should we also add a comment referring to DQV for more complex cases?

davebrowning (approved):

davebrowning:
As per meeting minutes https://www.w3.org/2019/02/27-dxwgdcat-minutes#x12

init-dcat-ap-de:
The definition is 
"Minimum time period resolvable in the dataset."
The usage note is
"If the dataset is a time-series this should correspond to the spacing of items in the series. For other kinds of dataset, this property will usually indicate the smallest time difference between items in the dataset."

So the definition changes according to the use case. How do I describe the frequency of the time-series and the resolution of the dataset?
How do I know whether "temporal resolution" is the frequency of a time-series or the resolution?

makxdekkers:
Good point. The mention of a time-series is a bit iffy, as we have already said we need to treat that issue more fundamentally in future work.
Could we just simplify the usage note to:
"_This property will usually indicate the smallest time difference between items in the dataset_."
I think this simplified statement could also be seen as including 'spacing of items'.

PR #777, opened 2019-02-27
DCAT spatial resolution

dr-shorthair:
Add `dcat:spatialResolutionM` as discussed in #84 

1. Fixed UoM makes it a `owl:DatatypeProperty` (i.e. value is a literal, xsd:decimal)
2. is `dcat:spatialResolutionM` too obscure? Would `dcat:spatialResolutionInMetres` be better?


riccardoAlbertoni:
+1 to  renaming  dcat:spatialResolutionM to dcat:spatialResolutionInMetres!

nicholascar:
+1 to Metres. I had no ideas what M was

agbeltran:
I agree on spelling out the units rather than the cryptic M

agreiner:
I like this solution, but I think it should be "spatialResolutionInMeters" to be consistent with American English.

dr-shorthair:
> it should be "spatialResolutionInMeters" to be consistent with American English.

... would be more convincing if Americans actually used the metric system. 
W3C editorial standard is to use US-english spellings within text, but I don't believe that extends to tokens. 

'metre' is 
- the name of the SI unit, 
- the spelling in all other English-speaking jurisdictions, 
- (most importantly) the spelling in all non-English languages that use the latin alphabet.

agreiner:
Scientists in the U.S. do use the metric system. I only point this out because W3C style does call for American English in its docs. I would assume that includes tokens.

dr-shorthair:
@draggett perhaps you can adjudicate here. Does W3C style require that 'metre' use the US spelling ('meter') in either or both text and tokens? Or does the US-english style only apply to general language and syntax constructions in the narrative, but not to technical terms?

makxdekkers:
@dr-shorthair Well, not all languages that use the Latin script spell 'metre'; the world's [second-most spoken language](https://en.wikipedia.org/wiki/List_of_languages_by_number_of_native_speakers), Spanish, being one such example ;)

dr-shorthair:
@makxdekkers whoops, I really should have checked :-) 

- SI, en-GB, en-CA, en-AU, en-NZ, french, catalan ... - metre
- en-US, german, dutch, danish, swedish, norwegian ... - meter
- italian, spanish, portuguese, galician - metro

However, even given this diversity, I suggest that the SI or ISO designation "metre" is best in a technical context.

agreiner:
I think that what you see for the standards will depend on who presents it. See, for example, https://www.nist.gov/pml/weights-and-measures/si-units-length and http://www.npl.co.uk/si-units/metre/. The principle of least surprise seems applicable here, though. A mismatch in the locale for the document and that for the tokens seems like it will lead to confusion. I think we would agree that minting tokens in some other language would be confusing. This seems to me the same issue.

dr-shorthair:
Thanks @agreiner . 

I've looked for W3C precedents. The most instances are found in the [Spatial Data on the Web best practice](https://www.w3.org/TR/sdw-bp/), and indeed the spelling 'meter' is used consistently within the text. There are many fewer examples in tokens, and all from external vocabularies, in which the spelling is mixed depending on whether the origin in European or American. 

Ultimately the tokens are secondary to the semantics, so as long as those are clear I accept your consistency argument. I'll update the PR to match. 

Nevertheless ... [Map of countries officially not using the metric system](https://www.zmescience.com/other/map-of-countries-officially-not-using-the-metric-system/)

agbeltran (approved):

agbeltran:
We decided on accepting this PR in the last meeting (https://www.w3.org/2019/02/27-dxwgdcat-minutes#x12), and after the changes to 'meters' and a revision of both the editor's draft (https://rawgit.com/w3c/dxwg/dcat-issue84-sres-simon/dcat/index.html#Property:dataset_spatialresolution) and the RDF file, I am going to merge it now.

Issue #728, opened 2019-02-03
Clarify definition of dct:accrualPeriodicity in the context of DCAT


dr-shorthair:
The DCMI definition of `dct:accrualPeriodicity` is [1] 

> "The frequency with which items are added to a collection"

However, the DCAT definition is [2] 

> "The frequency at which dataset is published"

which has a rather different meaning. In particular, the rate of addition of data to a time series might be quite different to the frequency of its release or publication. 

This needs to be clarified. 

[1] http://dublincore.org/documents/dcmi-terms/#terms-accrualPeriodicity 
[2] https://w3c.github.io/dxwg/dcat/#Property:dataset_frequency

dr-shorthair:
This issue was triggered by a comment submitted by Daniel Pop [3]

[3] https://lists.w3.org/Archives/Public/public-dxwg-comments/2019Jan/0013.html

andrea-perego:
Concerning specifically [the relevant Daniel Pop's comment](https://lists.w3.org/Archives/Public/public-dxwg-comments/2019Jan/0013.html):

- Whether update frequency can be applied as well at the distribution level: this is very much up to how a dataset is updated and this information is documented. In my experience, I saw that this is typically done at the dataset level, but I don't see any issue why this shouldn't be done also for distributions - especially if they have different update frequencies

- Daniel raises indirectly the issue of whether distributions with different update frequency should lead to the creation of distinct datasets. This is one of the cases where the approach to be chosen depends totally on the data provider and community practices. Nonetheless, the fact that in DCAT update frequency is at the dataset level must not by itself mean that we need to have different datasets to address this use case

dr-shorthair:
The concern applies primarily to datasets that are time-series, which might be issued on a different schedule to the rate of addition of data to the dataset. For example, a dataset showing rainfall at 5-minute intervals might be formally published and registered just once a day. 

The DCAT definition appears to specify the latter, while the DCMI definition encompasses both. I think we need to either

- make this explicit - add a usage note "In this context the use of `dct:accrualPeriodicity` concerns the rate at which the dataset as-a-whole is published, which is a narrower usage than permitted by the original DCMI definition. To describe the rate of addition of items within a `dcat:Dataset`. use XXX instead. "  
- provide XXX to describe the internal spacing of items within a dataset. (also see #84 )

Else 

- revert to the DCMI definition, and clarify that this refers to the periodicity of accrual of items within a dataset, and recommend a different way to describe the publication schedule. 

[...]

Issue #62, opened 2018-01-18
Datasets vs. Catalog relation [RDSCR]


jpullmann:
<section class='requirement' id='RDSCR'><h3>Datasets vs. Catalog relation [<a href='https://www.w3.org/TR/dcat-ucr#RDSCR' target='_blank'>RDSCR</a>]</h3><p>Clarify the relationships between Datasets and zero, one or multiple Catalogs, e.g. in scenarios of copying, harvesting and aggregation of Dataset descriptions among Catalogs.</p><br/><em>Related use cases</em>: <a href='https://www.w3.org/TR/dcat-ucr#ID35' target='_blank'>Datasets and catalogues [ID35]</a> </section>

chris-sepa:
Surely this issue only relates to Datasets with a relationship to multiple Catalogs.
Zero relationships - not in a Catalog so not a DCAT Dataset and out of scope.
One relationship - this is what the current DCAT vocabulary describes.

makxdekkers:
@chris-sepa In my opinion, it should be perfectly legal to declare something the be a `dcat:Dataset` without requiring that it is in a `dcat:Catalog`.

chris-sepa:
@makxdekkers  True, anybody can say anything about any thing. Perhaps I should have just said - not in a Catalog so out of scope.
In thinking about this I've realised I've been assuming that an Application Profile is associated with a Catalog and that all Dataset records in that catalog would conform to the same profile. Therefore a standalone Dataset record wouldn't describe the profile it conformed to, so not very useful.  However, where data set metadata is harvested from multiple Catalogs they may well have been created to conform to differing profiles. This would lead to a need to associate profile information with a Dataset (or a CatalogRecord?). Alternatively, a Catalog constructed from harvested records from other catalogs could be partitioned into sub-catalogs for each different profile.

agbeltran:
@makxdekkers we were looking at this issue and the associated [use case](https://www.w3.org/TR/dcat-ucr/#ID35) that you had proposed, and we were not clear on how DCAT restricts a dataset to belong to a catalog - while the text in the spec is very catalog-centric, the RDF doesn't seem to impose any restrictions; were we missing something? And if there is currently a restriction on a dataset having to belong to a catalog, what did you have in mind for addressing this? Thanks

makxdekkers:
[...]

PR #784, opened 2019-03-02
DCAT - definition of accrualPeriodicity

dr-shorthair:
Proposed resolution of #728
Preview: https://rawgit.com/w3c/dxwg/dcat-issue728-simon/dcat/index.html#Property:dataset_frequency


agbeltran:
I suggest we further clarify the definition from "The frequency at which dataset is published." to "The frequency at which dataset is published ***in the catalog***" - to make explicit what is the collection is the catalog (with reference to the property definition "The frequency with which items are added to a collection").

agbeltran:
Here the html diff between this addition and the ED: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fdxwg%2Fdcat-issue728-simon%2Fdcat%2Findex.html#Property:dataset_frequency

andrea-perego:
@agbeltran said:

> I suggest we further clarify the definition from "The frequency at which dataset is published." to "The frequency at which dataset is published _**in the catalog**_" - to make explicit what is the collection is the catalog (with reference to the property definition "The frequency with which items are added to a collection").

I'm a bit concerned about this proposal, as it would imply that a dataset must be in a catalogue. This is not always the case. Moreover, the update frequency concerns the data, so it's an intrinsic characteristic of the dataset, with no relationship with the catalogue(s) (if any) where the dataset is documented.

agbeltran:
Good point @andrea-perego - so, if we leave the text as it is, do you agree with the modifications in this PR?

andrea-perego:
@agbeltran said:

> Good point @andrea-perego - so, if we leave the text as it is, do you agree with the modifications in this PR?

 +1 from me.

andrea-perego (approved):

agbeltran:
The point of datasets not in a catalogue is still a pending issue: https://github.com/w3c/dxwg/issues/62

agreiner:
"This may be complemented by  dcat:temporalResolution to give the rate at which items are added to a time-series." is ambiguous. (Is it added at publication, or added at data collection?) I think you must mean something more like "to give the time between collected data points in a time series". The two are only the same when the data are streamed.

andrea-perego:
I agree with @agreiner that temporal resolution is about data collection frequency. The proposed clarification makes sense to me.

agbeltran:
I have modified the statement about ```dcat:temporalResolution``` as suggested by @agreiner and agreed by @andrea-perego - I think with that the modification is ready and will merge the changes.

PR #811, opened 2019-03-08
Roles

nicholascar:
8 roles listed in main PROF doc. Definitions worded as verbs since roles imply actions, otherwise wording from Google Doc. A couple of additional Usage Notes (see Guidance).


rob-metalinkage:
please drop in a link to view the rendered changes when possible :-)

https://raw.githack.com/w3c/dxwg/roles/profilesont/index.html#resource-roles-vocab

rob-metalinkage:
Needs a follow-up discussion on the namespace (and/or inclusion of roles), however as this is consistent with current examples role: is a good starting point

PR #815, opened 2019-03-13
Conneg ap merge list tokens

rob-metalinkage:
removes list tokens method now integrated with list profiles.


larsgsvensson (approved):
The removal of list-tokens is fine with me. I won't approve the whole PR though since it goes into prof-land as well... Here I wait for @nicholascar 's review before we merge.

nicholascar (approved):

PR #822, opened 2019-03-15
Move schema.org alignment to annexe

dr-shorthair:
This follows the resolution in the 2019-03-10 meeting. 
The annexe needs a little more work, to verify that all properties added or modified since the original version are accounted for. 
Also - it refers to an RDF representation which axiomatizes the mapping using various SKOS, RDFS and OWL predicates, which might overreach.


riccardoAlbertoni (commented):
I dislike that the schema prefix is 's' in the figure 3 and 'sdo' in the rest of the appendix and in the related ttl code. Is there a specific reason I might have missed for that?  If not, I would suggest making it uniform.  If changing the prefix into the figure is more complex than in the appendix and code, we might uniform the prefix for schema.org to 's'. What do you think?

Also, If I have correctly understood the discussion in the teleconference we had two weeks ago, I think we decide to get rid of all the mappings except for schema and prov. Do I remember wrongly? if not, I would suggest deleting/commenting the whole appendix B.

riccardoAlbertoni (approved):

davebrowning (approved):

davebrowning:
Note: while broadly editorial, this PR does add a new example (dataset-004-sdo.ttl file in example directory)

agbeltran (approved):

Issue #642, opened 2019-01-09
prof:inheritedFrom needs more convincing case and/or example


aisaac:
Examples are now provided for this, however this concept still appears to be unclear to some. This issue remains open pending a round of feedback based on improvements to examples and definition.

Original post here (edited above as its what appears in the doc)

As in the title...
This came to mind while writing the ESWC paper.
The property is not super-clear in https://w3c.github.io/dxwg/profilesont/ and there's no example for it.

Actually one could question whether it is needed, or whether it is just a "utility properties" to be derived from other statements, the same way that skos:broaderTransitive is in SKOS. That is to say, that they could be kept for amateurs only ;-) and put in an "advanced" section.

andrea-perego:
+1 to @aisaac . I have the same concern.

kcoyle:
I also have the concern that the concept of "inheritance" here isn't clear. What is inherited, and what are the implications of that inheritance? Does inheritance mean that the constraints are not also included in the profile being defined but must be included from the "inheritedFrom" resource? This is made more difficult by the fact that the definition of "resource descriptor" is not very clear so it's hard to imagine what kind of thing/resource one would inherit from.

rob-metalinkage:
constraints from a profiled specification must be implemented by the profile. This is a transitive inheritance - so please come up with a new word and evidence of its use in the wild, or propose some better wording.

from wiki.. "In most class-based object-oriented languages, an object created through inheritance (a "child object") acquires all the properties and behaviors of the parent object (except: constructors, destructor, overloaded operators and friend functions of the base class). Inheritance allows programmers to create classes that are built upon existing classes,[1] to specify a new implementation while maintaining the same behaviors (realizing an interface), to reuse code and to independently extend original software via public classes and interfaces. The relationships of objects or classes through inheritance give rise to a directed graph. "

what is not clear about this?

aisaac:
Note that the original issue is for an example and motivation, not questioning the notion of inheritance. In fact there's hope that such elements would help discussing the notion of inheritance, if need be. For the moment I feel we'd better wait until then.

kcoyle:
@rob-metalinkage The issue that I raised is that it isn't clear what happens when:

ontologyA has properties A, B, C, D and class X
profileB includes properties A, D
profileB "prof:inherits" ontologyA (not clear what this would mean)

Does profile B also inherit ontologyA B, C?

[...]

Issue #485, opened 2018-10-23
prof:profileOf sub-property of prov:wasDerivedFrom (!?)


riccardoAlbertoni:
Is there any specific reason why the property prof:profileOf is not defined as a sub-property of prov:wasDerivedFrom?

nicholascar:
The relationship has occurred to us but we are not sure it’s Really a derivative relationship. It may be a subPropertyOf prov:wasInfluencedBy instead.

Happy to bash this one out here.

nicholascar:
From [PROV-O' explanation of wasDerivedFrom](https://www.w3.org/TR/prov-o/#wasDerivedFrom):

*"A derivation is a transformation of an entity into another, an update of an entity resulting in a new one, or the construction of a new entity based on a pre-existing entity."*

It is appropriate to say directly that `:DCAT-AP prov:wasDerivedFrom :DCAT` when taking about the provenance of DCAT-AP as a *thing*. Of course this dosent convey the special sense of what it is to be a profile of something else.

Should we use a straightforward `prof:profileOf rdfs:subPropertyOf prov:wasDerivedFrom` to tie profileOf into PROV?

To be continued...

riccardoAlbertoni:
+1 to @nicholascar  that was my implicit suggestion ;)

I think prov:wasInfluencedBy leaves prof:profileOf too underspecified. It actually subsumes generation and other relations that do not properly characterize prof:profileOf. Anyway, I am open to change my mind if anyone has an answer to my original question.

aisaac:
So I guess a resolution on this issue depends on how we close #507.
But I have to say that my experience with PROV is that it's likely that any property will include some domain or range declaration that will be quite strong. And that is going to bring us to endless and unnecessary discussions about whether the inherited domains and range make sense :-/

kcoyle:
Where does this leave the assumption that "profileOf" implies formal inheritance? See #642 for such a statement in [comment](https://github.com/w3c/dxwg/issues/642#issuecomment-452839500). There is no implication of formal inhertance in prov:wasDerivedFrom - especially since it includes that statement that the result can be a "new entity", which could include a stand-alone entity.

andrea-perego:
[...]

PR #827, opened 2019-03-18
Prof defns

nicholascar:
Defns from Google Doc incorporated for 2PWD


kcoyle:
Do NOT merge this. The group has not reached consensus on these changes.

nicholascar:
@kcoyle This PR should be merged now. 

No-one is saying this is a final state: what we are doing is taking improved definitions from the Google Doc (invisible to the public) and putting them in the HTML doc in time for a PWD (visible to the public) for further comment. If we don't do this, we risk reviewers seeing old definitions.

The position we adopted in plenary discussions recently was to merge smaller elements frequently. This is best practice development where there's a moving item (our document) whose latest version should be visible as soon as possible.

As I stated in the establishment of the Google Doc for definitions, after a round of improvements, that doc would be deleted and further docs could be created for further improvements. The time for transferring definitions and deleting the Google Doc is now so we can release a 2PWD and then we can do it all again for a 3PWD.

rob-metalinkage:
Note that sub groups and editors perform editorial changes and raise and close issues. The plenary needs to stay focussed on release processes and cross-cutting concerns. We do need to check off on responses being made to reviewers.  We are working asynchronously  via the google doc - and i think we are close to reaching a state where there are no active proposals for improvements to definitions. I have updated one in response to #485.  I think we can ask in plenary whether anyone who has not contributed so far has any urgent issues they wish to translate into concrete proposals to improve definitions, if not then accept and reset - we can always open new issues with new feedback.

kcoyle:
Here's an example:

Definition
Current: A Profile has a Resource Descriptor

Makx (Issue 755): An accessible form of the Profile with a certain role in a certain format
Karen (Issue 755): The described profile resource

Proposed: 

A resource which describes the nature of an artifact and the role it plays in relation to a profile

Makx offered some text, I offered some text, and then (unknown) wrote something entirely different and put it in the pull request. Where did this wording come from? Who agreed to it? Did the group working on this even see it before it was added to the document? This is totally out of left field, is not a consensus of the folks commenting - so why is it being put in the document? That's why I object to this change.

rob-metalinkage:
The editors ran a sprint and consolidate all comments into a proposal - which has been now out there for a week for  further comment. At some point we need to take snapshot and get wider review.  Feel free to suggest substantive changes to definitions or the PR. but we cannot just stall it.

rob-metalinkage (commented):
Please just confirm the PR matches the google doc and indicate the specific date. At this stage we have no unanswered questions and no concrete proposals to update proposed definitions, so we can snapshot and reset ready for further suggestions.

kcoyle:
It "matches the Google doc" because you wrote the text in the Google doc then added it here. It doesn't make use of the comments made by others in that document. I don't know about this "sprint" because there was no meeting announced, no agenda, and no minutes. Can we go back to the Google doc and try to gain consensus?

rob-metalinkage:
there was a doodle poll, and an announced sprint, as per the action in the plenary, at a typical time slot. We also chose the google doc as a way of making it open, and it has been effective at gathering useful comment. As editors we need to reconcile different suggestions - it wouldnt make sense just to transcribe them all. Largely we have accepted the best wording except where (as explained in comments) the sense was not quite correct -perhaps too general or too specific. At any rate  the suggestions have now been available for some time - we need to snapshot and start again. The PR doesnt reflect an agreed final state - it is a time-dependent snapshot of the best we can do so we can elicit wider feedback ASAP.  Otherwise we are violating (even worse) the W3C policy of publish often, I beleive the temporary google doc approach has allowed participation as intended, and as intended it needs to be wrapped up, and if necessary restarted to allow for wider participation.  

Please follow procedure and accept the PR (providing it matches the state of the google doc - which is explicity the editor's response to comments that have been available for review for over a week - and raise issues for specific definitions you would like to propose further improvements.

rob-metalinkage:
PS the statement "makes no use of suggestions" is palpaby incorrect - in most cases we have followed very closely Maxks original suggestions. What is the procedure for rejecting a review based on factual error?

nicholascar:
See this review of the changed document: https://raw.githack.com/w3c/dxwg/prof-defns/profilesont/index.html#specification

kcoyle:
Rob, that "sprint" was the one that was announced for times that no one could attend except you a Nick. The Doodle poll that we asked for was for a meeting the following week, not those same times. (All of this is in emails.) There was no agenda, no minutes - so we do not know who attended or what was decided. The W3C process requires openness - setting meetings for maximum attendance, reporting back on decisions, showing how decisions were made, providing minutes for record-keeping. 

The Google Doc is getting active comments, and I think with a little work we could set up some polls. Those polls would then be evidence of consensus reached. 

Meanwhile, I hope that Makx can weigh in because he was one of the people who wanted the definitions improved and he has commented quite a bit on the Google Doc. Nothing should go to commit unless the folks who are involved have been consulted. It isn't fair to them to ask them to work on it and then not get their agreement (active agreement) on changes. In the W3C process, silence is just that: silence. You need active consent.

rob-metalinkage:
The google doc - which people have been usefully contributing - is the mechanism to allow people to make suggested changes when we cannot easily convene. As W3C liaison points out the editor's draft does NOT need to represent a consensus position - as long as issues are noted. If a review opportunity is provided and a deadline passes editors need to make a call and progress matters. We explicitly stated we would accept inputs up till the next WD planned for last week.  I havent given consent explicitly for every DCAT commit and i wouldnt expect the editors to stop progress on that basis. The draft reflects the current state - we have an open issue, solicited feedback and are proposing new words in the draft.  Applying an alternative interpretation is ignoring the process.

makxdekkers:
@kcoyle @rob-metalinkage I don't have a big stake in this. I suggested some wording but do not feel strongly about those suggestions being taken up. The only one that I have a real problem with is `prof:isInheritedFrom` because I still don't understand what it is supposed to mean.

kcoyle:
Rob, I see a problem with communication here. I do not see evidence that there was a deadline announced to the group. Nor was there an announcement that the deadline had been reached and decisions had been made. To get better input, you need to encourage people to participate. Again, we also do not have transparency for any meetings related to PROF. We could let this go through on a purely procedural level, but there are big problems relating to process. I'll check with Peter and Philippe and get back to this.

kcoyle:
RE: Makx's comment - is there a way to note that in the draft? So we don't lose it.

kcoyle (approved):
I thought Makx would approve this, but since I didn't I'm sort of doing it for him. Makx, if you read this, I hope that's ok with you. It's how I interpreted your comment.

rob-metalinkage (approved):

rob-metalinkage:
OK - I will create a new PR with a reference to an issue to consider further improvements in deinition, explanation or examples for this specific item.

rob-metalinkage:
Actually it is already there #642 - I'll update the header comment of the issue that appears in the draft.

Issue #808, opened 2019-03-07
Formally align PROF to DCAT


nicholascar:
Formally include DCAT alignment in the Profiles Vocabulary. The alignment would be made like this: `prof:Profile rdfs:subClass of dcat:Resource .`

### Changes
* `prof:Profile rdfs:subClassof dcat:Resource .`
* `prof:hasArtifact` replaced by `dcat:downloadURL` and `accessURL`
* `prof:ProfileResource rdfs:subClassOf dcat:Distribution .`

NB : prof:hasResource is not a sub-property of dcat:distribution because the domain of dcat:distribution is dcat:Dataset.   Implementations may declare this is if required, to catalog profile descriptions using dcat:Dataset .  prof:Profile and dcat:DataSet are not disjoint.

#### Why now?
* DCAT had not previously defined abstract dcat:Resource
* It was not clear if something described in a document could be a dcat:Dataset
* It was open for debate whether a dcat:Distribution was “informationally equivalent” to a Dataset

### Pros
* Some PROF properties are able to be deprecated in favour of long-standing DCAT properties
* Other previously proposed PROF classes and properties can be interpreted in the light of DCAT classes and properties which are widely understood
* Use of DCAT indicates how to apply basic metadata properties to profiles since, if a prof:Profile is a dcat:Resource, then users would expect to see dtc:title etc. for it
dcat:downloadURL and accessURL is more nuanced that hasArtifact
* Resolves the debate about whether to force properties of distributions to be attached to the accessURL or allowed in the cataloged component (e.g. dct:format) 

### Cons
* PROF now has a dependency on DCAT (revised) which is a new (implication: somewhat un-tested) ontology
* PROF has to both differentiate its classes and properties from other DCAT classes and properties while still being DCAT-compatible

### Implications:
* Revisit if ProfileResource should be named more like Distribution (ProfileDistribution)

kcoyle:
"Resolves the debate about whether to force properties of distributions to be attached to the accessURL or allowed in the cataloged component (e.g. dct:format)"

Please say more about this, because I do not see how this "resolves the debate." What about this would change the validity of the use of dct:format for a catalog record rather than a digital file?

[...]

PR #828, opened 2019-03-18
Updated PROF alignments, as per DXWG wiki

nicholascar:
In time for 2PWD


rob-metalinkage (commented):
can you provide link to the preview please?

nicholascar:
See this preview of the changed document: https://raw.githack.com/w3c/dxwg/prof-alignments/profilesont/index.html#alignments

rob-metalinkage:
I think that the section "PROF is considered a specialisation of the revised version of the Dataset Catalogue Vocabulary [VOCAB-DCAT-2] for the purpose of cataloguing profiles. With this in mind, the main PROF classes - Profile and Resource Descriptor, specialised (are sub classes of) DCAT's Resource and Distributions respectively."   needs to be preceded with a statement that the alignment is not normative, but is provided as a recommended way to consider general metadata needs when describing profiles.

andrea-perego:
Not sure about this statement:

> [...] the fact that both prof:Profile and adms:Asset are non-disjoint sub classes of dcat:Resource, albeit that the latter is such via being a sub class of dcat:Dataset which subclasses dcat:Resource, means that resources could easily be dually typed as being of both prof:Profile and adms:Asset.

As prof:Profile is said not to be a subclass of dcat:Dataset (so I guess it is disjoint from it), whereas adms:Asset is a subclass of dcat:Dataset, I don't see how a profile can be also an adms:Asset (which I think it is).

Is there a strong motivation why prof:Profile shouldn't be made a subclass of dcat:Dataset?

nicholascar:
@andrea-perego:

> As `prof:Profile` is said not to be a subclass of `dcat:Dataset` (so I guess it is disjoint from it), whereas `adms:Asset` is a subclass of `dcat:Dataset`, I don't see how a profile can be also an adms:Asset (which I think it is).

Since `prof:Profile` is not declared as disjoint with `dcat:Dataset`, you can't conclude (guess) that it is disjoint, only that it's not been declared a subclass/equivalent etc. So, the logic holds here that, as I originally wrote, "...resources could easily be dually typed as being of both `prof:Profile` and `adms:Asset`".

> Is there a strong motivation why `prof:Profile` shouldn't be made a subclass of `dcat:Dataset`?

Only the general sense that a Profile isn't necissarily a Dataset. Again, Profile it's not declared not a Dataset, just no strong commitment is made.

andrea-perego:
@nicholascar said:

> Since prof:Profile is not declared as disjoint with dcat:Dataset, you can't conclude (guess) that it is disjoint, only that it's not been declared a subclass/equivalent etc. 

Probably not formally, but the note to the PROF - DCAT alignment says:

> prof:Profile is not a sub class of dcat:Dataset

which means that a profile is not a dataset.

So, I think the point needs to be clarified, or the statement above relaxed.

Said that, happy to approve this PR provided that this issue is addressed for the next release.

andrea-perego (approved):

rob-metalinkage (approved):
thanks.  note that the comments relating to DCAT alignment need to be picked up in issue #808. Note that DCAT alignment is non-normative - so its up to implementors to decide if they want a Profile to be equivalent to a Dataset.

PR #833, opened 2019-03-19
Prof motivation section includes links to UCR sections

nicholascar:
We have not included direct references to the UCR doc in PROF to date so this is Stage 1 in addressing that. A further stage might be to call out individual Requirements and/or Use Cases at various points in the doc.


nicholascar:
See this preview of the changed document: https://raw.githack.com/w3c/dxwg/prof-voc-ucs/profilesont/index.html#motivating-requirements

rob-metalinkage (approved):
Yep - simple and obvious improvement. Editorial changes may be required in formatting but these can be picked up later.

kcoyle (commented):
Although I think it is odd that it mentions the Conneg document since it also says that it doesn't address that. In the future I would advise removing that from the text - it doesn't really make sense to have it there.

kcoyle (approved):
Along with my earlier comment

aisaac (approved):
I'm fine with these changes. I too think we might prefer to remove the mention to Conneg in the future, but for now we're probably better being safe and clarify the relationship between our different docs. I guess a reader could currently ask the question, of what happens with the conneg reqs, when they see the references to the other profile reqs.

Issue #840, opened 2019-03-24
Consider naming conventions for instances to make examples easier to read


rob-metalinkage:
#642 provides an example - a suggestion that naming conventions might make examples easier to follow was provided - this issue provides for the opportunity to get feedback on this particular issue. 

Note - a PR needs to be prepared and accepted for the provided examples so that these can be reviewed in the light of this issue.

PR #841, opened 2019-03-24
update issue reference

rob-metalinkage:
now example exists, issue #642 can be closed, and the substantive issue is now #840 and #842 opened up for discussion.


nicholascar (approved):
I'm happy with the Issue updates

kcoyle:
"Give the original issue has been addressed with an example" - @rob-metalinkage I believe that there were edits requested to the example, in particular ones requested by @aisaac. It doesn't look like those were done. Were they?

rob-metalinkage:
there is a new issue to consider those edits #840 - they are general stylistic edits around naming that apply to the set of examples - and to consider it properly people will need to see the examples, and the original issue was tangled up with multiple threads and concerns, so all we are doing is updating this so that we can move forwards with single concern issues. Can we please just follow a process rather than try to solve every problem under every issue and review cycle.

aisaac (approved):
As per my email at https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Mar/0712.html, considering that I was the only one objecting to the merge during our last call (and now, as it appears, for bad reasons), and considering that @kcoyle 's mail (https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Mar/0712.html) leaves the door open to basic merges this week, I am in favour of merging this PR, which is after all only acting the replacement of one rather messy issue with two cleaner ones.
By the way I am going to  try to populate #840 with the relevant stuff from @kcoyle and I in #642

PR #850, opened 2019-03-27
Prof 2 pwd

nicholascar:
Final PR with link fixes due to W3C validators for 2PWD.


kcoyle (approved):
I see that the Dublin Core term links go to, e.g.,  http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-Standard rather than the canonical names like http://purl.org/dc/terms/Standard. I'll check on this, but I'm pretty sure that the purl.org namespace is the one that must be used. Can we make it an issue for the future so we don't forget? I don't think it matters for the draft.

Also, Philippe suggested that we add a message in the drafts themselves saying that they are on hold but I never got around to asking him where that should be - if there's a logical place in the draft header. That could be a separate PR. I'll ask him today.

nicholascar:
@kcoyle I wondered about the use of http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-Standard rather than http://purl.org/dc/terms/Standard however I chose the former since the latter doesn't support any sort of fragment redirect. If you click on the latter, you just go to the whole spec, not the Standard part. In fact, since a DCMI server update, the former doesn't even take you to the spec, only the folder containing it!

From what I see elsewhere in W3C docs, HTML page links use descriptive page links, the former, rather than canonical namespace links, the latter.

So I'd prefer the former for these two reasons.

However, if you have leverage at DCMI, perhaps you could ask them to update their purl.org redirects to go to the spec fragment. That way, we can just use the canonical URL and still get the correct result.

rob-metalinkage (approved):
see what you mean about the DCMI links - and agree the best solution would be for the canonical links (purl) to redirect according to best linked data practice - and that is out of our control.

kcoyle:
Nick and Rob - DCMI has made a conscious choice to not re-direct to the "fragment". That doesn't change the fact that purl.org is the canonical namespace. Like others, DCMI is noticing a lot of problems with the PURL server and is thinking about options, but has used PURL for over a decade so any change will take a lot of thought. I'll ping Tom Baker.

rob-metalinkage:
Standing by for a better way if it emerges...

tombaker:
@nicholascar wrote: 
> In fact, since a DCMI server update, the former doesn't even take you to the spec, only the folder containing it!

This was indeed misconfigured in the new website and has been fixed.

According to the [DCMI Namespace Policy](http://dublincore.org/documents/dcmi-namespace), first published in 2001, metadata terms are persistently identified using PURLs.  DCMI makes no such commitment to the persistence of any other Web addresses, which function as locators (URLs) for Web pages, not as persistent identifiers, and could in principle change at any time.

The PURL service does suffer occasional outages, but as @kcoyle hints, it would be disruptive to adopt new persistent identifiers for DCMI metadata terms - something not to be undertaken lightly.

DCMI's position is that URLs such as  http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-Standard  should never be used to _identify_ `dct:Standard`.  In principle, one could write
```
[http://purl.org/dc/terms/Standard](http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-Standard)
```
so that the PURL displays as the visible text, but since http://purl.org/dc/terms/Standard now resolves correctly, there would be no practical reason to do so.

nicholascar:
@tombaker thanks for the context. Yes http://purl.org/dc/terms/Standard resolves but only to the page, not as a fragment to the actual item Standard. I like your suggestion of `[http://purl.org/dc/terms/Standard](http://www.dublincore.org/specifications/dublin-core/dcmi-terms/#terms-Standard)` since we can then make visible the identifier and link to the specific item for convenient use (clicking on it). I’ll try and roll that out for DCT things for the 3PWD.

Issue #802, opened 2019-03-06
Does "profile of" require all or some elements?


kcoyle:
*From the Shex community*: It is also unclear to us whether it is helpful to describe the relation of a profile to its underlying namespaces in terms of a relation of Profile to Standard.  The standard-to-profile relation is characterized as one of general-to-specific, but when namespaces simply provide just a handful of their terms to a profile, it seems like a stretch to say that the profile is "profiling" each of the namespaces. https://lists.w3.org/Archives/Public/public-dxwg-comments/2019Feb/0001.html

kcoyle:
From the profgui meeting of Feb 28:

2. "profile of"
What conditions entail that one thing is a profile of another thing?
a. - any use of elements (classes, properties), even if minimal (e.g.
just use one DC term)[4 - Tom Bakers response]
b. - use of all of the elements from a "base specification" and possibly
add elements from another another specification and/or other constraints.
c. - use of a "significant portion" of a base specification or other
profile.
d. - something is a profile of something else if the creator of the
profile declares it as such

Note that if there is to be any formal inheritance then some of these
definitions would not be usable.

Related issues:
https://github.com/w3c/dxwg/issues/486

https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Mar/0014.html

kcoyle:
response from Lars:

a. is definitely too weak, d. feels a bit arbitrary. I tend to b. rather than c. even though I understand that there are corner cases where this can be hard. (When it comes to machine-processability I tend towards relatively strict definitions...).

https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Mar/0079.html

kcoyle:
Response from Makx:

I agree with Lars that b. makes the most sense. For example, the European DCAT-AP mentions each and every property in DCAT-2014 with a usage note. A profile could state "don't use property x" but it should say something about all terms from the specification it is a profile of.

Makx

https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Mar/0080.html

smrgeoinfo:
I like the approach that 
1) a profile is a specification that restricts one or more other specifications in some way.
2) a specification consists of rules/conformance classes/constraints (choose your favorite terminology) that dictate how some particular data resource can be validated as conforming to the specification.
[...]

PR #861, opened 2019-03-28
Updates to profile guidance sections 2.2 & 2.3

aisaac:
Reflecting agreement in our last call

New text: https://rawgit.com/w3c/dxwg/aisaac-guidance-2.2-2.3/profiles/#RPFMBSPEC
Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fprofiles%2F&doc2=https%3A%2F%2Frawgit.com%2Fw3c%2Fdxwg%2Faisaac-guidance-2.2-2.3%2Fprofiles%2F

--


larsgsvensson:
If this was agreed in the last call, I guess it's OK.

Personally, though, I have some issues with the following text;

> conforming to the DPLA Metadata Application Profile means conforming to the Europeana Data Model, Dublin Core terms and Open Archives Initiative Object Reuse and Exchange. 

since I don't consider Dublin Core terms a profile (has no constraints/cardinalities etc. as opposed to the [Dublin Core Collection Description Application Profile](http://www.dublincore.org/specifications/dublin-core/collection-description/collection-application-profile/)). I'm not sure about OAI-ORE. I would say that EDM _uses_ DCTerms and OAI-ORE, not that it's a profile of either of them. But I might have missed some of the discussion leading up to this agreement...

aisaac:
Hi @larsgsvensson we're not saying that DC is a profile (which is what is hinted in one of your sentence). This said, I'm willing to recognize there could be discussions on whether EDM is a profile of DC or if it uses it. In fact I think #802  discusses this, i.e. how much 'use' of a standard is needed to qualify a profile as a profile of that standard. But I'm entirely comfortable imposing one view in the current document for now and reverting what's written if there's consensus against it when we people besides the group that met and discussed the examples. The EDM case added in this PR is a mere example for something else, not as an argument for what makes up a profile and not.

Basically as long as issue #802 is open I believe we're ok. And then the document will have to abide to the resolution for #802, which could imply changing the example. Again I'm totally fine with it.

larsgsvensson:
Hi @aisaac 

> Basically as long as issue #802 is open I believe we're ok. And then the document will have to abide to the resolution for #802, which could imply changing the example. Again I'm totally fine with it.

Works for me. That said, I think we in that case should point to #802 so that it's clear that there is controversy on what is a profile and how to use the terms "inherit", "use" etc.

aisaac:
@larsgsvensson fair enough! I've referenced issue #802 at
https://rawgit.com/w3c/dxwg/aisaac-guidance-2.2-2.3/profiles/#RPFPP

larsgsvensson:
Thanks @aisaac, that resolves my comment.

rob-metalinkage (approved):
I think the issue of "conformance to DC" is the vagueness that complicates the matter. If you have a specification such as a vocabulary where everything is optional, then anything that doesnt use DC terms is conformant. Things that use DC terms have to be conformant to axioms (such as range) but also the semantics as described.  So really a conformance test for DC might need to be expressed in the form of "uses at least one term from DC and every use of DC terms is consistent with DC semantics and axioms".   RDF/S + OWL  does tend to have a lot of implications about trivial cases that are not so intuitive outside theoretical concerns - such as A rdfs:subClassOf A - and what we are really saying by stating something is a profile of DC is that there exists conceptually some sense of conformance to DC, even if its not formally expressed. DC is a profile of itself for example, since the set of constraints can technically be empty.

aisaac:
Thanks @rob-metalinkage and @larsgsvensson .

@larsgsvensson does this mean you approve the PR?

@rob-metalinkage I agree the vagueness is an issue, but it goes beyond this PR - and hopefully is captured in all the other issues discussing what profile are and what they enforce!

larsgsvensson (approved):

larsgsvensson:
> does this mean you approve the PR?

Yes!

aisaac:
Thanks, @larsgsvensson !
This is a bit later than I had planned when I submitted the PR (wrt the 'freeze' on profiles) but I think it's better if I merge it before we re-open the discussions in several weeks.

PR #872, opened 2019-04-01
editorial changes to remove stray mentions of data services that are …

agreiner:
…not APIs

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -117,8 +117,8 @@ $(document).ready( function() {
     </p>
 
     <p>
-        Data is often provided through a service accessed through a form or API which supports selection of an extract, sub-set, or combination of data.
-        DCAT allows the description of a data access service to be included in a catalog.
+        Data is often provided through a service accessed through an API, which supports selection of extracts, sub-sets, or combinations of data.
+        DCAT allows the description of such a data access service to be included in a catalog.
     </p>
 
     <p>
@@ -264,24 +264,20 @@ $(document).ready( function() {
         A <b>dataset</b> in DCAT is defined as a "collection of data, published or curated by a single agent, and available for access or download in one or more serializations or formats".
         A dataset is a conceptual entity, and can be represented by one or more <b>distributions</b> that serialize the dataset for transfer.
         Distributions of a dataset can be provided via <b>data services</b>.
-        Detailed properties for a data service API are out of the scope of this version of DCAT.
     </p>
 
     <p>
-      A <b>data service</b> in DCAT is a collection of operations or API which provides access to data.
-      An interactive interface may provide access to the API operations.
+      A <b>data service</b> in DCAT is a collection of operations, made available through an API, which provides access to data.
       A data service typically provides selection, extraction, combination, processing or transformation operations over datasets that might be hosted locally or remote to the service.
       A data service might be tied to specific datasets, or its source data might be configured at request- or run-time.
       The result of a request to a data service is a representation of a part or all of a dataset or catalog.
       A data service might represent a data distribution service to select and download a distribution of a dataset or subset, or a data discovery service to find the description of a suitable dataset.
-      Other kinds of data service might include data transformation services such as coordinate transformation services, re-sampling and interpolation services, and various data processing services, but these are outside the scope of the current DCAT vocabulary.
     </p>
     <p>
       Descriptions of datasets and data services can be included in a <b>catalog</b>.
       A catalog is a kind of dataset whose member items are descriptions of datasets and data services.
-      Other types of thing might also be catalogued, but the scope of DCAT is currently limited to datasets and data services.
+      Other types of things might also be catalogued, but the scope of DCAT is currently limited to datasets and data services.
       To extend the scope of a catalog beyond datasets and data services it is recommended to define additional sub-classes of <a href="#Class:Resource"><code>dcat:Resource</code></a> in a DCAT application profile or other DCAT application.
-      To extend the scope of service descriptions beyond data distribution services it is recommended to define additional sub-classes of <a href="#Class:Data_Service"><code>dcat:DataService</code></a> in a DCAT application profile or other DCAT application.
     </p>
 
     <div class="note">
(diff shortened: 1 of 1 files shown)


dcat/index.html
```diff
@@ -264,24 +264,20 @@ <h3>DCAT scope</h3>
         A <b>dataset</b> in DCAT is defined as a "collection of data, published or curated by a single agent, and available for access or download in one or more serializations or formats".
         A dataset is a conceptual entity, and can be represented by one or more <b>distributions</b> that serialize the dataset for transfer.
         Distributions of a dataset can be provided via <b>data services</b>.
-        Detailed properties for a data service API are out of the scope of this version of DCAT.
```
dr-shorthair, line 269: This is a clarification and does no harm. It is in a section that is explicitly about the scope of DCAT, so should stay.

dcat/index.html
```diff
@@ -264,24 +264,20 @@ <h3>DCAT scope</h3>
...
     </p>
 
     <p>
-      A <b>data service</b> in DCAT is a collection of operations or API which provides access to data.
-      An interactive interface may provide access to the API operations.
```
dr-shorthair, line 274: Is this not correct? It does not say that the interactive interface is the service, merely that an interactive interface might be available to provide access (i.e. an application that uses the API).

dcat/index.html
```diff
@@ -264,24 +264,20 @@ <h3>DCAT scope</h3>
...
+      A <b>data service</b> in DCAT is a collection of operations, made available through an API, which provides access to data.
       A data service typically provides selection, extraction, combination, processing or transformation operations over datasets that might be hosted locally or remote to the service.
       A data service might be tied to specific datasets, or its source data might be configured at request- or run-time.
       The result of a request to a data service is a representation of a part or all of a dataset or catalog.
       A data service might represent a data distribution service to select and download a distribution of a dataset or subset, or a data discovery service to find the description of a suitable dataset.
-      Other kinds of data service might include data transformation services such as coordinate transformation services, re-sampling and interpolation services, and various data processing services, but these are outside the scope of the current DCAT vocabulary.
```
dr-shorthair, line 279: Is this misleading or unhelpful? I wrote this to provide some orientation to a significant class of users from the geospatial domain.

dcat/index.html
```diff
@@ -1889,7 +1885,7 @@ <h4>Property: access service</h4>
             <table class="definition">
                 <thead><tr><th>RDF Property:</th><th><a href="http://www.w3.org/ns/dcat#accessService">dcat:accessService</a></th></tr></thead>
                 <tbody>
-                <tr><td class="prop">Definition:</td><td>A site or end-point that gives access to the distribution of the dataset</td></tr>
+                <tr><td class="prop">Definition:</td><td>A data service that gives access to the distribution of the dataset</td></tr>
```
dr-shorthair, line 1902: how about just 'service' or 'end-point'?

dcat/index.html
```diff
@@ -2103,7 +2099,7 @@ <h3>Class: Data Service</h3>
         <table class="definition">
             <thead><tr><th>RDF Class:</th><th><a href="http://www.w3.org/ns/dcat#DataService">dcat:DataService</a></th></tr></thead>
             <tbody>
-            <tr><td class="prop">Definition:</td><td>A site or end-point providing operations related to the discovery of, access to, or processing functions on, data or related resources</td></tr>
```
dr-shorthair, line 2120: Agree that 'site' has potential to imply web-page, so should be removed.

dcat/index.html
```diff
@@ -2103,7 +2099,7 @@ <h3>Class: Data Service</h3>
...
+            <tr><td class="prop">Definition:</td><td>A collection of operations accessible through an application programming interface (API) that provide access to one or more datasets or data processing functions. </td></tr>
```
dr-shorthair, line 2116: However, I don't quite understand what the phrase "accessible through an application programming interface (API)" adds to "collection of operations". Furthermore, the term "end-point" was used to make it clear that a dcat:DataService is an _instance of_ a data service, as opposed to a _specification for_ a data service, so I'd like to see that stay in the definition. Is it perhaps just that there is competing jargon/terminology coming from different domains?

dcat/index.html
```diff
@@ -2128,10 +2124,10 @@ <h4>Property: endpoint description</h4>
             <table class="definition">
                 <thead><tr><th>RDF Property:</th><th><a href="http://www.w3.org/ns/dcat#endpointDescription">dcat:endpointDescription</a></th></tr></thead>
                 <tbody>
-                <tr><td class="prop">Definition:</td><td>A description of the service end-point, including its operations, parameters etc.</td></tr>
+                <tr><td class="prop">Definition:</td><td>A description of the services available via the root endpoint and its children, including their operations, parameters etc.</td></tr>
```
dr-shorthair, line 2141: Hmm. 'root endpoint and its children' introduces some new concepts, and assumptions about the structure of data service endpoints. I don't think we want to go there.

dcat/index.html
```diff
@@ -117,8 +117,8 @@ <h2>Introduction</h2>
     </p>
 
     <p>
-        Data is often provided through a service accessed through a form or API which supports selection of an extract, sub-set, or combination of data.
-        DCAT allows the description of a data access service to be included in a catalog.
+        Data is often provided through a service accessed through an API, which supports selection of extracts, sub-sets, or combinations of data.
```
dr-shorthair: Whoops. I thought 'a form' had been removed. Must have got dropped in a merge conflict somewhere.

dcat/index.html
```diff
@@ -117,8 +117,8 @@ <h2>Introduction</h2>
     </p>
 
     <p>
-        Data is often provided through a service accessed through a form or API which supports selection of an extract, sub-set, or combination of data.
```
dr-shorthair, line 122: Whoops. I thought 'a form' had been removed. Must have got dropped in a merge conflict somewhere.

dr-shorthair (changes requested):
See comments on individual changes

dcat/index.html
agreiner: I think it's confusing, because it suggest we haven't addressed the use of DCAT to describe APIs, though we have done that. We define a data service, and in doing so we provide detailed properties.

dcat/index.html
agreiner: While technically true, it confuses the issue. We are defining a DataService as an API. If we kept this, we would need to explain that an exposed API itself would need to be the thing being catalogued.

dcat/index.html
agreiner: I don't understand why you would put it in and then say that it is out of scope.

dcat/index.html
agreiner: The access service is a data service, which is not a site but an API. Defining it as a "service" isn't offering much in the way of information. Why not be more helpful and say that it is a data service?

dcat/index.html
agreiner: I think we want to talk about these things with terminology that is accurate. Most web services offer multiple endpoints, one of which is the root endpoint.

dcat/index.html
agreiner: I think it was in multiple places.

dr-shorthair:
https://github.com/w3c/dxwg/pull/908 addresses these proposed changes in a new branch. 

The head has moved on quite a bit since @agreiner's fork, and I can't merge there. 
Also, I wanted to modify some of @agreiner's suggestions slightly. 
So I made a new branch in this repository and re-implemented the required changes there.

dr-shorthair:
@agreiner - #908 has now been merged by the editors. Can you inspect the modifications there to verify that your concerns have been resolved?

agreiner:
I think it's fine.

dr-shorthair:
Superseded by #908

PR #873, opened 2019-04-02
Move Linked Data Platform into Appendix

davebrowning:
As per discussion in #254 at https://github.com/w3c/dxwg/issues/254#issuecomment-478317449


davebrowning:
Changes as discussed in #254 - checking my changes to the examples is both good and necessary.  Obviously `dcat-ld:` doesn't exist... is there a better way to express this?

davebrowning:
@jakubklimek - in the (time limited ) push to get to a proposed recommendation we've had to slim down a lot of the potential alignment work and ensure we're very clear in what we're including in the immediate release.  This includes the LDP relationship work, so I've drafted some mods..  Comments very welcome if you can make the time.

davebrowning:
Rendering available [here](https://raw.githack.com/w3c/dxwg/dcat-issue254-daveb/dcat/index.html#other-w3c-recommendations)

andrea-perego (commented):
@davebrowning , after having reviewed again this section, I'm even more convinced that it needs to be consolidated and further reviewed by the WG to be safely included in the CR.

My suggestion is to move it to the backlog, and address it as one of the future DCAT priorities.

riccardoAlbertoni:
davebrowning wrote:
> Changes as discussed in #254 - checking my changes to the examples is both good and necessary. Obviously `dcat-ld:` doesn't exist... is there a better way to express this?

I think it might be better to use the namespace "dcat-ldp" instead of "dcat-ld", "dcat-ldp" makes probably clearer that is the namespace for linked data platform stuff and not other kinds of linked data publication. 

 About the inclusion in the CR, I am split, I think it would be great to have something helping in the direction of SOLID,  but I have to admit my limited confidence with LDP. So it is ok to have it in the future priorities as suggested by @andrea-perego.

davebrowning:
I've brought this up to date and incorporated the simple suggestions.  Suggest we discuss at the next quorate sub-group meeting (or plenary - whichever gets around to it first).

dr-shorthair:
The proposed additional predicates are quite simple and obvious I think so I'm not too worried about breaking anything. But should be marked non-normative. 
Namespace proliferation is undesirable, but may be unavoidable.

dr-shorthair (approved):
Move to informative annex is an improvement over the current situation so should be done immediately. 
We can then decide whether it stays at all.

riccardoAlbertoni (approved):

Issue #377, opened 2018-09-20
Links across DCAT rec, rdf files and namespace pages


davebrowning:
The draft DCAT-rev (FPWD, editors draft, second PWD) use links in the same way as the [2014 DCAT recommendation](https://www.w3.org/TR/2014/REC-vocab-dcat-20140116/) does.  In general this means that

1. Most links in textual sections (especially if without a namespace prefix such as `dct:` or `dcat:`) use intra-document fragment links to the appropriate section in the Recommendation document.
2. Within the vocabulary specification in the Definition, Domain and Range fields, the entries link either to an rdf file (e.g [rdfs:literal](http://www.w3.org/2000/01/rdf-schema#Literal)) or to a textual spec (e.g. [foaf:homepage](http://xmlns.com/foaf/0.1/Document) - though this particular link isn't quite precise) or to the namespace document for DCAT (e.g.[dcat:Catalog](http://www.w3.org/ns/dcat#Catalog)).  This last then requires the user to choose from either an Turtle or RDF/XML rendering of the ontology.

In addition, our namespace document at http://www.w3.org/ns/dcat# barely passes muster against the policy https://www.w3.org/2005/07/13-nsuri.  More compliant human-readable documents can be found at https://www.w3.org/2009/08/skos-reference/skos.html, https://www.w3.org/XML/1998/namespace, https://www.w3.org/2001/XMLSchema amongst others.  [Machine readable versions seem more difficult to find]

As part of moving DCAT-rev forward, I think it would be useful to readers of the recommendation if

- we had a consistent approach to where the links in the document take you - either to a rdf files or to more explanatory documents targeted at humans across all of the references.  [Assuming suitable targets can be found, of course]

- We expand the namespace document to make it more useful to users/align with current standards and policies, and to save the extra 'click' needed to get anywhere useful.

I'm looking for a steer here from interested parties on what approach would be helpful, especially where tooling/machine readability may play a role - so suggestions on what we _should_ provide would be extremely useful, as well as any input from the profile guidance work etc.

davebrowning:
Looking for discussion and opinion on this:

As preparation for any new WD (or candidate rec etc), we should ensure that we are happy that any link use is consistent and useful.  This approach in the existing ED is in part as discussed in #671

[...]

Issue #725, opened 2019-02-01
Reflect all 'Usage notes' into DCAT RDF representation  


dr-shorthair:
The representation of `Usage notes` found in the rec document as `skos:scopeNote` elements in the RDF representation (i.e. the .ttl file) is incomplete for terms from external namespaces (mostly DCT). This should be remedied. It requires an entry for each term that we use that includes only a `skos:scopeNote` for each `Usage note` seen in the HTML document.

dr-shorthair:
Maybe also the `See also` entries that we have included in the normative vocabulary description (i.e. the tables in chapter 6)

dr-shorthair:
And the `Domain` and `Range` where they are different (tighter scope) than in the original namespace.

dr-shorthair:
This is a relatively mechanical task but still needs executing.

jakubklimek:
During the translation to Czech, I came across the annotations of external vocabularies and there is an issue with this approach. We SHOULD NOT annotate external properties and classes (such as DCTERMS) with DCAT specific texts, as those statements then apply to ALL instances of those properties, regardless of DCAT.

Example: I have an endpoint where I use `dcterms:issued` for the date a book was issued in some library dataset. Now I use the same endpoint to store some DCAT metadata, and I load the DCAT vocabulary there. Suddenly, when someone looks at a record about a book, they see for instance `This indicates the date of listing the dataset in the catalog and not the publication date of the dataset itself.`, which makes no sense in this context.

Therefore, if we absolutely need to state this information about the properties, we would have to create DCAT subproperties for everything and make the statements about them. Otherwise we should refrain from stating specific things about external vocabularies to avoid creating mess.

The argument that "the user can choose to separate DCAT vocabulary with those annotations into a separate graph` IMHO does not hold, as we should not presume any particular named graph use case.

In addition, this problem can be seen even in the context of DCAT alone, where we have e.g.:
```turtle
dct:issued skos:definition "Date of formal issuance (e.g., publication) of the item."@en .
dct:issued skos:definition "Date of formal issuance (e.g., publication) of the distribution."@en .
dct:issued skos:definition "The date of listing (i.e. formal recording) of the corresponding dataset or service in the catalog."@en .
```
with no way of distinguishing which definition goes with which usage of the property.

kcoyle:
[...]

PR #881, opened 2019-04-09
Updated namespace file

davebrowning:
Updates namespace file for DCAT 2, supporting both the new rec and the older 2014 one, as per #377


dr-shorthair:
There is a note at the beginning of the formal specification https://w3c.github.io/dxwg/dcat/#RDF-representation inserted early in our drafting which suggests that the RDF representation would be modularized. 
This was mostly in relation to a possible dcat-prov alignment axiomatization, also dcat-schema.org, but also potentially related to a DCAT-2014 profile. So we need to consider if these are good, and whether they are normative, and have a story about them to drop into this artefact?

dr-shorthair:
Needs to also provide links to:
- https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-prov.ttl 
- https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-schema.ttl

We also need to decide if these are normative or non-normative

dr-shorthair (approved):

davebrowning:
This comment summarises our intentions here.

The 2014 recommendation contains links to the namespace page at `https://www.w3.org/ns/dcat#`.  (For example, in the Class tables at https://www.w3.org/TR/2014/REC-vocab-dcat-20140116/#class-catalog.).  By default this returns a simple namespace page that provides links to the rdf definitions in either turtle or rdf/xml.  The namespace also supports these alternate forms directly via the `link` header.

DCAT2 uses the same name space (so we aren't introducting a new version in the URI) and we are providing replacement turtle and rdf/xml that are backward compatible, some other representations/encodings and a new namespace file.  Simply renaming the existing resources (by addtition of 2014 perhaps) and replacing them with new resources means that links from the 2014 spec will show the new resources.  Is this acceptable?   We can ensure that the turtle. etc  files and the namespace page include appropriate text to guide any human user, but we'd appreciate guidance on doing this for machine access.  Is something like ADMS (e.g. https://www.w3.org/TR/vocab-adms/#adms-last) appropriate and/or required ?  What would this mean for what we'd need to release?

riccardoAlbertoni:
> Needs to also provide links to:
> 
> * https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-prov.ttl
> * https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-schema.ttl
> 

As discussed in #725, we should also add a reference to 
 * https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat-external.ttl

and as discussed in a teleconference ( I can't recall which one) we might want to refer to the former 
 version of dcat as well.
* https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat2014.ttl 

> We also need to decide if these are normative or non-normative

I think they are all non-normative.

davebrowning:
And it should probably link to the json-ld examples from #1062 as well.

Actually - are we happy that those links are to the master branch in github....?  Might be easier to manage if the 'live' versions were served from the W3 system that holds the actual req.

davebrowning:
I don't think this is yet complete, but wanted to open it to wider review.  It can be seen at
https://raw.githack.com/w3c/dxwg/dcat-namespace-page/dcat/rdf/dcat.html

At this stage, the idea is that the namespace file would replace the existing one at https://www.w3.org/ns/dcat.html.  In addition, we would release (also in  https://www.w3.org/ns)

1. A new version of the file dcat.ttl (containing the V2 ttl) overwriting the 2014 one
2. A new version of the file dcat.rdf (containing the V2 RDF/XML) overwriting the 2014 one
3. A copy of the 2014 turtle called dcat2014.ttl
4. A copy of the 2014 RDF/XML called dcat2014.rdf.
5. A copy of the documentation in dcat-external.ttl

If we add a json-ld version  (as suggested in #1062 [here](https://github.com/w3c/dxwg/pull/1062#issuecomment-532683222) then it would follow the same pattern.

We do need to decide whether the examples directory is published onto the webserver or whether we just link back to github.  If we do publish onto the server then we probably need to consider a sensible directory structure.  At the moment we are linking to github gh-pages branch.

Comments & improvements please...

davebrowning:
Links for rdf/xml and jsonld versions of current RDF added (those files to be provided as part of #1062).

riccardoAlbertoni (approved):

riccardoAlbertoni:
> We do need to decide whether the examples directory is published onto the webserver or whether we just link back to github.

I think linking to github might be enough, but I am open to other views.

kcoyle:
I'll add the question of publishing examples (github or ??) to the list of questions that Peter and I are keeping. We'll need to get together with Philippe for these, so let us know if you have others.

davebrowning:
> 
> I think linking to github might be enough, but I am open to other views.

I'm coming round to that view - originally I was concerned on how the 'evergreen' process might work in github if gh-pages (ie the master) is live.  There are ways to manage that but it kind of depends on how 'continuously integrated' the evergreen process intends to be.  [Jumping ahead, of course.....]

rob-metalinkage (approved):
I _think_ i approve - but this PR has some rather involved discussion and its not very easy to see a summary of what exactly is proposed beyond the sub-groups recommendation to update the content the namespace resolves to in a backwards compatible way. If there is more to it than that it probably needs to be made clearer.

davebrowning:
@rob-metalinkage Thanks.  At the moment the primary aim is to make sure we don't break backward compatibility.  

If there is a need to provide support for people using these automatically, then I'd hope we can progress that since it wouldn't be part of the actual recommendation (so could be treated as a 'WG note' kind of activity).  Of course, we should probably get better guidance/feedback on that  as we go through the CR stage.  As it is this is (IMHO) providing equivalence to 2014, and is broadly in line with what I can see other WGs have published in the last 12/18 months

PR #898, opened 2019-04-15
making qsa realization for conneg non-normative

agreiner:
I modified the introduction to the QSA info and also reworked the text to make it clear that other URL schemes are okay. The last sentence of the QSA section was cut off, so I'm not sure what was missing. I filled in some text that seemed close to what would have been intended.


nicholascar (approved):
Looks pretty good - I like the changes Annette.

I've just dropped back to QSA in the Realization's title as this is necessary to keep the realization confined since the wording you had might crowd out other URL-based approaches. Also, I removed a change you made about using "dataset" in examples as this shouldn't be limited to datasets.

larsgsvensson (approved):
Yes, this reads nice and balanced. Thank you @agreiner and @nicholascar

agreiner:
No problem to change the word dataset there. My intention was to make the URL match what was described. The original isn't at all valid, which obscures the point you're trying to demonstrate. I think my "bad" example has an error, too, since it doesn't have the ?. It's possible I'm not understanding your point, though. It seems to me that one could legitimately use additional key/value pairs that are exposed to the client.

Issue #838, opened 2019-03-22
Cleaning the CR from references to backlogged and open issues


riccardoAlbertoni:
I am trying to clean the document from the references to issues,  If merges must be endorsed in the plenary I think we have to make the most in term of cleaning before it.

I have prepared two branches

[dcat-deleting-issues-riccardo](https://github.com/w3c/dxwg/tree/dcat-deleting-issues-riccardo): in which I have deleted the references not anymore pertinent, namely
- ISSUE 252: Examples of quality for services, as we moved it in a future primer
- ISSUE 411: Distributions, services and implementation-resource, as it was closed

Getting rid of these two should not be much problematic. 
Please revise the related PR so that we can merge it in the next plenary.

[dcat-commenting-backlogissues-riccardo](https://github.com/w3c/dxwg/tree/dcat-commenting-backlogissues-riccardo):
In which I have commented with the string "!--backlog"  the reference to the issues that have been put in the backlog, so that they are not anymore listed in the document. Instead of mention the single issues in the backlog, I have proposed to add a sentence in the document in which we warn the reader about the backlog ( see issue #837  "Mentioning the Backlog in the DCAT CR").  However, the string   "!--backlog"  is a hook to retrieve these commented issues back in the case we want to turn them in notes instead of making them disappear from the document.
 This branch has made invisible the following issues:
- issue 105:  Use owl:Restriction constraints to bind DC properties to DCAT classes
- issue 62: Datasets vs. Catalog relation [RDSCR]
- issue 71: Project context [RPCX]
- issue 70: Publication control [RPC]
- issue 78: Publication source [RPS] 
- issue 125: dcat:byteSize - check constraints 
- issue 90: Version definition [RVSDF]

There are also a certain number of issues which are referenced in the current DCAT document but are not in the backlog nor DCAT CR. 
- issue 65: Entailment of schema.org [RES]
- issue 110 : Review global domain axioms on dcat properties 
- issue 144 : DCAT 2014 profile of DCAT
- issue 145:  Distinguish query and item access semantics [RDISA]
- issue 149 : DCAT - DATS alignment (profile)
- issue 150: DCAT - H CLS alignment
- issue 151: DCAT - ISO 19115 alignment (profile)
- issue 152: DCAT - Datacite alignment (profile)
- issue 164: DCAT - DDI alignment (profile)
What do we what to do with these? Shall we move these in the backlog? 

[...]

PR #899, opened 2019-04-16
issue #838 - commenting the remained backlog issues 

riccardoAlbertoni:
This PR cleans all the remained backlog issues (i.e. 82, 83, 251).

After accepting this PR, I think we can close the issue #838 , as issues 85 and 254  are the only issues still mentioned in the DCAT CR  and they are tracked by distinct issues in DCAT CR milestone.


davebrowning (approved):

dr-shorthair (approved):
We now have a lot of commented-out material in the draft. There should be a backlog issue relating to most (all?) of these. Perhaps we should copy the commented material into the issue threads and de-clutter the main artefact?

PR #903, opened 2019-04-22
Added definitions and scope-notes for dct: prov: foaf: and odrl: props

dr-shorthair:
Content copied over from the rec document

As discussed in #725


dr-shorthair:
1. In cases where the DCAT CR has a 'definition' which is different to the external original, I have copied them into a `skos:definition` element ; 

2. In a few cases, the Usage Note is different in the context of different classes - typically replacing 'resource' with 'distribution' or similar. I have simply copied each usage note into a separate `skos:scopeNote`.

davebrowning (commented):
Actual change looks okay to me, but there is a conflict that needs addressing in the list of namespaces.  (Would have just fixed that but realised I wasn't quite sure how it had come about - think the main branch has had some added  (e.g. oa: earl:) but when I look in the conflict resolution screen it looks....strange....

riccardoAlbertoni (approved):

riccardoAlbertoni:
@davebrowning :
> Actual change looks okay to me, but there is a conflict that needs addressing in the list of namespaces. (Would have just fixed that but realised I wasn't quite sure how it had come about - think the main branch has had some added (e.g. oa: earl:) but when I look in the conflict resolution screen it looks....strange....

 I have solved the conflict, my opinion is that we can drop the "\<code\>\</code\>"  tags and keep the namespace definitions we currently have in the gh-pages.  I remember myself adding some of the additional namespaces as they were used in the document but not declared.

davebrowning (approved):

davebrowning:
Thanks @riccardoAlbertoni - that looks better now. [Not sure what github was showing me before but I'll put it down to my confusion...]

Issue #119, opened 2018-02-15
dcat:themeTaxonomy - check constraints


dr-shorthair:
In DCAT v1 the property `dcat:themeTaxonomy` is axiomatized 
```
dcat:themeTaxonomy
  rdf:type owl:ObjectProperty ;
  rdfs:domain dcat:Catalog ;
  rdfs:range skos:ConceptScheme ;
.
```

1. Verify that the range is appropriate and necessary
2. Verify that the domain is appropriate and necessary (see #110)
3. Consider whether any guarded constraints (using owl:Restriction) should be introduced (see #105)

dr-shorthair:
This looks like a very specific DCAT capability that is appropriately modeled

dr-shorthair:
No-one advocating for changes here. 

I'll bring this proposition to the next DCAT team telecon for resolution.

dr-shorthair:
+1 to no change

dr-shorthair:
Goal is to support dataset catalogs, not provide general purpose predicates for the semantic web

https://www.w3.org/2018/03/21-dxwgdcat-minutes
resolved: no change to domain/range of dcat:themeTaxonomy

akuckartz:
I think that this resolution is inconsistent with the others where the domain was removed. But that is just my view.

akuckartz:
Accessing https://www.w3.org/2018/03/21-dxwgdcat-minutes results in a 403 error.

dr-shorthair:
Yes - there was a bug generating the minutes. I clipped a copy of the IRC log and have forwarded it to W3C to process.

dr-shorthair:
On consistency @akuckartz the argument is that `dcat:themeTaxonomy` is a specific predicate to link a `skos:ConceptScheme` to a `dcat:Catalog` whose members may use elements from the concept scheme. It is not a goal of DCAT to be a source of general purpose predicates for the semantic web, it is to support catalogs of datasets and related resources. So it is reasonable for a catalog to indicate which classification systems are in scope for its members.

arminhaller:
As mentioned in the telco, I think the naming of the property is inconsistent with the purpose that it should be used to indicate the classification scheme for a Catalog. If that is the desired semantics, then catalogTheme or themeCatalog would certainly make sense. It if is generically called themeTaxonomy, I would not expect a domain. I was voting 0, though, no objection, but I agree with @akuckartz that it is inconsistent.

pwin:
[...]

PR #911, opened 2019-04-26
Relax range of dcat:themeTaxonomy 

dr-shorthair:
add a scope note recommending use of a skos:ConceptScheme

Response to #119


dr-shorthair:
Note that existing records in which the `dcat:themeTaxonomy` property points to a `skos:ConceptScheme` will still be valid, so backward compatibility is maintained at that level.

davebrowning (approved):

davebrowning:
@riccardoAlbertoni, @andrea-perego, @agbeltran  - any dissenting views on this (as discussed in #119)

riccardoAlbertoni (approved):

PR #913, opened 2019-04-26
Pwin edit dcat 20190425

pwin:
editing and re-structuring; addition of string content to <a> tags where missing; removal of commented-out HTML

Still more to be done, but I don't want it to get too far behind other changes

awaiting outcome of #911 before changing text

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -56,9 +56,9 @@ $(document).ready( function() {
         <li>addition of a <a href="#Class:Resource"><code>dcat:Resource</code></a> class for representing any asset than can be included in the catalog.  This is
             now the super-class of <a href="#Class:Dataset"><code>dcat:Dataset</code></a></li>
         <li>addition of <a href="#Class:Data_Service"><code>dcat:DataService</code></a>, as a sub-class of <a href="#Class:Resource"><code>dcat:Resource</code></a>, so that catalogs can now hold information about service end-points that provide access to data assets</li>
-        <li>addition of ways to represent <a href="#bag-of-files">loosely structured catalogs</a>, where there is no distinction between a dataset and its distributions, such as in many open data repositories</li>
-        <li>more details for the ways of representing <a href="#examples-dataset-provenance">dataset provenance</a> and <a href="#quality-information">quality</a></li>
-        <li>an <a href="#dcat-sdo">alignment</a> between the DCAT vocabulary and the schema.org vocabulary</li>
+        <li>addition of ways to represent <a href="#bag-of-files"></a>, where there is no distinction between a dataset and its distributions, such as in many open data repositories</li>
+        <li>more details for the ways of representing <a href="#examples-dataset-provenance"></a> and <a href="#quality-information"></a></li>
+        <li>an <a href="#dcat-sdo"></a> between the DCAT vocabulary and the schema.org vocabulary</li>
     </ul>
 
 
@@ -70,7 +70,7 @@ $(document).ready( function() {
     </p>
     <h3 id="dcat_history">DCAT history</h3>
     <p>The original DCAT vocabulary was developed and <a href="http://vocab.deri.ie/dcat">hosted</a> at the Digital Enterprise Research Institute (DERI), then refined by the <a href="http://www.w3.org/egov/">eGov Interest Group</a>, and finally standardized in 2014 [[?VOCAB-DCAT-20140116]] by the <a href="http://www.w3.org/2011/gld/">Government Linked Data (GLD)</a> Working Group.</p>
-    <p>This revised version of DCAT, Version 2, was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[?DCAT-UCR]] gathered from peoples' experiences with the DCAT vocabulary from the time of the original version, and new applications that were not considered in the first version. A summary of the changes from [[?VOCAB-DCAT-20140116]] is provided in the <a href="#changes">Change History</a> section. </p>
+    <p>This revised version of DCAT, Version 2, was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[?DCAT-UCR]] gathered from peoples' experiences with the DCAT vocabulary from the time of the original version, and new applications that were not considered in the first version. A summary of the changes from [[?VOCAB-DCAT-20140116]] is provided in the <a href="#changes"></a> section. </p>
 
     <h3 id="external_terms">External terms</h3>
     <p>DCAT incorporates terms from pre-existing vocabularies where stable terms with appropriate meanings could be found, such as <a href="http://xmlns.com/foaf/0.1/homepage"><code>foaf:homepage</code></a> and <a href="http://purl.org/dc/terms/title"><code>dct:title</code></a>.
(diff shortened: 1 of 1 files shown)


dr-shorthair:
@pwin - I think you may have done a bunch of unnecessary work (and undone some earlier work by @andrea-perego ). 
Where a link's `href` points to a section anchor, the section number and title gets inserted automatically as the link text.

pwin:
thanks for the heads-up @dr-shorthair 
I have reverted all cases where the hyperlink points to an internal anchor

dcat/index.html
```diff
@@ -225,31 +224,23 @@ <h3>DCAT scope</h3>
...
-          Member items in a catalog should be members of one of the sub-classes, or of a sub-class of these, or of a sub-class of <a href="#Class:Resource"><code>dcat:Resource</code></a> defined in a DCAT application profile or other DCAT application.
+          Member items in a catalog should be members of one of these sub-classes, or of a sub-class of these, or of a sub-class of <a href="#Class:Resource"><code>dcat:Resource</code></a> defined in a DCAT application profile or other DCAT application.
           <a href="#Class:Resource"><code>dcat:Resource</code></a> is effectively an extension point for defining a catalog of any kind of resource.
         </li>
         <li>
           <a href="#Class:Dataset"><code>dcat:Dataset</code></a> represents a dataset in a catalog.
           A dataset is a collection of data, published or curated by a single agent.
-          Data comes in many forms including numbers, words, pixels, imagery, sound and other multi-media, and potentially other types, any of which might be collected into a dataset.
```
dr-shorthair, line 234: Why remove this?

dr-shorthair (commented):
See one small question re. deletion at line 234

dcat/index.html
```diff
@@ -45,32 +45,32 @@
...
+        <li>addition of ways to represent <a href="#bag-of-files"></a>, where there is no distinction between a dataset and its distributions, such as in many open data repositories</li>
+        <li>more details for the ways of representing <a href="#examples-dataset-provenance"></a> and <a href="#quality-information"></a></li>
+        <li>an <a href="#dcat-sdo"></a> between the DCAT vocabulary and the schema.org vocabulary</li>
     </ul>
 
 
     <p>
-        The detailed differences between the two documents can be seen <a href="https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2018%2FWD-vocab-dcat-2-20180508%2F&doc2=https%3A%2F%2Fwww.w3.org%2FTR%2F2018%2FWD-vocab-dcat-2-20181011%2F">here</a>.
+        The detailed differences between Version 2 and Version 1 can be seen <a href="https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fwww.w3.org%2FTR%2F2018%2FWD-vocab-dcat-2-20180508%2F&doc2=https%3A%2F%2Fwww.w3.org%2FTR%2F2018%2FWD-vocab-dcat-2-20181011%2F">here</a>.
```
riccardoAlbertoni, line 66: If we want to show the difference between dcat 2 and 1 then the link should be updated, we can't put the very last dcat2 dated URI as we haven't yet voted for the CR, but in the meanwhile,  we can use  "https://w3c.github.io/dxwg/dcat/" which is a good approximation, what do you think?
```suggestion
        The detailed differences between Version 2 and Version 1 can be seen <a href="https://services.w3.org/htmldiff?doc1=https://www.w3.org/TR/vocab-dcat/&doc2=https://w3c.github.io/dxwg/dcat/">here</a>.
```

riccardoAlbertoni (approved):

dcat/index.html
pwin: @dr-shorthair I re-worked the preceding section about scope to include this point (see line 216) so that the definition of dcat:Dataset is cleaner

dcat/index.html
andrea-perego: Or maybe we can revise the text as follows:

> The detailed differences between [[VOCAB-DCAT-20140116]] and and this version of DCAT

dcat/index.html
```diff
@@ -45,32 +45,32 @@
...
-    <p>This revised version of DCAT was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[?DCAT-UCR]] gathered from peoples' experience with the DCAT vocabulary from the time of the original version, and new applications that were not considered in the first version. A summary of the changes from [[?VOCAB-DCAT-20140116]] is provided in <a href="#changes"></a>. </p>
+    <p>This revised version of DCAT, Version 2, was developed by the <a href="https://www.w3.org/2017/dxwg/">Dataset Exchange Working Group</a> in response to a new set of Use Cases and Requirements [[?DCAT-UCR]] gathered from peoples' experiences with the DCAT vocabulary from the time of the original version, and new applications that were not considered in the first version. A summary of the changes from [[?VOCAB-DCAT-20140116]] is provided in the <a href="#changes"></a> section. </p>
```
andrea-perego, line 73: @pwin , I think the use of "Version 2" as in this line and others, is a bit redundant and maybe confusing. I would rather drop it, and just use "this version", or rather replace it with an abbreviation (e.g., "DCAT 2"), and include it in parenthesis. E.g., instead of:

> This revised version of DCAT, Version 2,

we could simply say

> This revised version of DCAT

or

> This revised version of DCAT (DCAT 2)

andrea-perego (changes requested):
@pwin , about the use of "[in] the <a href="#section-id"></a> [section]", I think "the" and "section" are redundant and do not help reading. ReSpec already adds the § ("section") symbol, so no need to add "section". Moreover, ReSpec adds not only the section number, but also the section title (which sometimes is very long), so adding "section" at the end makes the text less fluent.

I would propose to roll back these changes.

andrea-perego (changes requested):
Another thing I would propose to fix:

The revisions use "Version 1" to refer to DCAT 2014. I would personally avoid this approach, and instead use the relevant bib ref ([[VOCAB-DCAT-20140116]] or simpy [[VOCAB-DCAT]]), which provides an unambiguous indication of the relevant DCAT version.

andrea-perego (commented):
@pwin , I saw you replaced cross-refs to examples with "as follows", "following". 

I think that it is preferable to provide an unambiguous link to an example, also when it follows the text. Anyway, if the other reviewers are happy with your approach, I won't object.

andrea-perego (changes requested):
I saw that some NOTEs stating that a property / class has been added etc. have been included at the end of the relevant section.

In a previous call, it was decided to move them just after the section title, and @riccardoAlbertoni started revising the spec accordingly, and so I've done in the subsequent PRs.

I suggest applying the same principle here.

andrea-perego:
> @pwin , I saw you replaced cross-refs to examples with "as follows", "following".
> 
> I think that it is preferable to provide an unambiguous link to an example, also when it follows the text. Anyway, if the other reviewers are happy with your approach, I won't object.

Sorry, @pwin , I now realise that you may not have done these changes, as the PR where I added the cross-refs to examples was created after yours. My apologies.

andrea-perego:
> I saw that some NOTEs stating that a property / class has been added etc. have been included at the end of the relevant section.
> 
> In a previous call, it was decided to move them just after the section title, and @riccardoAlbertoni started revising the spec accordingly, and so I've done in the subsequent PRs.
> 
> I suggest applying the same principle here.

Sorry, again. This issue is also probably due to PRs created after yours.

dcat/index.html
pwin: the last option is the least ambiguous.  happy with that

dr-shorthair:
This branch has not been synced with the master for a long time - is it still alive?

pwin:
It needs revisiting

andrea-perego:
Moved this PR to the DCAT Ratification milestone, as it is editorial, and therefore can be sorted out after DCAT CR transition.

andrea-perego:
I suggest we close this PR, as the DCAT spec has changed substantively since then, so it is likely to need a newly created PR (time permitting).

pwin:
Agreed

andrea-perego:
Thanks, @pwin .

Issue #920, opened 2019-05-08
dcat.ttl - editorialNotes


davebrowning:
editorialNotes in dcat.ttl need reviewing/revising (as does the `owl:versionInfo`) to reflect the current state of the document.

davebrowning:
Best to do this across the file in English and then do the translations?

riccardoAlbertoni:
@davebrowning wrote:
>editorialNotes in dcat.ttl need reviewing/revising (as does the owl:versionInfo) to reflect the current state of the document.

The owl:versionInfo review you are suggesting reminds me that we might want to doublecheck dcat.ttl by using the [OntOlogy Pitfall Scanner (OPSS)](http://oops.linkeddata.es/response-advanced.jsp#). 

There are different issues raised when applying OPSS on the current dcat.ttl, some perhaps are not so important but others might provide hints for improvements.
For example,  
- Results for P08: Missing annotations. 28 cases suggest the following elements have no rdfs:label defined: 
› http://www.w3.org/ns/dcat#startDate 
› http://www.w3.org/ns/dcat#endDate 
› http://www.w3.org/ns/dcat#centroid 
› http://www.w3.org/ns/dcat#bbox 

- we might want to doublecheck why results for P12: Equivalent properties not explicitly declared raise that the following properties could be defined as equivalent: 
› http://www.w3.org/ns/prov#hadRole, http://www.w3.org/ns/dcat#hadRole  

- Results for P41: No license declared.

riccardoAlbertoni:
I am preparing a distinct PR  to fix the Missing annotation ( rdfs:label) for  ›http://www.w3.org/ns/dcat#startDate
› http://www.w3.org/ns/dcat#endDate
› http://www.w3.org/ns/dcat#centroid
› http://www.w3.org/ns/dcat#bbox

davebrowning:
Work moved from #921 to #926

PR #924, opened 2019-05-09
Trying to address some of the issues highlighted by OPSS 

riccardoAlbertoni:
Related to my comment on issue #920 

- I've added missing rdfs:label for startDate, endDate, centroid, bbox
- I've added  https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document as dct:license for the owl:Ontology. Reading the https://www.w3.org/Consortium/Legal/2015/doc-license,   copyright-software-and-document seemed the most suitable W3C license for code such as the DCAT TTL representation.  Do other editors agree on the choice?

Still not a hint  on why OPSS raise that the following properties could be defined as equivalent:
› http://www.w3.org/ns/prov#hadRole, http://www.w3.org/ns/dcat#hadRole
Any Idea?


dr-shorthair:
http://www.w3.org/ns/prov#hadRole, http://www.w3.org/ns/dcat#hadRole are not equivalent - they have different `rdfs:domain`. This is explained in the document. 

It is because PROV is fully activity-centric, so all relationships are via an activity.

riccardoAlbertoni:
@dr-shorthair 
> http://www.w3.org/ns/prov#hadRole, http://www.w3.org/ns/dcat#hadRole are not equivalent - they have different `rdfs:domain`. This is explained in the document.
> 
> It is because PROV is fully activity-centric, so all relationships are via an activity.

Yes,  I was aware that dcat:hadRole is meant to be distinct from prov:hadRole,  and that is why I was trying to figure out why  OPSS is suggesting they should be the same.

I worried that there were some flaws in the axiomatization of dcat:hadRole which might induce OPSS in suggesting the equivalence.
I have noticed that If I change the name of dcat:hadRole the equivalence suggestion from OPSS disappears. I guess OPSS suggests the equivalence relying on the similarity between the name of properties.  So it is not a matter of axiomatization flaws and we can simply ignore the suggestion from OPSS.

dr-shorthair:
> I guess OPSS suggests the equivalence relying on the similarity between the name of properties. So it is not a matter of axiomatization flaws and we can simply ignore the suggestion from OPSS.

OK - I understand. The localname is deliberately the same (as we are effectively correcting an unnecessary constraint in PROV) so I thinnk we should proceed.

dr-shorthair (approved):

davebrowning (approved):

PR #925, opened 2019-05-12
Verify all recommended properties are in diagram

dr-shorthair:
and move diagrams to /images/ directory


agbeltran:
In Fig 2 (https://rawgit.com/w3c/dxwg/dcat-diagrams/dcat/index.html#UML-DCAT-Qualified-Relations) only one of the properties is indicated with cardinality - I would suggest either to add all cardinalities or to remove that one?

dr-shorthair:
Good catch @agbeltran - I've removed the stray cardinality indicator. 
Note that while we are showing PROV-O properties I also included `prov:wasGeneratedBy`. 
So I wonder if there shoudl be a reference to this figure from https://rawgit.com/w3c/dxwg/dcat-diagrams/dcat/index.html#examples-dataset-provenance or https://rawgit.com/w3c/dxwg/dcat-diagrams/dcat/index.html#Property:dataset_wasgeneratedby as well?

agbeltran:
Yes, I think referencing the figure from both sections (6.6.7 introducing the property 'was generated by' and D.2 with provenance examples) would be useful for people to see the broader context in which the property can be applied.

riccardoAlbertoni (approved):
Crosschecking the properties in figure 2 with those mentioned in the document, I have found some minor inconsistencies, we might want to address. 

- the properties dct:license and dct:right are duplicated in the class dcat:Catalog, they are both inherited by dcat:Resource and listed among those defined in the class, I guess we can drop one of the two occurrences.
- In Class dcat:CatalogRecord, the property dct:contributor is mentioned in the figure but it is not listed in the document. We might want to add it in the document or alternatively delete it from the figure ... 
- the properties dct:license and dct:right are duplicated in the class dcat:DataService, they are  both inherited by dcat:Resource and listed among those defined in the class, I guess we can drop one of the two occurrences.

The classes dct:PeriodOfTime and dct:Location do not appear in any of the figures, but this sounds very reasonable to me.

dr-shorthair:
Thanks @riccardoAlbertoni - I think it is clean now.

agbeltran (approved):

PR #919, opened 2019-05-07
DCAT: Editorial fixes

andrea-perego:
Preview: https://raw.githack.com/w3c/dxwg/andrea-perego-dcat-ed-rev/dcat/index.html

HTML diff:

https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2Findex.html&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fandrea-perego-dcat-ed-rev%2Fdcat%2Findex.html

Summary of changes:

- https://github.com/w3c/dxwg/pull/919/commits/973de8c76575d47303dafab98f9c460a30c0f476 :
  - Removed usage note from `dcat:startDate` / `dcat:endDate`, and updated their range, consistently with other properties using a datetime format
  - Minor editorial fixes
- https://github.com/w3c/dxwg/pull/919/commits/ed7862da4ebb2c1fa3c8ddee735840c358ed2143 :
  - Moved section and references to `dcat:bbox` just after `locn:geometry`
  - Extended note in [Section 11.2](https://raw.githack.com/w3c/dxwg/andrea-perego-dcat-ed-rev/dcat/index.html#spatial-properties) with additional pointers to SDWBP
- https://github.com/w3c/dxwg/pull/919/commits/9510be815be30f173f56019be7ff4c358f08abc1 : Added `class="nohighlight turtle"` to all PRE elements
- https://github.com/w3c/dxwg/pull/919/commits/ff50631e9b2579d0de2c1a2f298096f73635b58e : Fixed typo reported in https://github.com/w3c/dxwg/issues/922 :
  > Minor type in the usage note for spatial resolution: "For other kinds of spatial dataset," should probably be "For other kinds of spatial datasets,"
- https://github.com/w3c/dxwg/pull/919/commits/7a7239cfdd6fe6df63cde9728b1768f100c8531e : Addressed issue reported in https://github.com/w3c/dxwg/issues/922 :
  > dcat:compressFormat and dcat:packageFormat: There is no note that those were added during the revision.
- https://github.com/w3c/dxwg/pull/919/commits/ee27db4f68dc8fc94a36a7536d36cba442fec87c : Fixed typo Reported in #922:
  > dcat:Role: The note says that the property was newly added, but dcat:Role is a class...
[...]

PR #926, opened 2019-05-12
Update to editorial notes and change notes in ttl file

davebrowning:
in line with #919... and #920. 

Remove editorialNotes marking work in progress
Align change notes with the work in #919

### DIFF
--- a/dcat/rdf/dcat.ttl
+++ b/dcat/rdf/dcat.ttl
@@ -1087,7 +1087,7 @@ dct:conformsTo
 dct:creator
   skos:scopeNote "Resources of type foaf:Agent are recommended as values for this property."@en ;
   skos:definition "The entity responsible for producing the resource."@en ;
-  skos:changeNote "Property added in this context in this revision of DCAT"@en ;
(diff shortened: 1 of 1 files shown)


davebrowning:
This should only get merged after #919 has been merged.

Question: should the ttl file have an entry for `dct:location` since we have one for `dct:PeriodOfTime`?

dr-shorthair:
My view is that we should mention _every_ class and property that is 'recommended for use'. 
WHere these have usage notes or definitions in context that  vary at all from the originator's, these should also be included as `skos:definition` and `skos:scopeNote`.

agbeltran:
I agree with @dr-shorthair's point.

davebrowning:
Thanks - I've done that now (though since I don't have any support for ttl syntax someone should have a check....)

I also think that this lines up with #919 (as it did before) but re-checking all the notes etc can perhaps wait till its time to merge this.....

dr-shorthair (approved):
I did a syntax check.

pwin (approved):
look ok

PR #929, opened 2019-05-15
Update to editors

davebrowning:
Initially adding w3c ids since documentation suggests echidna (or something) needs it..


davebrowning:
This would be a good chance to make sure we're happy with affiliations as per  https://w3c.github.io/manual-of-style/#affs  (I'm checking what I want to do....)  Respec doesn't support it well, though it can be done (kind of)

andrea-perego:
My apologies, @davebrowning , I have also added them in #919, which has now been merged.

Checking the diff, they seem to be the same.

davebrowning:
That's okay - probably best to just drop this one, then!

Issue #967, opened 2019-06-30
dcat conformance question


smrgeoinfo:
Section 4 on conformance states:
"Access to data is organized into datasets, distributions, and data-services.
An RDF description of the catalog itself and its datasets, distributions, and data-services is available"

then in section 5:
"Member items in a catalog should be members of one of the sub-classes, or of a sub-class of these, or of a sub-class of dcat:Resource defined in a DCAT profile or other DCAT application. dcat:Resource is effectively an extension point for defining a catalog of any kind of resource."

Question-- If I extend dcat:Resource to describe e.g. software, a vocabulary, an API or other resoruce but defining other subclasses of dcat:Resource,  will my catalog be a DCAT catalog?

dr-shorthair:
Good question Steve. I think per section 4 the answer is strictly 'no' but the wording in section 5 does leave that door open, so section 5 needs to be tightened a bit, else section 4 tweaked to allow for this. 

The question is whether a catalog that is designed consistent with DCAT design, but whose contents describe things other than datasets or dataservices, should be considered a conformant DCAT catalog. I could swing either way, so would be interested in other views.

riccardoAlbertoni:
@dr-shorthair wrote:
> The question is whether a catalog that is designed consistent with DCAT design, but whose contents describe things other than datasets or dataservices, should be considered a conformant DCAT catalog

I tend to say 'yes'. Otherwise, we risk discouraging the use of DCAT in unforeseen scenarios.

Let us be rather inclusive and let the users decide by themselves whether when and how DCAT can serve or not their non-data catalogs.    

A catalog of non-data resources would interoperate only on the notion of "catalog" and "resource" but I can't exclude this is useful for someone. Why should we exclude this? Is there any concrete risk I am not seeing?

In my opinion, the overall message of DCAT is clear: "Use it for Data catalog,"  and edge cases like the one discussed by Richard does not put it at risk.

pwin:
I think we should be comfortable with much wider usage than just cataloging 'data' content to any other type of asset.  The key use of DCAT imo is the publication on the web of a catalogue of resources .  most anything that we catalogue will be some sort of information

smrgeoinfo:
[...]

PR #977, opened 2019-07-04
adding cataloged resources explicitly ( for issue #967)

riccardoAlbertoni:
This PR addresses the issue #967 by explicitly mentioning the cataloged resources in the conformance section.

In this way, we convey the idea that a catalog of resources other than datasets and services is conformant.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -178,7 +178,7 @@
     <p>A data catalog conforms to DCAT if:</p>
     <ul>
         <li> Access to data is organized into datasets, distributions, and data-services. </li>
-        <li> An RDF description of the catalog itself and its datasets, distributions, data-services, and other cataloged resources (if any) is available (but the choice of
+        <li> An RDF description of the catalog itself, the corresponding cataloged resources, and distributions is available (but the choice of
             RDF syntax, access protocol, and access policy are not mandated by this specification).</li>
         <li> The contents of all metadata fields that are held in the catalog and that contain data about the catalog itself and its datasets, distributions, data-services, and other cataloged resources (if any) are included in this RDF description and are expressed using the appropriate classes and properties from DCAT, except where no such class or property exists.</li>
         <li> All classes and properties defined in DCAT are used in a way consistent with the semantics declared in this specification.</li>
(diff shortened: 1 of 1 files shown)


andrea-perego (commented):
I wonder whether this sentence

> the catalog itself and its datasets, distributions, data-services, and other cataloged resources (if any)

could be read as "you can have other resource types, but you need to have datasets, etc. to be conformant".

Maybe we can re-phrased it as

> the catalog itself and the corresponding cataloged resources

dcat/index.html
```diff
@@ -178,9 +178,9 @@ <h2 >Namespaces</h2>
     <p>A data catalog conforms to DCAT if:</p>
     <ul>
         <li> Access to data is organized into datasets, distributions, and data-services. </li>
-        <li> An RDF description of the catalog itself and its datasets, distributions, and data-services is available (but the choice of
+        <li> An RDF description of the catalog itself and its datasets, distributions, data-services, and other cataloged resources (if any) is available (but the choice of
```
```suggestion
        <li> An RDF description of the catalog itself, the corresponding cataloged resources, and distributions is available (but the choice of
```

dcat/index.html
```diff
@@ -178,9 +178,9 @@ <h2 >Namespaces</h2>
...
             RDF syntax, access protocol, and access policy are not mandated by this specification).</li>
-        <li> The contents of all metadata fields that are held in the catalog and that contain data about the catalog itself and its datasets, distributions, and data-services, are included in this RDF description and are expressed using the appropriate classes and properties from DCAT, except where no such class or property exists.</li>
+        <li> The contents of all metadata fields that are held in the catalog and that contain data about the catalog itself and its datasets, distributions, data-services, and other cataloged resources (if any) are included in this RDF description and are expressed using the appropriate classes and properties from DCAT, except where no such class or property exists.</li>
```
```suggestion
        <li> The contents of all metadata fields that are held in the catalog and that contain data about the catalog itself, the corresponding cataloged resources,  and distributions are included in this RDF description and are expressed using the appropriate classes and properties from DCAT, except where no such class or property exists.</li>
```

dr-shorthair (approved):

davebrowning (approved):

PR #993, opened 2019-07-13
Update index.html

### LINKED ISSUE #500: Consider requirements for specifying the order in which content negotiation is performed
It would appear that it will necessary to negotiate for profile before language or content-type.  Memento specifies that time-based negotiation MUST take place before any other negotiation.

larsgsvensson:
Added text to clarify order of content negotiation
Resolves #500 
Refers to ACTION-349


nicholascar (approved):
Great addition Lars!

rob-metalinkage (approved):
Noting that the issue of a redirect response still makes this problematic ;-(

PR #1014, opened 2019-07-25
Conneg Definitions

nicholascar:
As requested by the Plenary session, here are the new definitions added to the Conneg doc. I have worked up some of the text around the definitions and also added some usage notes, as indicated to Plenary, to ensure compatibility with terminology.

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -89,34 +89,32 @@
   <section id="definitions">
     <h2>Definitions</h2>
     <dl>
-      <dt><dfn>specification</dfn></dt>
+      <dt><dfn data-lt="specification">data specification</dfn></dt>
       <dd>
         <p>
-          An act of identifying something precisely or of stating a precise requirement. - <a href="https://en.oxforddictionaries.com/definition/specification">Oxford English Dictionary</a>
+          A specification, with human- and/or machine-processable representations, that defines the content and structure of data used in a given context.
         </p>
         <p>
-          One form of a <em>specification</em> is a <em>standard</em> which is a "basis for comparison; a reference
-          point against which other things can be evaluated" - [[DCTERMS]]
+          This document refers to both "data specifications" and also the more general "specifications". When the latter general case is referenced, its definition is taken
+          to be that of the DCMI Metadata Terms [[DCTERMS]]'s definition for a <em>Standard</em>, which is "A basis for comparison; a reference point against which other things
+          can be evaluated."
         </p>
+        <p><em>Source: deliberations of the DXWG.</em></p>
       </dd>
-      <dt><dfn>profile</dfn></dt>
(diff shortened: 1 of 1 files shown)


larsgsvensson (dismissed):
Minor nitpick: In line 111 it should probably read 'In this document, these, and "data profiles", are referred to as just "profiles".' (there's one "just" too many...)

kcoyle (changes requested):
It appears that line 99

&lt;dt>&lt;dfn data-lt="specification">data specification&lt;/dfn>

contradicts lines 102-104

     This document refers to both "data specifications" and also the more general "specifications". When the latter general case is referenced, its definition is taken
          to be that of the DCMI Metadata Terms [[DCTERMS]]'s definition for a <em>Standard</em>, which is "A basis for comparison; a reference point against which other things
          can be evaluated."

because the dfn data-lt points to specification, not data specification, yet "specification" has a different meaning from "data specification". I believe that the data-lt line should read:

&lt;dt>&lt;dfn data-lt="data specification">data specification&lt;/dfn>

rob-metalinkage:
it doesnt actually contradict - as a data specification is a specialised form of specification fully consistent with the more general dct:Standard definition. Either way would be actually correct, but it doesnt hurt to use the more specific term "data specification" when we are talking about the data profiles being negotiated, not the conneg-by-ap specification itself, or its implementable functional profiles.

nicholascar:
@kcoyle are you ok with me leaving it as-is following @rob-metalinkage ‘s suggestion?

Or would you prefer me to add in a new definition for “specification”: <dt><dfn data-lt="specification">specification</dfn> as per the current DC-inspired text, as well as the existing <dfn data-lt="data specification">data specification</dfn>? If I were to just change it as you suggest and not add in anything for just “specification”, then any references to the general specification will be left hanging (they will have a definition in text but not in a linkable definition list form).

kcoyle:
The document as it is states that there is a distinction between "specification" and "data specification". This definition is only for "data specification" and "specification" is given a different definition in lines 102-104. Yet this definition will be linked from any uses of "specification". That seems to be inaccurate. Either "specification" is given the DC Terms definition and is separate from data definition, OR it is a substitute for "data specification" within this document. I don't see how it can be both. 

You could either do:

&lt;dt>&lt;dfn data-lt="specification">data specification&lt;/dfn>

on line 99, or you could change 102-104 to say something like:

"Within this document we use "specification" and "data specification" interchangeably, with the same meaning as given in the definition for "data specification".

With that latter you could leave the dfn statement as it is now.

Do you see the difference?

nicholascar:
I have added the DCT definition of "specification" to the list of definitions and ensured all links to definitions are correct.

kcoyle (approved):

rob-metalinkage (approved):

PR #1015, opened 2019-07-25
Prof definitions

nicholascar:
Here are the Plenary-agreed definitions added in to PROF. I've had to re-work the Introduction extensively not to change content so much but just to avoid terminology conflicts ("data profile" v. "profile") and repetition. Plenary agreed to the not about "we refer to "data profile" as just "profile" in this document".


rob-metalinkage (changes requested):
I'm sorry but i dont agree that the profiles vocabulary is conflating the concept of a data profile with the more general case. There is nothing that says a profile cannot be specifying non-data aspects of some conformance target that is not an information resource, such as implementation choices for a service protocol - making some optional methods mandatory for example.

Subtle changes to wording are required - such as "A profile, also known as a <em>data profile</em>" => A profile, such as a <em>data profile</em> that constrains information content ..

this is actually more consistent with the first line "Profiles aim to increase interoperability within a community of users by introducing constraints, extensions or combinations on the use of more general specifications."

rob-metalinkage (approved):

larsgsvensson (approved):

Issue #418, opened 2018-09-25
Profile guidance definitions


agreiner:
The definitions in the profile guidance need some work.
 def of profile
"a particular function", together with the intro text's discussion of "any individual system", is too focused on a single application. If it were only to be used by a single system, there would be no need for profiles as any kind of shared spec.

def of specification
"an act of" is not the def we want. we use it to discuss a thing, not an act.

def of standard
again, wrong definition. What we are talking about is a published specification generally accepted as authoritative.

agreiner:
Also, the Related Definitions seems like it should be in references. The definition of profiles there is too easily mistaken for our own definition. I like the phrase "a particular application, function, community, or context." I think that is actually what we are talking about.

aisaac:
Thanks for the comments @agreiner . I'm not going to make a mention to this issue in the document though, because I think it's already captured in #371 which is already mentioned there.
This said I'll put a link to this issue in #371, making sure that that issue won't be closed before this one is closed.

nicholascar:
This issue was created before the main doc structure change and, as mentioned above, is covered by Issue https://github.com/w3c/dxwg/issues/371. I propose closing this one to just reduce the total number of open issues as that would be better than just not including it in the doc as it's eacier to include all open Issues tagged appropriately in the doc as this is then an additional logic rule (don't include specific Issue x) that isn't obviously apparent.

Closing could use a comment (from @aisaac or @agreiner) that indeed this is a duplicate of, or is subsumbed by, Issue https://github.com/w3c/dxwg/issues/371.

aisaac:
I sense we should keep this one open. I like closing issues, but I'm afraid if we put @agreiner's opening comment here as a comment in #371 it will be soon overflowed by discussions on the structure (which #371 is rather about for the moment)

heidivanparys:
> def of specification
> "an act of" is not the def we want. we use it to discuss a thing, not an act.

It seems that the wrong definition from [Oxford English Dictionary](https://en.oxforddictionaries.com/definition/specification) was chosen. The one referred to in the guidance refers to the process of specifying, the one actually meant is the one that should refer the the outcome of specifying.

[Oxford English Dictionary](https://en.oxforddictionaries.com/definition/specification) also has a definition for the outcome, _a detailed description of the design and materials used to make something_.

[...]

PR #1016, opened 2019-07-25
Guidance definitions

nicholascar:
Here are the Plenary-approved definitions which I've PRed into Conneg & PROF for the Guidance document. The only content change in this PR is the definitions addition but I've extensively reformatted the HTML of the document to regularise it to allow for easier future editing. I've separated the formatting & definitional addition commits.


rob-metalinkage (changes requested):
fine for guidance to be specifically about data profiles - but this is a special case of the more general sense of profiles (OGC service profiles are not data profiles - even they do also specify data profiles as part of a wider functional behaviour.) All is needs is a statement recognising this fact and scoping the document to data profiles.

nicholascar:
I've indicated that this doc is scoped to data profiles only.

rob-metalinkage (approved):

aisaac (commented):
Thanks for putting the new definition in, @nicholascar !

Some comments: 
- The definition is still the old one in the abstract. Maybe we could just remove the abstract?
- The sentence with "We have taken this variability into account [...]" cannot be really replaced "Within this document we concentrate on data profiles [...]". There is variability even for the notion of 'data profile', I would say, and we need to still acknowledge this. The second part of the sentence is not correct by the way. Maybe it's simpler to remove that sentence for the moment.
- I think it's still useful to keep the reference to https://www.w3.org/2017/dxwg/wiki/ProfileContext after the definitions (well, it is now, after I've updated the page ;-) )
- in "which can have the status of a Standard" the capitalization after this bit was on purpose (but well it's not crucial as long as the note below the paragraph is kept
- I'm ok removing the appendix with definitions at but I'd like to keep the reference to https://github.com/w3c/dxwg/issues/418 so that we don't forget to close (and maybe have one last check) later.

Isn't it possible to uncouple the coding changes? Accepting them would be useful!

pwin:
Is this so stale that it should be closed and if necessary brought forward as a new piece of work that is up to date with the current specs?

aisaac:
I think the new specs wouldn't necessarily bring many changes. But I agree it's completely stale :-(

aisaac:
This said it would be a pity to throw this commit away. Not sure how to handle it...

PR #1018, opened 2019-07-29
Conformance profiles

nicholascar:
As discussed, the Conformance Profiles section with no dependence on PROF


larsgsvensson (approved):

aisaac:
Catching up with emails and minutes I found this...
And I anticipate it will lead to problems and long discussions, while perhaps not limited.
First I am not sure this encoding of profiles really needs resources to be created in a namespace. Unless you want to create a full apparatus for representing content negotiation conformance? Next to PROF? If it's in PROF this is going to be very hard to follow (and for people in our WG to buy)
But well maybe this is not the space for discussion. So please take the lines above as a heads-up of comments to come...

For now (and more related to this PR) what I'd like to see is more wording like the one in the Note, which says this is about "functional" profiles and not "data" profiles. If this is about profiles for content negotiation, couldn't we directly call them "negotiation profiles"? Or "realization profiles", as these things are called "realizations" at https://raw.githack.com/w3c/dxwg/conformance-profiles/conneg-by-ap/index.html#realizations and elsewhere?
(and yes, I'd be even ready to suggest that RRD and the abstract model are also a kind of "realization" to make this happen, and also to have you use the infamous recursive trick again ;-) )

Similarly we could also try to avoid term collisions by using "compliance" instead of "conformance", with the understanding that "compliance" would be used to express how a system (not data) adheres to some specification. Or perhaps even better, as this part is about systems implementing realizations of the models, the term 'implementation' could be used. I know this is in a section about conformance (to CONNEG) but I don't see much harm in claiming that conformance to CONNEG can be obtained by having systems "implement" (or comply with) one of the realizations of the abstract model (or the abstract model itself).

All this may be seen terminological nitpicking, but I think that again it could save us precious discussion time when the group will comment on the spec, and in fact it does not twist your intention in terms of the specification.

rob-metalinkage (approved):

PR #1019, opened 2019-07-29
example as per Issue 287

nicholascar:
As requested in last teleconference. I've adjusted the text in the preceding paragraph to this example too and also added '<' & '>' to all `Content-Profile:` examples in the doc, as per the ID formatting.


rob-metalinkage (approved):
Looks good - will re-read in final form to check flow.

larsgsvensson (approved):

PR #1026, opened 2019-07-31
Conneg action 343

nicholascar:
Major additions to the QSA sections, some minor adjustments elsewhere.

See it rendered: https://raw.githack.com/w3c/dxwg/conneg-ACTION-343/conneg-by-ap/index.html


rob-metalinkage (changes requested):
couple of clean ups:

as long as reach => as long as each 

there are broken links that do not render - e.g.  " to be demonstrated as per #conformancetoprofiles. "
and
"detailed below in #xxxx." 

and i'm not sure that the QSA discovery is quite there yet - it doesnt allow resources to bootstrap within content - only servers to inject link headers - and thats the impossible step to retrofit even if we have a good redirection based conneg capability in place.

"or to be able to be documented in "  => " and may be described in"

larsgsvensson (changes requested):
Nice work and a substantial improvement!
Minor nits:
1. In examples 5 and 6, the first line should read `HEAD /resource/a HTTP/1.1` (There are no full URIs in the request line)
2. In example 6 the link header occurs twice; once with each link on one line and once more nicely formatted)
3. In the last line of example 6, it would  be better to include a specific reference to the link header: "[In the Link header, Content-Location URIs are only used as context URIs for resources conforming to profile <urn:example:profile:1>]
4. In §7.2.1.2 there you use "must" and "may" in a fashion that seems normative. Should that be "MUST" and MAY"? I also propose to remove the sentence "Other content negotiation dimensions ... requested language." since the specification of those are outside of our specification.
5. In §7.2.1.3 (Token use) we should add something saying that for a specific resource there MUST exist a 1:1 mapping between profile tokens and profile URIs.

nicholascar:
I think I've addressed all of Lars' comments and 1, 2 & 4, but not 3, of Rob's sentences so haven't addressed "QSA discovery is [not] quite there yet".

larsgsvensson (approved):
Yes, looks fine now, and I think you've also addressed my comment no 3!

rob-metalinkage (approved):
Have approved noting that the issue of "too concrete" examples of QSA key discovery still needs to be addressed IMHO.

nicholascar:
Yes, noted Rob. Thanks for approving - I need may of the changes here in to then do a range of next stings including considering removing the "too concrete bits".

larsgsvensson:
@nicholascar scripsit:
> I need may of the changes here in to then do a range of next stings including considering removing the "too concrete bits".

Do we need an issue for this so that it doesn't get lost?

PR #1030, opened 2019-08-06
Conneg profiles appendix

nicholascar:
A non-normative description of the specification and profiles of *Content Negotiation by Profile* using *Profile Vocabulary* terms.

Rendered at https://raw.githack.com/w3c/dxwg/conneg-profiles-appendix/conneg-by-ap/index.html


rob-metalinkage (changes requested):
The 5 profiles of it are then, as per regular PROF modelling, termed prof:Profiles. The 5 profiles are presented in a profile hierarchy
=>
The relationship between the abstract model and implementation realizations are described are presented in a profile hierarchy.

Also not so sure about a few things - 
1) why arent http and rrd  profiles of the abstract model? 
2) do we need to muddy the waters by having the abstract a profile of the document - there is no separate sense of conformance to the document independent of conformance to one or more profiles... whoudl rather skip this

larsgsvensson (changes requested):
I agree with Rob that the turtle code and the figure don't match (the code says
```
cnpr:rrd a prof:Profile ;
    prof:isProfileOf <https://www.w3.org/TR/dx-prof-conneg/> .
```
but in the figure `cnpr:rrd` is related to the abstract model and not to "content negotiation by profile".

Further, I miss the relation between §2.1 and Appendix A.1: Since the code in the appendix is a representation of the concepts in §2.1 we should make that explicit and also try to use the same language in the turtle code in the appendix as we do in §2.1 (including usage notes etc.)

And finally, I'm not sure what the status of "Example 12" is -- is it really an example or is it more of a listing? Since it's not normative, we cannot say it's a specification of those conformance classes, so I'm a bit at odds here...

nicholascar:
@rob-metalinkage:

1. *why arent http and rrd profiles of the abstract model?*
One was a typo, now fixed. The other bypassed since I've removed the Abstract Model profile
2. do we need to muddy the waters by having the abstract a profile of the document
*No, I've removed it*

@larsgsvensson:

* *in the figure cnpr:rrd is related to the abstract model and not to "content negotiation by profile"*
Fixed, as per the points made above for Rob
* *I miss the relation between §2.1 and Appendix A.1...*
I've made the RDF match the tabular data from §2.1. Good call by the way! The explanations there were better anyway.
* *I'm not sure what the status of "Example 12"...*
Bypassed by making it something different: a code listing.

larsgsvensson (approved):

rob-metalinkage (approved):

PR #1037, opened 2019-08-10
Conneg action 356

nicholascar:
All of Antoine's points 
+ 
renaming Realizations to Application Profiles 
+
completion of QSA keyword discovery

Note this last point references the formal data model introduced in the branch [conneg-altprofiles-model](https://github.com/w3c/dxwg/tree/conneg-altprofiles-model) which means the couple of hyperlinks to it won't resolve until that branch is merged.

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -457,8 +457,8 @@ Preference-Applied: profile="urn:example:schema"
       <p>
         All content negotiation takes place between a <a>client</a> and a <a>server</a> over the Internet with the
         former requesting a representation of a <a>resource</a> or a <em>resource</em>'s <em>metadata</em> through a
-        <a>request</a> and receiving it via a <a>response</a>. In some cases, a <em>server</em> may have to make a
-        <em>request</em> of a <em>client</em> and receive a <em>response</em>.
+        <a>request</a> and receiving it via a <a>response</a>. For any given pair of communicating machines, the
+        the roles of <em>client</em> and <em>server</em> may be reversed as information is requested back and forth.
       </p>
       <p>
         An Internet <em>resource</em> may have many aspects over which a <em>client</em>/<em>server</em> pair of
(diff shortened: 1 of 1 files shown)


nicholascar:
Please note that the definitions in the document in this PR as slightly disjoint from content but that is addressed in PR #[1039](https://github.com/w3c/dxwg/pull/1039).

rob-metalinkage (approved):
Slightly confused why the many same edits appear in two PRs - have approved both though.

larsgsvensson (changes requested):
Two issues:
1. In [§6.1](https://raw.githack.com/w3c/dxwg/conneg-ACTION-356/conneg-by-ap/index.html#abstractmodelcontext) it says: "In some cases, a server may have to make a request of a client and receive a response." I learned that in client/server interaction, the one that initiates the interaction is always the client, so can it be that a server makes a request of a client?
2. In [§7.2.1.3](https://raw.githack.com/w3c/dxwg/conneg-ACTION-356/conneg-by-ap/index.html#qsa-token-use) it says “there MUST exist a 1:1 mapping per resource and session between profile tokens and profile URIs that is discoverable by clients.” Does this mean that a token/URI mapping can be valid for just one session so that a client might need to revalidate its token mappings for each interaction? (And what exactly is a “session”?)

nicholascar:
@larsgsvensson addressing your points above. 

For 1., text changed to:

> For any given pair of communicating machines, the the roles of <em>client</em> and <em>server</em> may be reversed as information is requested back and forth.

For 2., text changed to:

> As per [§ 6.2 Profile Identification](https://raw.githack.com/w3c/dxwg/conneg-ACTION-356/conneg-by-ap/index.html#profileid), all profiles MUST be identified by URI since URIs are universally unique and therefore do not require registration to avoid identifier duplication. However, systems conforming to either of the QSA functional profiles of this specification, instead of using URIs to identify profiles during client/server interactions, MAY use un-registered and non-universally unique tokens to identify profiles but, if they do, they MUST provide a way for clients to discover profile token/URI mappings.

larsgsvensson (approved):

PR #1039, opened 2019-08-11
Conneg redefs

nicholascar:
Tidying up definitions & notes about profiles


rob-metalinkage (approved):
OK - I have added one more issue #1040 to address a concern raised during review - but that can be treated independently of these edits

larsgsvensson (approved):

Issue #455, opened 2018-10-08
Provide an example of both a profile and a resource adhereing to it


nicholascar:
Currently we haven't actually articulated how a resource claiming to conform to a profile indicates that cliamed conformance. Is it by simple use of `dct:conformsTo`? If so, what's the pattern of going from a resource to the Profile it conforms to then to a Resource Descriptor with some sort of Constraints Resource Role so it can be tested?

kcoyle:
By "resource" do you mean instance data?

nicholascar:
Yes, poorly worded of me since we have a *resource* property in use in the ontology too and it’s a general word.

So if I have a profile, described using the ont which may include Resource Descriptor instances for things like SHACL constraints, I’m now talking about an instance of the profile, like a metadata record claiming to adhere to the profile. 

I’ll make an instance of my demo profile here shortly for testing: https://github.com/CSIRO-enviro-informatics/csiro-epub-dcap

nicholascar:
There are now multiple instances of resources conforming to profiles in the document, e.g. each of the `Resource Descriptors` in [Example 1](https://w3c.github.io/dxwg/profilesont/#example-1-a-full-example-described-using-the-profiles-vocabulary-in-rdf-turtle) and [Example 8](https://w3c.github.io/dxwg/profilesont/#example-8-adms-profile-of-dcat-according-to-prof-in-rdf-turtle) however a specific, short, example showing instance data conforming to a profile seems appropriate/required to explain conformance to profiles in isolation.

nicholascar:
Addressed by PR #1046

PR #1046, opened 2019-08-29
conformstTo isolated example

nicholascar:
Addressing Issue #455 

https://raw.githack.com/w3c/dxwg/prof-Issue-455/profilesont/index.html


kcoyle:
The line
```
dct:conformsTo <http://www.w3.org/ns/shacl#> ;
```
looks to be a correct use of dct:conformsTo. I will say that I still am not content with the fact of blank nodes being given a dct:format, but we can discuss that elsewhere.

kcoyle (approved):
Approving only this portion:

dct:conformsTo <http://www.w3.org/ns/shacl#> ;

rob-metalinkage (approved):
Will look at examples post merge again - tend to agree blank nodes are probably not a good idea in the wild, but as valid RDF they might helpt focuse the attention on the relevant part...

Issue #290, opened 2018-06-27
A short token to specify a profile may be used as long as there is a discoverable mapping from it to the profile's identifying URI [ID5] (5.5)


nicholascar:
Entered from Google Doc

nicholascar:
Suggested Re-wording: "Requirement: A short token to specify a profile may be used as long as there is a discoverable mapping from it to the profile's identifying URI"

rob-metalinkage:
Profile negotiation requires an identifier for the profile to be passed from client to server, and from some source of metadata (such as the server itself) to the client (either at run-time or at some prior configuration step).

With MIME encoding based negotiation, tokens are registered at IANA

There will be potentially many more profiles across application domains than distinct encodings, as profiles are simple content rules scoped to communities of practice, whereas encodings require software clients to be implemented.

Use of URIs as profile identifiers is necessary for discovery of details - so such URIs are available as tokens for content negotiation.

This has some drawbacks however - URIs are hard to encode in query strings, and if a server supports many profiles it may be a burden for humans to read a large set (for example to choose an option). 

One option is to use CURIE syntax prefix:token in which case this would be sematically equivalent to the full URI form. This would require the server implementation to be willing to advertise prefixes it understands, or clients to specify prefix assumptions and to match either full URI or CURIE forms.

there may also be a default prefix, and a global registry of well known profiles - in which case the profile negotiation specification should perhaps declare the default prefix namespace.

IMHO this should be discussed at F2F, taking advantge of joint session with JSON-LD team who are looking at "framing"[1]  which is an allied (but perhaps not identical) concept 

[1] https://json-ld.org/spec/latest/json-ld-framing/

kcoyle:
Added @rob-metalinkage 's comment to f2f agenda.

larsgsvensson:
There was a longer discussion between @azaroth42 and @nicholascar in the google doc so I copy that here for completeness

Rob S:
-1, unnecessary complication and introduces non-uniqueness challenges. Short tokens are good for humans, but humans are not the actors for content negotiation. Each token would need to be publisher specific to avoid collisions, requiring additional time and architecture for no obvious gain.

Nick:
Unnecessary for machine yes but the scope of this profile neg includes humans!

Use Cases like #239 are specifically in scope so Requirements like this one follow from that.

Rob S:
Yes, and I'm questioning that scope :) What's the use case in which a token is discoverable by a human who then types it into the address bar of a browser in a particular place in order to receive content designed for machine interpretation?

[...]

PR #1060, opened 2019-09-09
Assesment of token positions

larsgsvensson:
Text proposal at https://raw.githack.com/w3c/dxwg/larsgsvensson-token-assessment/conneg-by-ap/index.html#profileid
My text assesses the two current positions. I'm not sure whether we've reached consensus yet, so I've left the last paragraph open... The text is put into an issue that we can turn into a note as soon as #290 is resolved.


nicholascar (approved):
Some wording changes maybe but then this Issue text will eventually be removed anyway...

rob-metalinkage (approved):
Again - some wording changes are necessary we can do as a new PR - 

e.g. this is too strong (standards!) 

Further, there are several systems already deployed that build on standards that use tokens for profile identification.

standards => APIs

and the characterisation of the  con-argument is not that the approach is flawed, but the proponents themselves wish to move directly to URI HTTP, which is fine.  Given that the known implementers need the QSA with tokens approach their extended scope should trump those views.

Issue #501, opened 2018-10-26
Registration of target attribute "profile" for the Link-Header


larsgsvensson:
The examples given for the use of the http `Link` header uses the target attribute `profile` that is not yet registered. Cf. https://tools.ietf.org/html/rfc5988#section-5.4

nicholascar:
I don’t understand as I see it register here: 
https://www.iana.org/assignments/link-relations/link-relations.xhtml#link-relations-1

larsgsvensson:
https://www.iana.org/assignments/link-relations/link-relations.xhtml#link-relations-1 is the registry of link _relation types_ (i. e. what comes in `rel="..."`. The issue about `profile="..."` requires a new target attribute in `Link`

rob-metalinkage:
profile is defined - and explained here https://tools.ietf.org/html/rfc6906

AFIACT it is consistent with our model, but not as explicitly defined (seems a bit vague really)

note that is says:
"Note that, at the same
   time, it is possible for profile target URIs to use dereferencable
   URIs and to use a representation (which is outside the scope of this
   specification) that represents the information about the profile in a
   human- or machine-readable way."

larsgsvensson:
I might have gotten this wrong, so apologies up front...
[RFC 6906](https://tools.ietf.org/html/rfc6906) defines the _link relation type_ "profile" (as in
`Link: <urn:example:profiles:1>; rel="profile"` relating the context URI to `<urn:example:profiles:1>` as the profile URI for the context URI).
The case I'm trying to model is that two URIs are linked through the relation "alternate" (saying that one URI is an alternative representation of the other; `rel="alternate"`). In this case I want to say which profile the alternative representation ("link target", in [RFC 5988 lingo](https://tools.ietf.org/html/rfc5988#section-5.1)) has. Since the link relation is already "alternate", I need to have a [target attribute](https://tools.ietf.org/html/rfc5988#section-5.4) "profile", so that I can say
```Link: <http://some.domain/alternative/representation>; rel="alternate"; type="text/turtle" profile="urn:example:profile:1"```
I don't think I can do that with RFC 6906.

rob-metalinkage:
ahh - i see - you dont want the link relation "profile" at all - we need to explain this better both both cases.

so rel="profile" is to designate the profile of the current resource
and rel="alternate"  profile="X"

Agreed - we can see

what about the list of alternates (profiles) supported - does this need a special link rel too?

larsgsvensson:
[...]

PR #1063, opened 2019-09-10
Indicate tokens through separate Link header

RubenVerborgh:
This pull request implements a suggestion [discussed](https://github.com/w3c/dxwg/issues/501#issuecomment-528813489) in #501.

Rather than requiring an extension to the `profile` link relation in order to support tokens, we introduce a new `token` attribute.

Pros:
- orthogonal separation of the token mechanism from other mechanisms
- does not require modifications to existing IETF RFCs and drafts
- avoids repeated indications of the _token_

Cons:
- requires an extra `Link` as opposed to more compactly piggybacking on an existing one

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -1137,11 +1137,14 @@ Accept: text/turtle
 HTTP/1.1 200 OK
 Content-Type: text/turtle
 Content-Location: http://example.org/resource/a.profx.ttl
-Content-Profile: &lt;urn:example:profile:x;token=px>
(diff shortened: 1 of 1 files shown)


larsgsvensson (approved):
We need to have a section in the document where we specify the `Link` attribute `token`, too, similar to the [specification of `profile` in the IETF doc](https://profilenegotiation.github.io/I-D-Profile-Negotiation/I-D-Profile-Negotiation.html#rfc.section.6). I can take care of that.

RubenVerborgh:
@larsgsvensson Sounds good. Feel free to add to this PR, or a later one. Will be important to mention that the scope of the _token_ is the current resource.

nicholascar (approved):
Happy to merge in and do more work following to explain everything. Very happy indeed for Lars to add that explanatory text!

PR #1066, opened 2019-09-10
Added specification of Link attribute token

larsgsvensson:
Preview at https://raw.githack.com/w3c/dxwg/larsgsvensson-link-attribute-token/conneg-by-ap/index.html#link-attributes

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -865,6 +865,14 @@ Content-Profile: http://example.org/profile/x, \
 	  where "token" is defined as in 
 		<a href="https://tools.ietf.org/html/rfc7230#section-3.2.6">section 3.2.6</a>
 	  of [[RFC7230]].
+	</p>
+	<pre id="eg-link-attribute-token" class="example nohighlight" aria-busy="false" aria-live="polite"
+           title="Using the token attribute to link a profile URI to a token">
+# The profile URI in the "anchor" element is linked to the token "px"
+# Further, the relation "type" is used to inform that the anchor resource is of type "prof:Profile"
+
+Link: &lt;http://www.w3.org/ns/dx/prof/Profile&gt;; rel="type"; token="px"; anchor=&lt;urn:example:profile:x&gt;
+	</pre>
       </section>
   </section>
   <section id="functional-profiles">
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -867,11 +867,14 @@ Content-Profile: http://example.org/profile/x, \
 	  of [[RFC7230]].
 	</p>
 	<pre id="eg-link-attribute-token" class="example nohighlight" aria-busy="false" aria-live="polite"
-           title="Using the token attribute to link a profile URI to a token">
+           title="Using the Link attribute "token" to link a profile URI to a token">
 # The profile URI in the "anchor" element is linked to the token "px"
-# Further, the relation "type" is used to inform that the anchor resource is of type "prof:Profile"
 
-Link: &lt;http://www.w3.org/ns/dx/prof/Profile&gt;; rel="type"; token="px"; anchor=&lt;urn:example:profile:x&gt;
+# Further, the relation "type" is used to inform
+# that the anchor resource is of type "prof:Profile"
+
+Link:
+  &lt;http://www.w3.org/ns/dx/prof/Profile&gt;; rel="type"; token="px"; anchor=&lt;urn:example:profile:x&gt;
 	</pre>
       </section>
   </section>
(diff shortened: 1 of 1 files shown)


conneg-by-ap/index.html
```diff
@@ -841,6 +841,43 @@ <h3>Alternate Representations Data Model</h3>
...
+  <section id="link-attributes">
+    <h2>Link Attributes</h2>
+      <section id="link-attributes-token">
+	<h3>Token</h3>
+	<p>
+	  When used in a <code>Link</code> header field,
+	  the attribute <code>token</code> is used to specify
+	  the mapping between the profile URI
+	  (specified using the <code>anchor</code> attribute)
+	  and a <a>token</a> used as an alternative to the full profile URI.
```
RubenVerborgh, line 853: Minor comment, but perhaps the "mapping" language does not need to end up in the document. We talk about this mapping in the context of this WG, but the concept itself is probably not important to the document. So it could be fine to just write that the token is a (typically shorter) string that can be used by clients as an alternative to the full profile URI.

conneg-by-ap/index.html
```diff
@@ -841,6 +841,43 @@ <h3>Alternate Representations Data Model</h3>
...
+	</p>
+	<p>
+	  It should be noted that the scope of the mapping between the token and the profile URI
+	  is per default limited to the current resource
+	  (i. e. the resource identified by the current request URI).
+	  Servers MAY use a larger scope for their mappings
+	  but clients should not depend on that
+	  unless the server documentation explicitly gives other instructions.
```
RubenVerborgh, line 861: …through some other means?

conneg-by-ap/index.html
```diff
@@ -841,6 +841,43 @@ <h3>Alternate Representations Data Model</h3>
...
+	</p>
+	<p>
+	  The ABNF for the profile attribute's value is <code>token</code>,
```
RubenVerborgh, line 864: For the `token` attribute value?
Is `quoted-string`?

Is the Web Linking RFC not more appropriate here?

conneg-by-ap/index.html
```diff
@@ -841,6 +841,43 @@ <h3>Alternate Representations Data Model</h3>
...
+	  where "token" is defined as in 
+		<a href="https://tools.ietf.org/html/rfc7230#section-3.2.6">section 3.2.6</a>
+	  of [[RFC7230]].
+	</p>
+	<pre id="eg-link-attribute-token" class="example nohighlight" aria-busy="false" aria-live="polite"
+           title="Using the Link attribute "token" to link a profile URI to a token">
+# The profile URI in the "anchor" element is linked to the token "px"
```
RubenVerborgh, line 871: Just wondering whether `px` is a realistic name of a token. I also reused it myself, but I have no idea if this would be typical. 2 chars seem really short.

RubenVerborgh (approved):
All good, thanks! Just minor nits.

conneg-by-ap/index.html
nicholascar: Yes, really short. I've been using things like *dcat* or *prov* (Turtle prefixes) or well-known things like ISO + standard numbers. See examples of tokens in old view/format system in use: http://pid.geoscience.gov.au/sample/AU100?_view=alternates

conneg-by-ap/index.html
nicholascar: Are you asking what other means could even be possible? 

Perhaps something like OAI-PMH's metadatPrefix listing where the server lists the prefixes for all profiles for all resources in one place. Very different mechanism but wrappable as Conneg-by-P I've always thought.

conneg-by-ap/index.html
nicholascar: Are you proposing that the document just says "token is a (typically shorter) string that can be used by clients as an alternative to the full profile URI" and then examples, like the one you submitted in your earlier PR, carry the demonstration? I'd prefer this written explanation be included but perhaps I've missed what you're suggesting.

nicholascar (approved):
Fine by me too. Some replies to @RubenVerborgh 's points inline. I've not merged already only so @larsgsvensson can read all those comments before merging.

conneg-by-ap/index.html
RubenVerborgh: Yes, that's what I am proposing. Along with the removal of the "mapping" language.

conneg-by-ap/index.html
RubenVerborgh: I meant to explicitly say something like (verbatim) "through some other means" to emphasize that such means are not part of the spec.

conneg-by-ap/index.html
RubenVerborgh: Yes, that's what I am proposing. Along with the removal of the "mapping" language.

conneg-by-ap/index.html
RubenVerborgh: ^ let's perhaps use one of these?

rob-metalinkage (approved):

PR #1067, opened 2019-09-11
Fixes for the issues in #1066

larsgsvensson:
Fixes for @RubenVerborgh's comments in #1066 

* Removed "mapping" language
* Added "through some other means" to "server gives other instructions"
* The value of the "token" attribute is now `( token / quoted-string )` since that is what's used in [§3 of RFC 8288](https://tools.ietf.org/html/rfc8288#section-3)
* Replaced the dummy example with one from [Nick's geoscience list](http://pid.geoscience.gov.au/sample/AU100?_view=alternates)

Preview in https://raw.githack.com/w3c/dxwg/larsgsvensson-link-attribute-token/conneg-by-ap/index.html#link-attributes-token

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -848,33 +848,36 @@ Content-Profile: http://example.org/profile/x, \
 	<p>
 	  When used in a <code>Link</code> header field,
 	  the attribute <code>token</code> is used to specify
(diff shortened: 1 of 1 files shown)


rob-metalinkage (approved):
looks good.

nicholascar (approved):

Issue #1061, opened 2019-09-09
Suggested updates to definition of PROF constructs


aisaac:
@nicholascar @rob-metalinkage As discussed in our last PROF call, here are suggestions for updating the definitions and notes of some PROF constructs. For now I've focused on the ones of section [8.3 Class: Profile](https://w3c.github.io/dxwg/profilesont/#Class:Profile) as this takes time, and I want to see how you react to my first suggestions before continuing this :-)

- [prof:Profile](https://w3c.github.io/dxwg/profilesont/#Class:Profile) still uses the old definition for profile, at odds with the WG definition ([which is present in the PROF spec](https://w3c.github.io/dxwg/profilesont/#definitions)). The definition should be the official one, verbatim. I can live with keeping the note _"This definition includes [...]"_ in the definition, but it could also live a better life as a new scope note or usage note, in order to keep the "core" definition clean. 

- [prof:hasResource](https://w3c.github.io/dxwg/profilesont/#Property:hasResource). The current definition (_"A resource which describes the nature of an artifact and the role it plays in relation to a profile"_) doesn't work very well imo. The resource doesn't describe by itself - that's the job of a document about the resource - rather, the resource is used as the subject of statements that describe the artefact. So I suggest to replace the definition by _"Relates a Profile to an artefact that implements (some aspects of) it, fulfilling a specific role in relation to the Profile."_

- [prof:isProfileOf](https://w3c.github.io/dxwg/profilesont/#Property:isProfileOf). The current definition (_"The subject of 'is profile of' defines constraints on the object which playes the role of a base specification"_) has a typo and is not well aligned with the new definition of profile (as it focuses on constraint). The style is also at odds with the other definitions (well, at least the one I'm leaning towards ;-) ). I suggest _"Relates a Profile to another data specification that it constrains, extends, combines, or provides guidance or explanation about the usage of"_. 
The usage note should also be refreshed. I suggest _"A Profile may be a profile of one or more specifications. This property is optional, allowing any specification to be declared at the root of a profile hierarchy."_ .This removes the bit on inheritance, which is probably better in other section(s) that discuss the semantics of profile hierarchies. Considering how the notion of inheritance is controversial in the WG, it's better to lower the number of times it appears in the document. The suggested note also gets rid of unnecessary/debatable detail (I don't think using the Profile class is necessary for the root for a profile hierarchy; anyway what's crucial for defining the root is to be a dead-end of a `prof:profileOf` chain, not to be of a given type...)

[...]

Issue #795, opened 2019-03-06
Comments on "inheritance"


kcoyle:
From the ShEx community:

We were unsure what to make of "inheritance", as we are not aware of any generalized notion of inheritance that would make sense across implementation technologies. It would be helpful to motivate this with a testable use case, e.g., a server implementation which responds to Accept-Profile by calculating the "best" response for processing a Profile Ontology description. A test of inheritance would include the notion that inheriting imposes further constraints, and that the entity being inherited from may in turn inherit from something else. ShEx shape inheritance has this behavior.

https://lists.w3.org/Archives/Public/public-dxwg-comments/2019Feb/0001.html

rob-metalinkage:
This is implicit in the definition, and the improved wording for the defintions and usage notes will help, but it needs to brought out more explicitly since the notion seems so hard to grasp - transitive inheritance via isProfileOf is the key machine readable axiom.  This is to be addressed via #732

kcoyle:
@rob-metalinkage You probably will need to explain what you mean by "transitive inheritance" as that may not be in everyone's vocabulary. We've also talked in many other issues about defining explicitly what inheritance means in terms of axioms. Nick did an example at one point, but if we expect it to be actionable then some precise algorithms are going to be needed.

nicholascar:
"transitive inheritance" is now demonstrated in the PROF ED in examples and also in a [long note after the `prof:isTransitiveProfileOf` property](https://w3c.github.io/dxwg/prof/#Property:isTransitiveProfileOf).

The axiom `dct:conformsTo owl:propertyChainAxiom ( prof:isProfileOf dct:conformsTo )` also now present in the ontology defines the entailments that can be made with `prof:isProfileOf` and, since no equivalent axiom is declared for `prof:isTransitiveProfileOf`, it is clear how the usage of the properties differs.

Regarding the sentence:

> It would be helpful to motivate this with a testable use case, e.g., a server implementation which responds to Accept-Profile by calculating the "best" response for processing a Profile Ontology description.

This is a matter for ConnegP and notes on "best" responses, or at least how responses from hierarchies are calculated, [are given in the ED of that document](https://w3c.github.io/dxwg/connegp/#getresourcebyprofile).

Issue #521, opened 2018-10-31
Consider the addition of a conformanceTarget property


nicholascar:
Following the OGC's Modular Specification [[modspec]], consider adding a property conformanceTarget (domain: prof:Profile, range: ?) that allows for the linking of Profiles to the *sorts of thing* or *instances of the sorts of thing* that are expected to conform to it.

What might the inverse property of this be?

kcoyle:
Nick, can you give a functional link to the OGC spec? Thanks.

smrgeoinfo:
In answer to @kcoyle question [The Specification Model - A Standard for Modular specifications (08-131r3) ](https://portal.opengeospatial.org/files/?artifact_id=34762)

smrgeoinfo:
An example spec using this approach:
https://docs.opengeospatial.org/is/16-008/16-008.html . 
Requirements for three standardization target types are considered:
• Logical Model
• Encoding
• Data instance.

smrgeoinfo:
The Energistics Energy Industry Profile of ISO 19115-1:2014 (v1.1) (https://www.energistics.org/wp-content/uploads/2017/03/EIP_v1.1.pdf) Defines a URI to identify instance documents conforming to the profile in section 4.2.9 Metadata Standard:
http://w3.energistics.org/energyml/profiles/EIP/v1.1/metadataStandard_citation.xml

dr-shorthair:
Also see this paper: [Configuration management of a system of interdependent standards](https://doi.org/10.1109/SIIT.2011.6083604) as mentioned in https://github.com/w3c/dxwg/issues/72#issuecomment-382207466

dr-shorthair:
![figure 1](https://user-images.githubusercontent.com/608303/47889448-23b40700-de9e-11e8-9e8e-fa300ee9ca5f.png)
_Figure 1. Conceptual model of normative dependencies in ISO/TC 211 standards_

![figure 2](https://user-images.githubusercontent.com/608303/47889462-3b8b8b00-de9e-11e8-9c74-b5f3beb48286.png)
_Figure 2. Conceptual model of a modular standard (from OGC, 2009)_

nicholascar:
It looks like we might have bypassed this Issue by 1. detailing the competency questions used in the definition of `prof:isProfileOf` and by 2. dealing with resources via roles.

For 1.:
**IF** A isProfileOf B
**AND** X conformsTo A
**THEN** X conformsTo B

For 2.:
[...]

PR #1074, opened 2019-09-13
Editorial fixes for Prof 3 pwd candidate prior to introducing final substantive change proposals.

nicholascar:
See the rendered HTML: https://raw.githack.com/w3c/dxwg/prof-3PWD-candidate/prof/


rob-metalinkage (changes requested):
That a lot of editorial changes - at least in separate commits.  There are a few grammar fixes needed on section on constraints and I think the property chain axiom is a substantive change - I'd rather have the removed and in a separate PR as there is an active issue discussing this and its the only change to the model in this PR.

nicholascar:
Following a discussion with @rob-metalinkage: I've marked the axiom a "feature at risk" with an issue (#1078) since it's the substantive addition in this PR. With it marked as at risk (in two places: the Introduction and at the point of definition in the Specification), it can be included now and removed from the CR if indeed it becomes problematic.

aisaac:
I'm trying to go through the pile of changes....
@nicholascar @rob-metalinkage why is the Requirements appendix removed? The note says it's redundant. What replaced it?

nicholascar:
@aisaac the Requirements Appendix was meant to be a list of Requirements & responses to them, see https://www.w3.org/TR/dx-prof/#requirements. However, we didn't use that approach much preferring instead to place Issues in the doc and deal with their implications via PRs. So the Requirements listing was just a method that was outside of our editing processes. 

For the Requirements that were directly from the UCR, we listed them elsewhere: https://www.w3.org/TR/dx-prof/#motivating-requirements.

I think we should go back and check off against UCR requirements, especially to check what Guidance need not do, but we just don't need the listing as we had it in this appendix in 2PWD.

aisaac:
@nicholascar ok I think I was lead on the wrong track by the note saying 'duplicate'. 'obsolete' would have been more precise!

aisaac:
For the record I'm lukewarm on the change to the "family of documents" section. But well I guess that you had to adapt. and we can still tweak it depending on what happens in the coming two months.
I would argue for removing the typio on "This document is one from set of documents on profiles" though. It's "This document is one from a set of documents on profiles"!

aisaac:
Big typo in the intro on "A specification is "A basis for comparison; a reference point against which other things can be evaluated." (see the definition below) and a A profile, such as a profile or perhaps an application profile, is defined as "A [data/application] specification that constrains, extends, combines, or provides guidance or explanation about the usage of other [data/application] specifications""

aisaac:
There's going to be a lot of push-back if the definitions do not exactly represent the agreement at #963. And I'm probably going to be part of it. The formal disappearance of the definition of "data profile" (even if you tried to provide a workable alternative) is going to trigger a lot of discussions. Some of them already happened earlier (I don't know in which issue). I am far from convinced that representing a more general view on profiles is worth jeopardizing the spec and firing up all these discussions. The more general view could be added on top of that, at a later stage.

aisaac:
For the record, I am **not** reviewing the sections on related work. I have no time checking the amount of text, I've made (or would make) comments as part of other issues. And I believe that whatever text (minus typos of course!) is worth sharing with the group at this stage.
Same for the ShEx example.

aisaac:
I believe there's a typo on "that is derived from, but separate to, the namespace for PROF". "separate" is also to be used with "from", isn't it?

aisaac:
I'm not sure why 521, 507, 802, 453, 795 are not mentioned in the doc anymore. As long as they remain open on github while they're under discussion, that's fine, but this makes it less clear, what's the rationale for having issues referenced or not in the document.

aisaac:
In 6.8 Repetition in "Constraint Languages are frequently used in profile profile definitions"

aisaac (changes requested):
OK I've reviewed what I could review. All comments that I think could (some should) be addressed are now above!

aisaac:
I forgot to say - impressive work, @nicholascar !

kcoyle:
There seems to be content here that is not resolved in issues. For example, @aisaac are you satisfied with the resolution of "prof:isTransitiveProfileOf needs more convincing case and/or example #643"? Also, we have: "Improve defintion of prof:isTransitiveProfileOf #1071", "Suggested updates to definition of PROF constructs #1061" not to mention about 80 other open issues. What is being proposed for all of the open issues that are not considered "at risk"?

aisaac:
@kcoyle I think it is ok to have a PR doing things not in issues, and not tackling all open issues. There could be another later PWD candidate to address the open issues, for sure. Maybe it's already in preparation. This PR is already big; I think in general they should be more granular.

kcoyle:
Regarding the definition of profiles (for this and for the conneg document), the approved language is:

>Data Profile

>A data specification that constrains, extends, combines, or provides
guidance or explanation about the usage of other data specifications.

>Data Specification

>A specification, with human- and/or machine-processable representations,
that defines the content and structure of data used in a given context.

While it probably makes sense to say in the definitions area that the word "profile" will be used in the document to mean what is defined as "data profile", the group did vote to define the term as "data profile."

kcoyle:
@aisaac You may have missed the information on the deadline. The date and time of tomorrow's meeting is the final deadline for a CR-ready version of PROF and of conneg. There will not be additional "PRs" except in support of a Working Group note. That is why I am questioning the disposition of open issues. I'm trying to understand whether this version is considered final or if the editors are planning to continue the work as a note.

rob-metalinkage:
The PR is not addressing the definitions - thats a substantive change that needs to be isolated to a separate PR. I think the problem is in the title - i'll rename it..  
Fix up typos and grammar for now - dont change scope of this PR.  This will help reviewing PRs with the substantive changes.

aisaac:
Thanks for the clarification @rob-metalinkage . @kcoyle this was indeed my approach when looking at this PR. Not requiring that it does everything. In fact I implicitly said earlier to @nicholascar that it would be great if we would stop piling things in here, because otherwise I can't review them :-)

nicholascar:
@aisaac I've addressed 4 of your comments requesting changes (see commits above linking to comment IDs) and regarding https://github.com/w3c/dxwg/pull/1074#issuecomment-531942290 (Issues removed): 

* #521 - the proponant (Rob) is happy for this to be added as Future Work and a poll so far indicates watching participants' agreement so I thought it would serve no purpose to leave in the doc. We should address it, and all other future-work tagged issues after PWD release
* #507 - duplicate for the now "at risk" issue (#1078) in the doc which I added to highlight the proposed axiom. There are about 4 issues that all get at the same thing
* #802 - one of the semi-duplicates for the above. The axiom answers the question so if it's accepted then this Issue is for Guidance (which I think is the right place for it)
* #453 - this has been addressed in a commit above, as per your request. I thought it would be so I'd like to see the PR pulled through to ED so I can tag the Issue as due-for-closing. Seems onerous to keep back-and-forth tagging (chicken and egg)
* #795 - I think this is now well addressed by the isTransitiveProfileOf semantics (linked to above) and the axiom so I've taken it out of this PR. I will reference the proposed solution, and the example in the doc from SKOS as per Karen's comment in the Issue, in an email to the ShEx group that raised it.

aisaac (approved):
OK @nicholascar thanks for the explanation and the last edits which indeed fix the typos that I had found and revert the definition to the official one.

And I agree with removing the reference to the couple of issues you flag as "(near-)duplicates". Even if they are still open, the document makes reference to "nearby" issues. It would be confusing to point the reader to all the issues in our web - it's already quite difficult for ourselves...

rob-metalinkage (approved):

Issue #1064, opened 2019-09-10
Remove "Alternate ways of conveying URI / token mappings"


RubenVerborgh:
There currently is a not in the conneg-by-ap document titled "Alternate ways of conveying URI / token mappings". However, every single listed option is incompatible with the `Content-Profile` header as defined.

I propose to remove this note once we have finalized the token mechanism (for instance, through #1063).

larsgsvensson:
@RubenVerborgh scripsit:
> I propose to remove this note once we have finalized the token mechanism (for instance, through #1063).

+1

larsgsvensson:
Addressed in PR #1075

nicholascar:
The "Alternate ways of conveying URI / token mappings" is no longer in the ED at https://w3c.github.io/dxwg/conneg-by-ap/

PR #1075, opened 2019-09-14
Updated section "Token mappings"

larsgsvensson:
Updated normative text on how to use the `Link` header
Removed Note about "Alternate ways of conveying URI / token mappings" as per #1064

Preview at https://raw.githack.com/w3c/dxwg/larsgsvensson-update-token-mappings/conneg-by-ap/index.html#listprofiles-tokens

Since this introduces normative text, please review carefully and also please don't merge until we've had @RubenVerborgh's review

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -1162,7 +1162,7 @@ Link:
             need to provide a <a>token</a> / URI mapping via this <strong>list profiles</strong> function.
 	    When implementing the HTTP functional profile of this specification,
 	    a server MUST include a <code>Link</code> header ([[RFC 8288]])
-	    for each <code>profile</code>/<code>token</pair> in its response.
+	    for each <code>profile</code>/<code>token</code> pair in its response.
 		  Each <code>Link</code> header MUST use the <code>Link Target</code>
 		  "<code>http://www.w3.org/ns/dx/prof/Profile</code>"
 		  and MUST contain the following <code>Target Attribute</code>s:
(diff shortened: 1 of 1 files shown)


RubenVerborgh (approved):
Reviewed and confirmed technically accurate.

As an editorial note, I wonder whether we need the concept of a "token / URI mapping". We might be able to write down the same thing by just referring to tokens. Will add optional suggestion as comment.

conneg-by-ap/index.html
```diff
@@ -1157,16 +1157,43 @@ <h4>HTTP Alternate Representations Data Model Implementation</h4>
         <section id="listprofiles-tokens">
           <h3>Token mappings</h3>
           <p>
-            If HTTP <a>server</a>s wish to allow <a>client</a>s to identify <a>profile</a>s via <a>token</a>,
+            If HTTP <a>server</a>s wish to allow <a>client</a>s to identify <a>profile</a>s via <a>token</a>s,
             in addition to the the mandatory identification of <a>profile</a>s via via URI, the <a>server</a>s will
-            need to provide a <a>token</a> / URI mapping via this <strong>list profiles</strong> function. In this
-            HTTP functional profile of this specification, this may be implemented in a number of ways. The preferred
-            way is to add "token" parameters to the HTTP Content-Profile and Link HTTP headers.
+            need to provide a <a>token</a> / URI mapping via this <strong>list profiles</strong> function.
```
RubenVerborgh, line 1162: We _could_ write this instead as "needs to indicate for every supported token to which profile URI it corresponds". This eliminates the concept of a "token / URI mapping" (and associated questions such as whether it needs to be complete, injective, etc.).

nicholascar (approved):
fixed a couple of small HTML things, now all approved

PR #1076, opened 2019-09-14
Marked http implementation as at-risk

larsgsvensson:
Addresses ACTION-371
Preview at https://raw.githack.com/w3c/dxwg/larsgsvensson-http-at-risk/conneg-by-ap/index.html#http and https://raw.githack.com/w3c/dxwg/larsgsvensson-http-at-risk/conneg-by-ap/index.html#features-at-risk

PR #1077, opened 2019-09-14
Updated Changes section

larsgsvensson:
I've updated the changes section so that it looks more like the previous one. Also I've commented out all editorial changes since people implementing the spec will be more interested in what normative content has changed.
This PR depends on changes introduced in #1076 so don't merge this one until after that has been handled.

Addresses ACTION-375


nicholascar (approved):

rob-metalinkage (approved):
seems like a little conflict to resolve before merge - but presuming just use this PR to resolve those

PR #1083, opened 2019-09-17
Added order of precedence text

rob-metalinkage:
Issues 505 and 575 editorial tweaks.


conneg-by-ap/index.html
```diff
@@ -840,6 +840,15 @@ <h3>Alternate Representations Data Model</h3>
         also given in Functional Profiles of this specification are also given in that appendix.
       </p>
     </section>
+    <section id="precendence">
+      <h3>Order of Precedence for Implementation Profiles</h3>
+      <p>
+        It is possible that multiple functional profiles MAY be implemented by a service, including HTTP, QSA and possibly other approaches. In such clases a client MAY specify conflicting choices via different mechanisms. The order of precedence determines which mechanism MUST be used. 
+      </p>     
+      <p>
+        The order of precedence is QSA over HTTP - on the basis that human agency may be used to determine URL elements whereas client software is more likely to determine header content, and header content is less visible to human agents. If other profiles of the abstract model are defined each MUST specify order of precedence.
```
larsgsvensson: I don't think this is about visibility to human agents but about URIs trumping headers: If my URI says .../a/resource.profx.xml and the headers say `Accept: text/turtle` and `Accept-Profile: urn:example:profileY` (=`profy`, not `profx`), I still expect to get XML content conforming to `profx`.
Suggestion: "on the basis that the more exactly a URI points to a resource, the more likely it is that the agent requests that precise resource irrespective of which HTTP headers are sent with that request."

larsgsvensson (changes requested):

larsgsvensson:
Since this is also for #575 I add @aisaac as a reviewer, too.

larsgsvensson (approved):

aisaac (approved):

Issue #1088, opened 2019-09-20
Conneg Editorial Nits


larsgsvensson:
@tombaker has [reported some editorial nits](https://lists.w3.org/Archives/Public/public-dxwg-wg/2019Sep/0682.html) in the conneg-by-ap WD:

I see that PROFPWD is listed under B1 and assume this can be updated and moved to B2 later. (Can someone please confirm?)

Some fixable glitches:

Section 1
These narrower profiles constrains

4
demonstrations

5.4
so that profiles, (the comma)

6.2
may no be capable of dereferencing

6.3
representation of the requested resource represented conforming

7.1
An functional profile

nicholascar:
This Issue is the sole subject of PR #1090

larsgsvensson:
The corresponding PR has been merged.

PR #1090, opened 2019-09-22
Issue 1088 - PROF ref to B2 & 6 x typo errors fixed

nicholascar:
Addressing all points in #1088


kcoyle:
By my understanding, these are the kinds of editorial changes that do not need to be made before Oct. 3 - not saying you shouldn't do them, but that if you want to take more time that should be ok.

larsgsvensson:
Yes, I also thought of letting #1088 be a bucket where we can collect several minor nits and then fix them in one go.

rob-metalinkage (approved):

larsgsvensson (approved):

PR #1104, opened 2019-09-29
Fix Protege load errors

### LINKED ISSUE #1100: Wierd Protege 5.0 load issue with lastest versions of dcat.rdf and dcat.ttl - opens as inconsistent
As an add on to #944, I opened both the latest version of dcat.ttl and dcat.rdf by URL from within Protege 5.5.0 Build beta-8.  In both cases, the ontologies loaded OK, but contained spurious Error1, Error2, and Error3 (only in the .rdf version) entities (see screen shot below).

Strangely enough, if I download the files from github and then open them in Protege, I get the very same errors!  However, if I remove the spurious subclass statements, save the file, then reopen it, the errors stay gone!  Since there are no obvious Error# axioms in the online files, this seems quite strange!  Not quite sure where the error is coming from.

Screen Shot 2019-09-26 at 3 09 19 PM

dr-shorthair:
Remove sub-class axiom (restriction) and add type declarations to suppress Protege errors
Fixes #1100

### DIFF
--- a/dcat/rdf/dcat-external.jsonld
+++ b/dcat/rdf/dcat-external.jsonld
@@ -1,102 +1,102 @@
 {
   "@graph" : [ {
     "@id" : "_:b0",
-    "seeAlso" : "https://orcid.org/0000-0001-9300-2694",
-    "homepage" : "http://www.andrea-perego.name/foaf/#me",
-    "name" : "Andrea Perego"
+    "seeAlso" : "http://fadmaa.me/foaf.ttl",
+    "name" : "Fadi Maali"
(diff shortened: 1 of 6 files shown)


davebrowning (approved):
Don't have the means to test, but change - to dcat-external.ttl, and dcat.ttl - matches the specification/description.

riccardoAlbertoni (approved):

riccardoAlbertoni:
I have tested the dcat.ttl  and dcat-extended.ttl, as the other RDF encodings have been automatically generated.

PR #1110, opened 2019-10-02
RDF and JSON versions generated using OWL-API

### LINKED ISSUE #1100: Wierd Protege 5.0 load issue with lastest versions of dcat.rdf and dcat.ttl - opens as inconsistent

dr-shorthair:
Fixes #1100


davebrowning (approved):

riccardoAlbertoni (commented):
I  have tried to upload these files on Protege 5.5.0, and they do not arise errors.

However, I've noticed that the following warning appears when loading dcat.rdf, dcat.jsonld

> Illegal redeclarations of entities: reuse of entity http://www.w3.org/ns/dcat#theme in punning not allowed [DataProperty: theme, ObjectProperty: theme]

Such a warning is not present when loading dcat.ttl.
Should we get rid of this?

riccardoAlbertoni:
In case we are considering to regenerate the rdf and json serializations to address my previous comment, I would suggest to do it after accepting the PR #1125 and #1127.

andrea-perego:
@riccardoAlbertoni said:

> In case we are considering to regenerate the rdf and json serializations to address my previous comment, I would suggest to do it after accepting the PR #1125 and #1127.

Which is the status of this PR? I see both PRs #1125 and #1127 have been merged. Can be safely merge also this one?

riccardoAlbertoni:
I think we might want to regenerate the JSON and RDF to include the changes made in PR #1125 and #1127. As far as I have understood no all the usual tools to convert are working as expected.  

@dr-shorthair:  It seems that you have figured out the right combination of tools,  would you mind to taking care of this conversion?

dr-shorthair:
OK - done.

riccardoAlbertoni (approved):

PR #1116, opened 2019-10-03
first draft of recharter document

pwin:


kcoyle:
line 115 "Maintain and revise the Data Catalog Vocabulary, DCAT, taking account of related vocabularies and the extensive work done as part of developing version 2 that is part of the product backlog."

Not clear what this means - possibly "Maintain and revise the Data Catalog Vocabulary, DCAT, including additional feature requests from the DCAT user community."

pwin:
What is means is that we have to maintain the vocabulary (respond to issues etc) and revise with the 'future-work' tag and any other proposals that come forward with the research to develop the UCR.

Perhaps that is a clearer way of stating it.  What do you think @kcoyle ?

kcoyle:
How about: "Maintain and revise the Data Catalog Vocabulary, DCAT, including features deferred from version 2 and any additional feature requests from the DCAT user community."

makxdekkers:
> How about: "Maintain and revise the Data Catalog Vocabulary, DCAT, including features deferred from version 2 and any additional feature requests from the DCAT user community."

Or maybe even shorter: "_Maintain the Data Catalog Vocabulary, DCAT, taking into account feature requests from the DCAT user community_."

As I see it, 'revision' is part of maintenance and the features deferred from version 2 are also feature requests from the community.

Can I also suggest to delete the first four paragraphs of the "Goals" section? I think the paragraph that starts "The goal ..." is the only one that is about, well, the goals. If we really want to keep the first part, it could go into a "Context" section following the Goals, but I don't think it is even necessary.

makxdekkers:
I am also not sure about the sections 2.1 Input Documents and 2.3 Success criteria under the Scope section. Should the Scope section not say just what is in scope and what is out of scope? Input documents and Success criteria seem to be separate things from the scope and maybe should have their own top-level sections, for example after the Deliverables section.

makxdekkers:
In the Deliverables section, the list needs to be updated: DCAT 1.1 is not what the group will be working on. Do we still want to keep working on the application profile guidance and the profile negotiation? I understood that the 'evergreen' status is just for DCAT, not for the other deliverables of DXWG. Could those other topics maybe be handled by dedicated community groups? It's not that I think these topics are not important but widening the scope of a group that is primarily responsible for DCAT also means spreading resources (of which we have not a lot) more thinly. I also note that work on application profiles and profile negotiation is not mentioned in the Goals or the Scope section.

makxdekkers:
In the section on Out of Scope it says "_The Dataset Exchange Working Group will not create application profiles or metadata standards that only apply to very specific domains (such as particle physics, accountancy, oncology etc.)_"
Does this mean we won't create application profiles at all, or only not create domain-specific application profiles? What about the plural in 'metadata standards'? DCAT is a singular.
Could we just say that the group will not consider requirements that are (strictly?, predominantly?) domain-specific?
And then we could maybe mention in the In Scope section that the group will primarily look at cross-domain requirements.

agbeltran:
Thanks, Peter. About the timelines, given that there are already several use cases and requirements from the current phase that have not been addressed and that we are entering this 'ever green' form, is it still necessary to have a period exclusively dedicated to use cases and requirement gathering?

If helpful for people, you can see the new charter here: https://raw.githack.com/w3c/dxwg/recharter_1/charter/

pwin:
@makxdekkers  re: https://github.com/w3c/dxwg/pull/1116#issuecomment-538252093 
If we are to move Conneg to CR in the next month or two then we need to include it in the charter (afaik) to carry it through the remainder of rec track

pwin:
@agbeltran  - re Timeline, The activities are generally concurrent, and the dates are for delivery.  I've made that clearer.  Getting UCR clarified early on will give us a chance to make these more realistic.

kcoyle:
@makxdekkers As for the other deliverables, Philippe explained that to re-charter we basically must stick to the original deliverables - otherwise it becomes a new charter and will go through a whole new process, which he advises against.

kcoyle:
I don't think we should remove the part about "the group will not create application profiles" - if we remove that we expand our charter, and that could be a red flag for W3C management. Is there really an intention to create profiles, or just to give guidance on profiles?

kcoyle:
Philippe has reviewed the document and will give us a few "tweaks".

aisaac:
Just a comment that the POE group could be replaced by "ODRL Community Group" with URL https://www.w3.org/community/odrl/

PR #1119, opened 2019-10-08
copy edits to section 7.4

nicholascar:
Karen's edits.

Since Karen made the edits, consider me making the PR a review from me - I'm happy. Still need two others though.

### DIFF
--- a/conneg-by-ap/index.html
+++ b/conneg-by-ap/index.html
@@ -10,7 +10,7 @@
   <section id="abstract">
     <h2>Abstract</h2>
     <p>
-      This document describes how Internet clients may negotiate for content provided by servers based on the 
+      This document describes how Internet clients may negotiate for content provided by servers based on the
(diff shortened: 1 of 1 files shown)


rob-metalinkage (changes requested):
Copy edits in general improve the text and clean nits - good stuff

A small number of potential ambiguities have crept in we need to fix first:

suggests a data representation may conform to a single profile - but of course it may conform to all more general profiles as well

saying "specification _or_ a profile" introduces a false dichotomy - by principles of reflexivity a specification is a profile of itself - and a profile is a specification by definition.  

"list of conformant profiles"   - its not profiles which are conformant so this is misleading - just "list of available profiles" is better,

"own definitions of conformance" is over-specific - many specifications will not explicitly "define" conformance - but instead refer to (or rely on implicitly) shared nations of conformance.

larsgsvensson (approved):

rob-metalinkage (approved):

PR #1132, opened 2019-10-20
Conneg editorial

nicholascar:
editorial issues picked up after implementation


kcoyle (changes requested):
We don't have a Working Group vote to move this to CR status, and I don't think that we can do that in this charter. So it needs to stay as a WD, and we need to figure out what we need to do within the charter.

nicholascar:
Sure, will change it back shorty: was just following a request from poll comments to present it as a CR doc for voting.

rob-metalinkage (approved):
NB if not going out as CR need to push out as a PWD as early as possible..

kcoyle (changes requested):
Lines 778-779 are unclear:

"This specification provides functional profiles of it for two environments: HTTP and human browser (Query String
        Arguments) as well as a single action functional profile relevant to multiple environments."

If "of it" means "of the abstract model" then make that change.

There is no definition of what a "single action functional profile" is. Is "single-action" really necessary there? As it is unclear, it would be best to just say "a functional profile. Also "relevant to multiple environments" without giving examples is unclear. Again, if this isn't necessary it would be best to drop it. I'd suggest leaving the previous wording as is.

kcoyle (changes requested):
line 790 "too" -> "also"

line 792 ":" -> either comma or semicolon

line 2043 "conformance to the this" - remove "the" or "this"

2044 "PROF's recommendation" - PROF is a vocab and cannot recommend. I think you mean the recommended usage of PROF. But since you give an example, you can say "An example of PROF usage is for a claim of conformance to ..."

kcoyle (changes requested):
2064-5 "suggest for use." No idea what this means or how it fits into the sentence - is a word missing? It would also be good to break up that sentence as it has too many clauses and is very hard to read. Then something about "suggesting" might fit in there somewhere. I'd suggest a change but I don't understand the sentence.

larsgsvensson (commented):
Fine with me, but @kcoyle should have the last word if her comments have been addressed

kcoyle:
I think my edits are in there, although there seem to be line number changes since I made my comments so I may have missed some.

I would really like to fix this run-on sentence:

>This specification provides functional profiles of the abstract model for two environments: HTTP and human browser (Query String
        Arguments) as well as a functional profile that specifies a single action - confirming which profile(s) a server response conforms to -
        that is relevant to multiple environments, such as the HTTP and human browser environments.

and suggest (if the concepts have been captured):

This specification provides functional profiles of the abstract model for two environments: HTTP and human browser (Query String
        Arguments).  It also provides a functional profile that has the single function to respond with which profile(s) a server response conforms to, and which can be used in both environments.

agreiner:
Putting this here instead of just email, sorry for any duplication:
Can we get rid of the instances where header-based negotiation is called "http"? Query strings are part of http, too. Browsers use http, including headers. And query strings are as likely to be used programmatically as in a browser, so I would suggest we stick with header-based negotiation and query-string-based negotiation. Here I suspect you meant to refer to all the non-header-based approaches as a single entity. I'm not sure how best to refer to that, but maybe it would work to call it URL-based negotiation.

nicholascar:
@agreiner I don't have an in-principle objection to name changes but can this please be deferred until after the next PWD which is to be voted on in subgroup in 12 hours and then put to plenary next week? This change would require hunting down a lot of references throughout the document which I don't want to have to do just yet. 

Please just raise an Issue for this and we can address it post 3PWD.

nicholascar:
@kcoyle I've addressed your latest comment, please re-review

kcoyle (approved):

PR #1134, opened 2019-10-24
changing Link header rel="self" to rel="default"

nicholascar:
In testing aspects of my implementation and Rob's one in devlopment, we noticed that the Link header indicating a default response with rel="self" is misleading: I had inadvertently interpreted "self" to mean the representation returned which may or may not be the default prepresentation.

Since the profile conformance and the format of the representation returned are given by `Content-Profile` and `Content-Type` respectively, they indicate the "self" result and this we would be clearer in using rel="default".

Note that in the [HTTP header mapping to the AptR model](https://w3c.github.io/dxwg/conneg-by-ap/#http-altp), we have made the fact that this rel="self" is the default indicator by mapping it to `altp:hasDefaultRepresentation`.


larsgsvensson:
According to the [IANA link relation registry](https://www.iana.org/assignments/link-relations/link-relations.xhtml) there is no link relation "default". "self" is defined as "Conveys an identifier for the link's context" with a reference to [RFC4287](http://www.iana.org/go/rfc4287). If we want to use "default", we should register that relation.

dr-shorthair:
Perhaps 'canonical' which "Designates the preferred version of a resource (the IRI and its contents)."

nicholascar:
OK, we'll discuss "canonical", but the wording of "default" better indicates the intention of the link relation: it is the default - what you get if you send no futhter instructions, not preferred, the meaning of the canonical.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

nicholascar:
@larsgsvensson if you think this is sensible, I will register *default*. I 'm happy to do this and I see no other candidate.

rob-metalinkage (approved):

larsgsvensson:
@nicholascar scripsit:
> I will register default. I 'm happy to do this and I see no other candidate.

How would you define the "default" relation?

nicholascar:
Here is the definition for *self*:

> Conveys an identifier for the link's context. 
(https://www.iana.org/assignments/link-relations/link-relations.xhtml)

Or, in extended form, from the Atom spec:

> The value "self" signifies that the IRI in the value of the href attribute identifies a resource equivalent to the containing element.

with the full normative instruction:

> atom:feed elements SHOULD contain one atom:link element with a rel attribute value of "self".  This is the preferred URI for retrieving Atom Feed Documents representing this Atom feed.
(https://tools.ietf.org/html/rfc4287)

So, whether used in an Atom `<link>` element or in an HTTP header `Link`, the intention of `self` is not as we have used it. The original Atom intention was all about how to get the current resource's feed URI but that's been bastardised, like many other Atom concepts, for generic HTTP use. It's still, however, not what we want as all of the `Link` headers we provide may refer to the same context (URI), just with different `type` or `profile` properties.

How about `rel="canonical"`?

The definition for `canonical` reads:

> Designates the preferred version of a resource (the IRI and its contents).
(https://www.iana.org/assignments/link-relations/link-relations.xhtml)

looking into the spec defining it though, it's extended form reads:

> "canonical" can be described informally as the author's preferred version of a resource.  More formally, the canonical link relation specifies the preferred IRI from a set of resources that return the context IRI's content in duplicated form.
(https://tools.ietf.org/html/rfc6596)

To me this reads as much closer to what we want! It's still a little off (the "...IRI from a set of resources..." bit indicates the intention is to create `Link`s with differing target URIs, rather than the same target URIs but with differing Target Attributes) but the result is fine: this (the "canonical") `Link` header will give you the *author's preferred version of a resource* which I think can reasonably be interpreted as the default.


If you are both OK with this, I will push an updating commit to this PR to replace `rel="self"` with `rel="canonical"` throughout the document and not have to register anything.

rob-metalinkage (approved):

nicholascar:
I'm closing this PR in favour of PR https://github.com/w3c/dxwg/pull/1166 which uses *canonical* instead of *default*, as per discussion here.

PR #1163, opened 2019-11-12
Add ref to #505 - order or precedence and remove MUST

rob-metalinkage:
no mechanism is available - tweaked #505 to make question clearer. 
Removed overspecification using MAY and MUST as per discussions in issue
Hopefully this also clarifies and addresses @agreiner concerns here.

### DIFF
--- a/connegp/config.js
+++ b/connegp/config.js
@@ -5,7 +5,7 @@ var respecConfig = {
     previousPublishDate: "2019-04-30",
     previousMaturity: "PWD",
     prevRecURI: "https://www.w3.org/TR/2019/WD-dx-prof-conneg-20190430/",
-    testSuiteURI: "https://github.com/w3c/conneg-prof-testing",
+    testSuiteURI: "https://github.com/w3c/prof-conneg-testing",
     implementationReportURI: "https://w3c.github.io/dxwg/connegp-implementation-report/",
     canonicalURI: "TR",
     editors: [{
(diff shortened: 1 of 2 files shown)


agreiner (commented):
If we go with the developer-must-also-create-standard approach, I would suggest saying "If other potentially conflicting functional profiles of the abstract model are defined, then order of precedence will need to be established." There's no need to make someone figure out all the possible combinations of ways others may have tried to code their APIs. In fact, without a mechanism for registering these, it would be impossible.

nicholascar (approved):

nicholascar:
I've adopted @agreiner's suggested wording.

agreiner (changes requested):
I don't see my suggested text in there.

nicholascar:
We accepted both your suggestions and see https://github.com/w3c/dxwg/pull/1163/files#diff-530b9656f9dfedce3c4078b99c85686eR759, the sentence "If other potentially conflicting functional profiles of the abstract model are defined, then order of precedence will need to be established."  is now at line 759.

larsgsvensson (approved):
I approve the changes, although I'm always a bit hesitant to use passive constructs ("order of precedence will need to be established") since that does not make clear **who** is responsible for establishing this order.
I won't merge the commits, since I think we should wait for @agreiner's review before doing that.

agreiner (approved):
Interesting, that didn't show up in the changes shown when clicking "View Changes" until after my comment. Looks okay now.

PR #1166, opened 2019-11-19
Conneg self canonical

nicholascar:


rob-metalinkage (approved):

larsgsvensson (approved):
I'm fine with replacing `rel="self"` with `rel="canonical"` *for the default resource*.
I question, though, if we should completely discard the use of `rel="self"` since there can be value in showing the properties of the resource given in the `Content-Location` header field. Maybe we can make that OPTIONAL.

PR #1167, opened 2019-11-19
Conneg Implementation Report

nicholascar:
The start of an Conneg Implementation Report document, formatted as per the DCAT example.


connegp-implementation-report/index.html
```diff
@@ -126,13 +126,13 @@ <h3>Functional Profiles Identified</h3>
             </tr>
             <tr>
               <td><code><span style="white-space: nowrap"><a href="http://www.w3.org/ns/dx/conneg/profile/qsa">cnpr:qsa</a></span></code></td>
-              <td>QSA Functional Profile</td>
-              <td>For systems negotiating for content with the use of URI [[RFC3986]] Query String Arguments and using the <code>_profile</code> argument key and <code>_profile=all</code> key/value pair</td>
+              <td>URL QSA Functional Profile</td>
+              <td>For systems negotiating for content with the use of URL [[RFC3986]] Query String Arguments and using the <code>_profile</code> argument key and <code>_profile=all</code> key/value pair</td>
```
larsgsvensson: The mention of RFC 3986 in the context of URLs is not really correct since that is the **Uniform Resource Identifier (URI): Generic Syntax** specification. Perhaps "the use of URL Query String Arguments (cf. [[RFC3986]] for a definition of Query String Arguments in the context of URIs)".

larsgsvensson (approved):
I approve in general, but propose that we look at the relation between URI query strings and URL query strings in the context of using RFC 3986 as a reference

connegp-implementation-report/index.html
nicholascar: Done

rob-metalinkage (approved):
yep its just a stub - will get my stuff in as soon as I go live.

PR #1174, opened 2019-11-21
Update to Intro

davebrowning:
For PR (optional, but we need to do it anyway


dr-shorthair (approved):

andrea-perego (approved):

andrea-perego:
@davebrowning , @dr-shorthair , the current and old text points to https://github.com/w3c/dxwg/wiki/DCAT-implementation-evidence for instructions on how to report implementations, but the instructions are outlined at the end of the page, with also the following caveat:

> (Please ignore this section, the following instructions are not final, they are an attempt to exemplify possible instructions for collecting the DCAT implementations, which need to be revised.)

So, if we want to point to that page, we need to revise it. Another option is to send a mail to the public-comments mailing list or create a page in the WG wiki, and link to it (in this case, the page on the GH wiki is no longer necessary and should be deleted or marked as such).

riccardoAlbertoni (approved):

riccardoAlbertoni:
I have deleted the caveat and revised the page wiki so that we can keep the reference to this page.

@andrea-perego and other: Is this consistent with what you had in mind?

If yes I think we can merge this PR. 

> @davebrowning , @dr-shorthair , the current and old text points to https://github.com/w3c/dxwg/wiki/DCAT-implementation-evidence for instructions on how to report implementations, but the instructions are outlined at the end of the page, with also the following caveat:
> 
> > (Please ignore this section, the following instructions are not final, they are an attempt to exemplify possible instructions for collecting the DCAT implementations, which need to be revised.)
> 
> So, if we want to point to that page, we need to revise it. Another option is to send a mail to the public-comments mailing list or create a page in the WG wiki, and link to it (in this case, the page on the GH wiki is no longer necessary and should be deleted or marked as such).

Issue #1193, opened 2019-12-16
Rename DCATv2 RDF file dcat2.ttl


dr-shorthair:
DCATv2 uses the same namespace as DCATv1 (aka DCAT 2014). However, the RDF representations can be clarified through use of distinctive filenames. In our repo they are currently dcat2014.ttl and dcat.ttl, however, many users will likely have a copy of the original DCAT RDF stored as dcat.ttl, so there is potential confusion. 

I suggest it would be helpful for the new version to be in a file denoted `dcat2.ttl`. 

I think this will only require consequent changes to the 'namespace document' and not to the text of the recommendation.

dr-shorthair:
Fixed by https://github.com/w3c/dxwg/pull/1194

PR #1194, opened 2019-12-16
Change RDF filename to dcat2.ttl

dr-shorthair:
See #1193


pwin:
Makes sense, but in addition it might be good to have an additional "dcat.current.ttl" going forward

pwin (approved):
agreed, but also a 'dcat.current.ttl' might also be helpful

riccardoAlbertoni:
I have updated section 6.1 in the REC accordingly to Simon's changes,
So that, it points to dcat2.* instead of dcat.* files.

riccardoAlbertoni (approved):

PR #1229, opened 2020-03-30
Remove 'inherited' wrt properties of superclasses

### LINKED ISSUE #1228: Inheritance terminology not appropriate for RDFS and OWL
4 places in [the Recommendation](https://www.w3.org/TR/vocab-dcat-2/#Class:Catalog) refer to "_The following properties are inherited from the super-class_". The problem with this is that there is no concept of inheritance in the RDFS and OWL ecosystem.

From the OWL2 Primer:
> Furthermore, OWL is a strictly declarative and logical language. OWL therefore has none of the operational aspects of OOP, like methods, and similarly reasoning in OWL is strictly logical, with nothing comparing to inheritance, particularly inheritance with exceptions or overriding. - https://www.w3.org/2007/OWL/draft/ED-owl2-primer-20090414/

In my opinion, referring to Sub-Properties as Inheritance is mis-leading. The term is not defined in the normative references, and therefore leaves the statement ambiguous.

My personal experience implementing DCAT v2  is using OOP terminology causes implementers to incorrectly assume normal OOP semantics.

Please consider removing OOP terminology and replacing with RDFS and OWL terminology.

dr-shorthair:
fixes #1228


riccardoAlbertoni (approved):

agbeltran (commented):
The changes look good - were you planning to also add a reference (as per comment in the issue?)

dr-shorthair:
Unfortunately the reference is to a draft, which was superseded without the paragraph quoted. I've queried this https://github.com/w3c/dxwg/issues/1228#issuecomment-606392885

agbeltran (approved):
Thanks @dr-shorthair - approving these changes and the reference can be added later.

dr-shorthair:
The issue raised by @rawmatt probably also needs an addition to the note under the caption to Figure 1. Something like 

> 'Open arrow-heads indicate RDFS sub-classing (and not object-oriented inheritance).' 

I think that would do? i.e. no reference required, just direct people away from oo assumptions.

riccardoAlbertoni:
IMO, we can merge this pull and  label the related issues as "due for closing"

dr-shorthair:
Yes - I think good enough now. I can't merge as it is my pull.

Issue #1231, opened 2020-04-14
foaf:page, not foaf:homepage


rawmatt:
As an implementer, the [alignment with Schema.org](https://www.w3.org/TR/vocab-dcat-2/#dcat-sdo) is very useful. I use the [published alignment](https://w3c.github.io/dxwg/dcat/rdf/dcat-schema.ttl) with OWL-RL to infer my data catalog into it's SDO equivalents. This is useful for activities, such as driving [Google Dataset Search](https://datasetsearch.research.google.com).

I disagree with one mapping in the alignment, that for foaf:homepage.
```
foaf:homepage
  owl:equivalentProperty schema:url ;
.
```
The problem with using this relationship is that it is [symmetric](https://www.w3.org/TR/owl2-syntax/#Symmetric_Object_Properties). I agree that every foaf:homepage is also a schema:url, but the reverse is untrue. There are many schema:url that are not foaf:homepage.

schema:url is such a broadly defined property, far broader than foaf:homepage. These properties have very different breadth of applicability, and shouldn't be marked as equivalent.

The problem is that with this DCAT alignment, everywhere I have a schema:url, I am now seeing a foaf:homepage being inferred. And I have many schema:urls, and few of them are homepages.

The inference rules relevant here are prp-eqp1 and prp-eqp2.

The best mapping in the FOAF vocabulary to schema:url, is foaf:page. The definitions are very similar between schema:url and foaf:page. I propose changing the DCAT alignment to: 
```
foaf:page
  owl:equivalentProperty schema:url ;
.
```

This would achieve the desired result of mapping foaf:homepage to schema:url, but not the reverse. This is because [foaf:homepage](http://xmlns.com/foaf/spec/#term_homepage) is defined in FOAF as a sub-property of [foaf:page](http://xmlns.com/foaf/spec/#term_page).

DCAT would still contain foaf:homepage. The mapping to schema:url would now be through the super-property relationship from foaf:homepage to foaf:page, and then through the equivalence to schema:url.

dr-shorthair:
Good catch @rawmatt 
Note that the dcat-sdo alignment is non-normative, but good to know that you are finding it useful. 

The use of FOAF properties in DCAT at all is a bit of a legacy thing, coming from DCAT-2014. 
TBH my preference would probably be to replace all the FOAF predicates and classes with better ones from other namespaces, but we agreed to prioritize backward compatibility to protect the (mostly EU) implementations. 

Could you make a PR?

[...]

PR #1232, opened 2020-04-15
schema:url equivalent to foaf:page

rawmatt:
foaf:page is a closer match to schema:url, than foaf:homepage. Every foaf:homepage is a URL, but not the reverse. This fixes that, as foaf:homepage is a sub property of foaf:page. https://github.com/w3c/dxwg/issues/1231#issuecomment-613778775

### DIFF
--- a/dcat/rdf/dcat-schema.ttl
+++ b/dcat/rdf/dcat-schema.ttl
@@ -220,7 +220,7 @@ foaf:Organization
 foaf:Person
   owl:equivalentClass schema:Person ;
 .
-foaf:homepage
+foaf:page
   owl:equivalentProperty schema:url ;
 .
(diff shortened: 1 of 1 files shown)


dr-shorthair:
`foaf:homePage` is on `dcat:Catalog`. So the axiom `foaf:homePage rdfs:subClassOf foaf:page .` from FOAF is needed to complete the mapping. 

Do we assume that users will get that from the FOAF schema?

rawmatt:
Yes, I had picked up `foaf:homepage rdfs:subPropertyOf foaf:page`, from FOAF.

dr-shorthair (approved):

agbeltran (approved):

PR #1260, opened 2020-10-25
Remove domain & range indication in dct:hasPart

### LINKED ISSUE #1205: Issue with dct:hasPart as described in DCAT
I think there is possibly a set of issues with `dct:hasPart` as it is currently described in DCAT.

1. In the diagram and [in Catalog description](https://www.w3.org/TR/vocab-dcat-2/#Property:catalog_has_part) it has a domain (`dcat:Catalog`) and a range (`dcat:Resource`) defined.

As I [expressed elsewhere](https://github.com/w3c/dxwg/issues/725#issuecomment-513442046), I have a problem with DCAT imposing domains and ranges on external properties, as this is, in my opinion, basically saying "whoever uses `dct:hasPart` anywhere, from now on, the instances of subjects are (among other things) `dcat:Catalog`s and instances of objects are (among other things) `dcat:Resource`s.", which I think is undesirable.

However, even if I view this statement only "in the context of DCAT", this is still problematic due to a second issue:

2. In the usage note of [dcat:qualifiedRelation](https://www.w3.org/TR/vocab-dcat-2/#Property:resource_qualified_relation) it says:

> Used to link to another resource where the nature of the relationship is known but does not match one of the standard [DCTERMS] properties (dct:hasPart, ...

and in the usage note of the [Relationship class](https://www.w3.org/TR/vocab-dcat-2/#Class:Relationship) it says:

> Use to characterize a relationship between datasets, and potentially other resources, where the nature of the relationship is known but is not adequately characterized by the standard [DCTERMS] properties (dct:hasPart

From both usage notes I gather that if I want to model that one dataset is part of another dataset (or a dataset series), I would use `dct:hasPart`, not a qualified relation. This again has issues:

3. The work on dataset series was moved to DCAT 3 (#868 #806), however, the usage notes above suggest `dct:hasPart` should be used, not qualified relation, for "part of" relationship. This potentially interferes with the usage of qualified relation for dataset series in DCAT 3.
4. If `dct:hasPart` was used, even in scope of DCAT, my `dcat:Dataset` representing the series would also become `dcat:Catalog` thanks to the [domain definition](https://www.w3.org/TR/vocab-dcat-2/#Property:catalog_has_part), which I also see as an undesired side-effect, and potentially interfering with the usage of `dct:hasPart` for dataset series in DCAT 3 unless we say that a dataset representing a dataset series is in fact a catalog.

andrea-perego:
Relevant issue: https://github.com/w3c/dxwg/issues/1205

Preview:

https://raw.githack.com/w3c/dxwg/andrea-perego-dcat-fix-issue-1205/dcat/index.html#Property:catalog_has_part

Diff:

https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fandrea-perego-dcat-fix-issue-1205%2Fdcat%2Findex.html%23Property%3Acatalog_has_part#Property:catalog_has_part

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -837,7 +837,9 @@
                 <thead><tr><th>RDF Property:</th><th><a href="http://purl.org/dc/terms/hasPart">dct:hasPart</a></th></tr></thead>
                 <tbody>
                 <tr><td class="prop">Definition:</td><td>An item that is listed in the catalog.</td></tr>
+<!--                    
                 <tr><td class="prop">Domain:</td><td><a href="#Class:Catalog"><code>dcat:Catalog</code></a></td></tr>
+-->
                 <tr><td class="prop">Range:</td><td><a href="#Class:Resource"><code>dcat:Resource</code></a></td></tr>
                 <tr><td class="prop">Usage note:</td><td>This is the most general predicate for membership of a catalog. Use of a more specific sub-property is recommended when available. </td></tr>
                 <tr><td class="prop">See also:</td><td>Sub-properties of <code>dct:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr>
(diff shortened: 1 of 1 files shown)


riccardoAlbertoni (approved):

pwin (approved):
ok

PR #1262, opened 2020-10-25
Add draft section for dataset series

### LINKED ISSUE #806: Examples for dataset series
#81 was closed by #651. However, I still haven't found specific guidance for my two common dataset series cases:
1. Dataset of all addresses in the Czech Republic split into datasets for individual (6000+) towns. - This is a dataset series divided by geographical region. How do I specify a single entity (dataset?) with all the information?
2. Dataset with budget information per year. Each year is a separate dataset. How do I represent the entity (dataset?) with budget for all years?

Generally, how do I represent the series?
- n+first `dcat:Dataset` with no distributions connected to the multiple datasets using `dct:hasPart`? `dct:hasPart` for some reason is only specified in `dcat:Catalog`, not `dcat:Dataset` or `dcat:Resource`. I know that there is nothing prohibiting me from using `dct:hasPart` on `dcat:Dataset` for this, but I would wish that other people/implementations understood what exactly I mean by that.
- Should I use my own `dcat:Relationship` for that?

I think this is such a common use case that it needs to be explicitly dealt with. I was hoping #81 would deal with that, but the current resolution does not provide any guidance in this matter. Or am I missing something? Could we flesh out examples covering this?

### LINKED ISSUE #868: Dataset series
A significant outstanding issue that keeps coming up as the sideline to other conversations [1][2] is the need to have a recommended pattern for cataloguing dataset series. Budget data, satellite imagery, ... 

Usually most of the description (metadata) is the same, but the temporal coverage changes between members of a series. 

[1] https://github.com/w3c/dxwg/pull/789
[2] https://github.com/w3c/dxwg/issues/806

### LINKED ISSUE #1205: Issue with dct:hasPart as described in DCAT

andrea-perego:
Relevant issues: https://github.com/w3c/dxwg/issues?q=is%3Aissue+is%3Aopen+label%3Adataset-series

Preview: https://raw.githack.com/w3c/dxwg/dcat-dataset-series/dcat/index.html#dataset-series


dcat/index.html
```diff
@@ -3387,6 +3387,225 @@ <h2>Resource life-cycle</h2>
...
+
+<p>With "dataset series" we refer to data, somehow interrelated, that are published separately, although they could be  merged into a single dataset. An example is budget data split by year and/or country, instead of being made available in a single dataset.</p>
+
+<p>Dataset series are defined in [[ISO-19115]] as a <q>collection of datasets [&hellip;] sharing common characteristics</q>. However, their use is not limited to geospatial data, although in other domains they can be named differently (e.g., time series, data slices) and defined more or less strictly (see, e.g., the notion of "dataset slice" in [[VOCAB-DATA-CUBE]]).</p>
+
+<p>The reasons and criteria for splitting data into series are manyfold, and they may be related to, e.g., data characteristics, publishing process, and how they are typically used. For instance, data huge in size (as geospatial ones) are more easily handled (by data providers as well as data consumers) by splitting them in smaller ones. Another example is data that are released on a yearly basis, which are typically published as separate datasets, instead of appending the new data to the first in the series.</p>
```
riccardoAlbertoni: @andrea-perego: perhaps I would revise it as in the following, but your mastery of English is sure better than mine ;)
```suggestion
<p>The reasons and criteria for splitting data into series are manyfold, and they may be related to, e.g., data characteristics, publishing process, and how they are typically used. For instance, data huge in size (as geospatial ones) are more easily handled (by data providers as well as data consumers) by splitting them into smaller ones. Another example is data released on a yearly basis, which are typically published as separate datasets, instead of appending the new data to the first in the series.</p>
```

dcat/index.html
```diff
@@ -3387,6 +3387,225 @@ <h2>Resource life-cycle</h2>
...
+
+<p>Compared with the second option, the first one may have the advantage of simplifying metadata management, and avoiding the creation of datasets having an almost equivalent intensional description (i.e., datasets having the same values for almost all their metadata elements). On the other hand, this approach reduces the ability of being discovered, as distribution metadata are not rich as datasets' ones. Moreover, using distributions may result cumbersome or unfeasible when the number of child datasets is too high.</p>
+
+<p>As stated in the note to <a href="#Class:Distribution"></a>, DCAT recommends, as the default approach, typing both dataset series and child datasets as <code>dct:Dataset</code>'s, and linking them by using properties <code>dct:hasPart</code> and/or <code>dct:isPartOf</code>, and possibly soft-typing the dataset series via property <code>dct:type</code>.</p>
+
+<aside class="ednote">
+<p>The creation of a specific class for dataset series is under discussion.</p>
```
riccardoAlbertoni, line 3427: I wonder if we want to have a new dedicated issue on the creation of a dedicated class for dataset series. Maybe the issue can refer to discussions that happened as part of Issue 868, e.g., comment https://github.com/w3c/dxwg/issues/868#issuecomment-518337457 ect.
What do you think?

dcat/index.html
```diff
@@ -3387,6 +3387,225 @@ <h2>Resource life-cycle</h2>
...
+</aside>
+
+<p>The approach based on the use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</p>
```
riccardoAlbertoni, line 3430: I wonder if we should drop this line..  Having two ways of representing the same might hamper interoperability. As far as I understood, there are no many actual implementations using the "distribution" alternative .. and as pointed out by Makx this somehow breaks the  DCAT backbone.  I think it is correct to mention both, but we should try to not encourage using the dcat:distribution in this way.

riccardoAlbertoni: I acknowledge that the DBPEDIA case mentioned in the discussion might be one case, if not dropping the entire line we can rephrase
```suggestion
<p>The approach based on the use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, only if it addresses more effectively the requirements of a given application scenario.</p>
```

dcat/index.html
andrea-perego: Makes sense to me. But maybe we'd better wait until we have this PR merged, so that people can have a tangible proposal to review.

dcat/index.html
andrea-perego: Actually, to my knowledge, distributions is how people are addressing this - also because of lack of support in catalogue platforms for dataset series. 

About replacing "whenever" with "only if": I was also considering that wording, but it might be too strong - also because people may be obliged to opt for distributions in case their catalogue does not support dataset series. I would suggest we discuss this during one of the DCAT calls, and keep "whenever" for the moment.

riccardoAlbertoni: Fine to me, let's keep as it is for the moment.

riccardoAlbertoni (approved):

pwin:
"equivalent intensional description" -  I think this is meant to be "intentional" @andrea-perego  [ https://raw.githack.com/w3c/dxwg/dcat-dataset-series/dcat/index.html#dataset-series-specification para 2].  Either way, it perhaps needs to be worded in a slightly different way.

dcat/index.html
```diff
@@ -3387,6 +3387,233 @@ <h2>Resource life-cycle</h2>
...
+
+<ol>
+<li>The dataset series is typed as a <code>dcat:Dataset</code>, whereas its child datasets are typed as <code>dcat:Distribution</code>'s.</li>
+<li>Both the dataset series and its child datasets are typed as a <code>dcat:Dataset</code>'s, and the two are usually linked by using the [[DCTERMS]] properties <code>dct:hasPart</code> / <code>dct:isPartOf</code>.</li>
+</ol>
+
+<p>In both cases, the dataset series is sometimes soft-typed by using the [[DCTERMS]] property <code>dct:type</code> (e.g., this is the approach used in [[GeoDCAT-AP]], and adopted in [[DCAT-AP-IT]] and [[GeoDCAT-AP-IT]]).</p>
+
+<p>Compared with the second option, the first one may have the advantage of simplifying metadata management, and avoiding the creation of datasets having an almost equivalent intensional description (i.e., datasets having the same values for almost all their metadata elements). On the other hand, this approach reduces the ability of being discovered, as distribution metadata are not rich as datasets' ones. Moreover, using distributions may result cumbersome or unfeasible when the number of child datasets is too high.</p>
```
```suggestion
<p>Compared with the second option, the first one may have the advantage of simplifying metadata management, and avoiding the creation of datasets having the same values for almost all their metadata elements. On the other hand, this approach reduces the ability of being discovered, as distribution metadata are not rich as datasets' ones. Moreover, using distributions may result cumbersome or unfeasible when the number of child datasets is too high.</p>
```

andrea-perego:
> "equivalent intensional description" - I think this is meant to be "intentional" @andrea-perego [ https://raw.githack.com/w3c/dxwg/dcat-dataset-series/dcat/index.html#dataset-series-specification para 2]. Either way, it perhaps needs to be worded in a slightly different way.

Thanks, @pwin . I revised it via https://github.com/w3c/dxwg/pull/1262/commits/342bb013cdc9b2310f0579e6bd67cce7e0ba5f6e

riccardoAlbertoni (approved):

pwin (approved):
looks good

PR #1315, opened 2021-03-09
Replace dct: with dcterms:

andrea-perego:
See https://github.com/w3c/dxwg/issues/1314

Preview: https://raw.githack.com/w3c/dxwg/dcat-issue-1314/dcat/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-issue-1314%2Fdcat%2Findex.html

PR #1316, opened 2021-03-09
Replacing dct with dcterms in rdf and creating rdf files for DCAT 3

### LINKED ISSUE #1314: Namespace prefix for DC Terms
In our documentation we are using the prefix `dct:` to denote the Dublin Core Terms namespace, and `dcat:` for DCAT. 
I have found that users often conflate these, probably because the prefixes are so similar.  
I suggest using the prefix `dcterms:` instead.

riccardoAlbertoni:
This  PR  creates the rdf files for dcat3 and turns dct curie into dcterms. 

Usually, we keep the RDFs using the same namespaces as in the HTML. 
Besides, similarly to what we already did when moving from dcat2014 to dcat2, I think it is time to create specific rdf files for dcat3.


andrea-perego:
> This PR creates the rdf files for dcat3 and turns dct curie into dcterms.
> 
> Usually, we keep the RDFs using the same namespaces as in the HTML.
> Besides, similarly to what we already did when moving from dcat2014 to dcat2, I think it is time to create specific rdf files for dcat3.

I agree.

I think we need also to update the files with the examples.

agbeltran:
We should also update the figure - until we resolve other method, is it easy to regenerate the figure with the new prefixes @dr-shorthair ?

riccardoAlbertoni:
> We should also update the figure - until we resolve other method, is it easy to regenerate the figure with the new prefixes @dr-shorthair ?

I suspect we will change figure 1 more than one time in the next month. PRs on versioning and dataset series are likely to change the figure as well. Perhaps we'd better open a new GitHub issue titled "changes to figure 1"  where we pile the changes on to make them on the figure all at once.  What do you think?

agbeltran:
yes, good point @riccardoAlbertoni ! thanks

riccardoAlbertoni:
@andrea-perego and all: Apparently, DCAT 2 REC points to the GitHub examples directly, so if we change dct in there, we will lose the coherence with the  DCAT 2 REC.

I see three options, 1) we wait for changing the examples; 2) we create a brand new copy of the directory 'examples/' named 'exampleDCAT3',  in which dct is replaced by dcterms and we keep the pointed files in 'examples/' coherent with DCAT 2 REC; 3) we change the namespace and lose the coherence of examples with the current REC, as in the end, having dct in the rec and dcterms in the examples is not a big issue.

 Any preferences?

andrea-perego:
Let's wait for the moment. The issue raised by @dr-shorthair is more on the spec side than in the files of the examples and the RDF.

andrea-perego:
As PR https://github.com/w3c/dxwg/pull/1315 is ready to be merged, I'd merge this one into it.

@riccardoAlbertoni , can I proceed, or do you have other revisions to be added?

andrea-perego (approved):

riccardoAlbertoni:
yes, we can merge this into the other PR, we will deal with the examples later, following proposal 1.

Issue #1307, opened 2021-02-28
Should we specialize dcterms:hasPart for dcat:DatasetSeries?


riccardoAlbertoni:
For data catalog, `dcterms:hasPart` is specialized into `dcat:dataset`, `dcat:service`, etc. Should we have some sort of specialization also for `dcterms:hasPart` used in `dcat:DatasetSeries`? 

In particular, to distinguish more between datasets that are 'just a bag of files' (see examples in [Loosely structured catalog](https://www.w3.org/TR/vocab-dcat-3/#examples-bag-of-files)) from actual datasets/distributions part of the dataset series.

riannella:
Can we just use `dcat:dataset` as property from `dcat:DatasetSeries` and `dcat:Dataset` ?

(of course, I would like to see the inverse property as well  :-)

dr-shorthair:
My suggestion: 

```
dcat:hasSeriesMember a owl:ObjectProperty ;
    rdfs:subPropertyOf dcterms:hasPart ;
    rdfs:domain dcat:DatasetSeries ;
    rdfs:range dcat:Dataset ;
. 

dcat:inSeries a owl:ObjectProperty ;
    rdfs:subPropertyOf dcterms:isPartOf ;
    rdfs:domain dcat:Dataset ;
    rdfs:range dcat:DatasetSeries ;
. 
```

jakubklimek:
@dr-shorthair @riannella From the point of view of an application developer, let me again state that I would strongly prefer to standardize only one of the two mutually inverse properties.

Otherwise, compliant applications and queries need to always account for the possibility that in one data catalog, the datasets are connected to the series using one property, in another using the second, which makes them more complex with each inverse property pair.

agreiner:
I would suggest keeping only the inSeries property, so that one needn't worry about updating metadata for the series every time a new dataset is added to that series.

dr-shorthair:
Agree with @jakubklimek and @agreiner . Only one needed and `dcat:inSeries` is more useful.

riannella:
...but they are inverse properties...so you only need to define one, the other can be asserted or inferred.

dr-shorthair:
Yes. Agree.

andrea-perego:
I'm fine with defining only `dcat:inSeries` for the moment.

riannella:
Sorry @andrea-perego - when I said "you only need to define one" - I really meant you only need to "assert" one.
Both properties should be defined (in the ontology).

[...]

PR #1292, opened 2021-02-09
Dataset series  v2  (issue 1272)

riccardoAlbertoni:
Draft including dcat:DatasetSeries as a new class as discussed in #1272.

- Adding  dcat:DatasetSeries class description see https://raw.githack.com/w3c/dxwg/dcat-dataseries-issue1272/dcat/index.html#Class:Dataset_Series
- Revising guidelines about dataset series  see https://raw.githack.com/w3c/dxwg/dcat-dataseries-issue1272/dcat/index.html#dataset-series
- Specific editorial questions are included to track doubts and editorial choices that we might want to discuss.

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-dataseries-issue1272%2Fdcat%2Findex.html

TODO:  (once we have agreed on usage notes and definitions)
- update RDF
- update change history

PR #1328, opened 2021-03-14
DatasetSeries: adding dcat:inSeries and its inverse, updating the  dataset examples. 

riccardoAlbertoni:
Updating the PR #1292 on dataset series to include the result of the discussion we had in issue #1307 

This  PR replaces `dcterms:isPartOf`  and its inverse `dcterms:hasPart` with `dcat:inSeries` and its inverse `dcat:hasSeriesMember`.

Some people insist on keeping both directions,  and the examples in #1292 are using both `dcterms:isPartOf`  and its inverse `dcterms:hasPart`,  so for the moment,  the PR maintains both directions. 

DIFF with PR #1292 https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fdxwg%2Fdcat-dataseries-issue1272%2Fdcat%2Findex.html&doc2=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fdxwg%2Fdcat-dataseries-issue1272UpdatedWithDiscussionOn1307%2Fdcat%2Findex.html

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -4438,6 +4438,10 @@ mycity-bus:stops-2015-01-01
 <aside class="example" id="ex-dataset-series-containment" title="Yearly budget datasets grouped into a series"> 
 <p>In the following example, yearly budget data are grouped into a series. The series is typed as <code>dcat:DatasetSeries</code>, the child datasets are typed as <code>dcat:Dataset</code>. The series and the datasets are linked by using both <code>dcat:hasSeriesMember</code> and <code>dcat:inSeries</code>.</p>
 <pre>
+ex:EUCatalogue a dcat:Catalog ;
+  dcterms:title "European Data Catalogue"@en ;
+  dcat:dataset ex:budget , ex:employment , ex:finance ;
+  .
 
 ex:budget a dcat:DatasetSeries ;
   dcterms:title "Budget data"@en ;
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2543,6 +2543,9 @@
 <!--			
            <div class="issue" data-number="1307"> </div>
 -->
+           <aside class="note">
+               <p>Property added in DCAT 3.</p>
+           </aside>
             <table class="definition">
                 <thead><tr><th>RDF Property:</th><th><a href="http://www.w3.org/ns/dcat#DatasetSeries">dcat:hasSeriesMember</a></th></tr></thead>
                 <tbody>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -4432,7 +4432,7 @@ mycity-bus:stops-2015-01-01
 
 <h2>How to specify dataset series</h2>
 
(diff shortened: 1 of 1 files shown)


andrea-perego:
For the inverse of `dcat:inSeries` I suggest we use something shorter: `dcat:hasMember`.

Another option for the names of these properties is `dcat:hasParent` / `dcat:hasChild` , following the terminology used in ISO 19115 for dataset series.

dr-shorthair:
Need to add `inSeries` to the list of properties added to `dcat:Dataset` (in the intro section)

dr-shorthair (changes requested):
Need to add `inSeries` to the list of properties added to `dcat:Dataset` (in the intro text on Dataset section)

riccardoAlbertoni:
> Need to add `inSeries` to the list of properties added to `dcat:Dataset` (in the intro text on Dataset section)

Thanks,  done.

dr-shorthair (approved):

riccardoAlbertoni:
> For the inverse of `dcat:inSeries`, I suggest we use something shorter: `dcat:hasMember.`
> 
> Another option for the names of these properties is `dcat:hasParent` / `dcat:hasChild` , following the terminology used in ISO 19115 for dataset series.

@andrea-perego:
Yes, the names are long. However, I wonder if mentioning 'series' in the property names can help recall and use these properties more coherently. This also considering possible confusion with `dcterms:hasPart` and other parts-related properties.  What do you think?

andrea-perego:
> @andrea-perego:
> Yes, the names are long. However, I wonder if mentioning 'series' in the property names can help recall and use these properties more coherently. This also considering possible confusion with `dcterms:hasPart` and other parts-related properties. What do you think?

I don't think it's necessary to recall the scope of this property, as it has its own domain (`dcat:DatasetSeries`). For the same reason, including "series" in the property name looks redundant, and I am not very worried about being confused with `dcterms:hasPart`.

However, if the close semantics of "part" and "series" raises concerns on improper use, we should consider other options.

`dcat:hasChild` / `dcat:hasParent` are a good alternative, IMO, and they reflect the hierarchical structure of dataset series, as already mentioned in PR https://github.com/w3c/dxwg/pull/1292 and in https://github.com/w3c/dxwg/issues/1307#issuecomment-799139402

riccardoAlbertoni:
> > @andrea-perego:
> > Yes, the names are long. However, I wonder if mentioning 'series' in the property names can help recall and use these properties more coherently. This also considering possible confusion with `dcterms:hasPart` and other parts-related properties. What do you think?
> 
> I don't think it's necessary to recall the scope of this property, as it has its own domain (`dcat:DatasetSeries`). For the same reason, including "series" in the property name looks redundant, and I am not very worried about being confused with `dcterms:hasPart`.
> 
> However, if the close semantics of "part" and "series" raises concerns on improper use, we should consider other options.
> 
> `dcat:hasChild` / `dcat:hasParent` are a good alternative, IMO, and they reflect the hierarchical structure of dataset series, as already mentioned in PR #1292 and in [#1307 (comment)](https://github.com/w3c/dxwg/issues/1307#issuecomment-799139402)

If you think `dcat:hasSeriesMember` is too long and not acceptable then `dcat:hasMember` will be my choice.

Using `dcat:hasChild` and `dcat:hasParent` helps those familiar with ISO dataset series.  However, we target a larger audience. In this target, I suspect people might have different views of what is considered a child of a dataset (potentially, anything originated by a dataset can be considered a 'child'!? and that might be confusing).

riccardoAlbertoni:
I am going to merge this with dcat-dataseries-issue1272  PR #1292  so that we have only one updated version for the dataset series as having two seems confusing commenters.  We still need to decide the property name between series and dataset (i.e., if we used dcat:hasSeriesMember or dcat:hasMember), but it is an adjustment we can do later, in PR #1292.

Issue #1246, opened 2020-08-12
DCAT2.ttl: Request for minor changes


aidig:
1) Regarding the labels of properties such as dcat:dataset, dcat:distribution og dcat:catalog, it would be beneficial to prefer labels which highlight that the fact that these elements are properties and not classes by preferring labels such as “has dataset”, "has service", “has catalog” or “has distribution”. Currently, they differ only by way of capitalization. (also in normative spec)

2) Regarding the definition of dcat:Catalog, it would be more appropriate to move the examples to a comment or scopeNote. (also in normative spec)

3) The links provided in the scopeNote of dcat:Role and dcat:hadRole are unstable test addresses:  http://registry.it.csiro.au/def/isotc211/CI_RoleCode.& http://registry.it.csiro.au/def/isotc211/DS_AssociationTypeCode; The following is considered the correct url: https://def.isotc211.org/ontologies/iso19115/-1/2018/CitationAndResponsiblePartyInformation.rdf 

4) Regarding property dcat:spatialResolutionInMeters, it would be beneficial to add the following information from the DCAT spec (from bbox) in the TTL section describing dcat:spatialResolutionsInMeters or - ideally - replacing the current second scopeNote with this text: “The ability of to detect details in an raster image or grid is referred to as spatial resolution. The spatial resolution is stated in metres. The more pixels are included in a raster image of a certain area, the higher the spatial resolution meaning the more details can be observed”. 
(source: http://fis.uni-bonn.de/en/recherchetools/infobox/professionals/resolution/spatial-resolution)

andrea-perego:
Dear @aidig ,

Many apologies for replying so late.

You points 1 & 2 are being addressed in PR https://github.com/w3c/dxwg/pull/1337

About point 3, this issue is under discussion in https://github.com/w3c/dxwg/issues/975 , and we have yet to take a decision.

About point 4:

> 4. Regarding property dcat:spatialResolutionInMeters, it would be beneficial to add the following information from the DCAT spec (from bbox) in the TTL section describing dcat:spatialResolutionsInMeters or - ideally - replacing the current second scopeNote with this text: “The ability of to detect details in an raster image or grid is referred to as spatial resolution. The spatial resolution is stated in metres. The more pixels are included in a raster image of a certain area, the higher the spatial resolution meaning the more details can be observed”.
(source: http://fis.uni-bonn.de/en/recherchetools/infobox/professionals/resolution/spatial-resolution)

I'm not sure about this sentence:

[...]

PR #1337, opened 2021-03-25
Old attempt to solve  1246 - points 1 and 2

riccardoAlbertoni:
This PR addresses the first two points  of the issue https://github.com/w3c/dxwg/issues/1246

As for point 1 of the issue,
this PR changes `rdfs:label` for object properties named after the class they range to but with an uncapitalized starting letter. 
 
Relations labels whose  refs:label changed  from “ X “  to “has X” are  
- dcat:dataset  
- dcat:service 
- dcat:catalog 
- dcat:distribution 

Labels for the other properties are not changed as they do not ‘clash' with their range classes.

As for point 2, examples of resources in catalog definition are moved in usage note.

**Still to discuss**
Translations of labels for languages different from English are not provided, I wonder if we want to guess them via automatic translation services, leave them as they are, or ask native-speaking people to provide translations.

We will add new stuff in DCAT 3, so I would suggest planning language revisions of the translation when delivering the last version of DCAT3.
Perhaps we need a separate issue to track translations? 

**TODO (once we have agreed on the changes)**
-To Update other RDF serializations

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2522,6 +2522,7 @@
             <a href="#Property:resource_type">type/genre</a>,
             <a href="#Property:resource_update_date">update/modification date</a>,
             <a href="#Property:resource_qualified_attribution">qualified attribution</a>,
+            <a href="#Property:dataset_distribution">has distribution</a>,
             <a href="#Property:dataset_frequency">frequency</a>,
             <a href="#Property:dataset_spatial">spatial/geographic coverage</a>,
             <a href="#Property:dataset_spatial_resolution">spatial resolution</a>,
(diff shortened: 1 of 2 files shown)


agbeltran:
@riccardoAlbertoni as per my comments on the issue, I don't think we should change the labels on the spec, but it would be useful to add ```skos:altLabel``` in the RDF serialisations

riccardoAlbertoni:
> @riccardoAlbertoni as per my comments on the issue, I don't think we should change the labels on the spec, but it would be useful to add `skos:altLabel` in the RDF serializations

Thanks, Alejandra, 
That could be an alternative, let's see how the discussion on  https://github.com/w3c/dxwg/issues/1246#issuecomment-808413945 evolves.

riccardoAlbertoni:
This PR has been superseded by https://github.com/w3c/dxwg/pull/1348, it is kept just as a reference for past discussions, and it will be deleted as soon as we agree on #1348 and we close the related issues.

agbeltran:
#1348 merged, but #1246 will remain open until we address the two final points - this PR should be closed IMO

andrea-perego:
> #1348 merged, but #1246 will remain open until we address the two final points - this PR should be closed IMO

+1 to close it.

PR #1340, opened 2021-03-27
Issue #1308

### LINKED ISSUE #1308: Any other properties to mention in class dcat:DatasetSeries?
~~Should we add  <code>adms:next</code>, <code>adms:prev</code> in the <code>dcat:DatasetSeries</code> class description?~~
 
Should we include in the <code>dcat:DatasetSeries</code> class description the properties whose values are involved in [dataset series inheritance](https://www.w3.org/TR/vocab-dcat-3/#dataset-series-properties) (e.g. <code>dct:issued</code>,<code>dct:spatial</code>)? 
The expected constraints can be explained in their usage notes...

andrea-perego:
Relevant issue: https://github.com/w3c/dxwg/issues/1308

Summary of changes:
- Add `dcat:first`, `dcat:last`, `dcat:next`, `dcat:prev` to `dcat:Resource`
- Update list of inherited properties for subclasses of `dcat:Resource`
- Add `xhv` to the table with informative namespaces
- Add NOTE after the list of inherited properties of `dcat:DatasetSeries`, to say that some of the inherited properties may have particular semantics when used in this context
- Editorial fixes

Preview: https://raw.githack.com/w3c/dxwg/dcat-issue-1308/dcat/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-issue-1308%2Fdcat%2Findex.html

### DIFF
--- a/dcat/config.js
+++ b/dcat/config.js
@@ -122,8 +122,8 @@ var respecConfig = {
             title:"DBPedia ontology"
         },
         "DCAT-AP-SE": {
-            title: "DCAT-AP-SE: Swedish recommendation for DCAT-AP1.1",
-            href: "https://lankadedata.se/spec/DCAT-AP-SE",
+            title: "DCAT-AP-SE: Clarifications, translations and explanations of DCAT-AP for Sweden",
+            href: "https://docs.dataportal.se/dcat/en/",
             authors: ["Matthias Palmér"],
             etAl: false
         },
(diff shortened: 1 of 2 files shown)


rob-metalinkage:
Do you really want to pull in terms into DCAT as a W3C Rec from a W3C Note that itself declares it is a profile of DCAT? 

Understand you need to use these, and not want to re-invent the wheel (and my own implementations do exactly the same) - but the logically consistent approach would be to define a profile of DCAT that uses the parts of ADMS you need, as a profile of ADMS itself probably, that itself has the same status (Note) of ADMS.  You then guide people to this rather than add it to the normative Rec status DCAT specification.   It appears that a decision about not publishing profiles of DCAT is becoming an arbitrary limitation that leads to anti-patterns where profiling would provide a better option.

andrea-perego:
@rob-metalinkage , I suggest you check the status of the related discussion in issue https://github.com/w3c/dxwg/issues/1308

rob-metalinkage:
@andrea-perego ok - i have copied comments in there, as that issue has clearly not addressed this points yet.

riccardoAlbertoni (approved):

davebrowning (approved):
As discussed, there will likely need to be some further discussion but the updates in the PR seem fine to merge now

PR #1348, opened 2021-04-08
new attempt to solve 1246 ( adding altLabel instead of changing rdfs:label)

riccardoAlbertoni:
This PR addresses the first two points of the issue https://github.com/w3c/dxwg/issues/1246, coherently with the resolution taken in https://www.w3.org/2021/03/31-dxwgdcat-minutes#r03. 
It replaces the previous attempt made in PR #1337.

As for point 1 of the issue,
This PR adds `skos:altLabel` for object properties named after the class they range to but with an uncapitalized starting letter. 
 
we have added  `skos:altLabel`   “has X”  where X is one of  the following properies  
- dcat:dataset  
- dcat:service 
- dcat:catalog 
- dcat:distribution 

Labels for the other properties are not changed as they do not ‘clash' with their range classes.

As for point 2, we have moved the examples of resources from the definition of `dcat:Catalog` to a new usage note. This change is done both in the RDF and the DCAT document.

All the   RDF serializations are updated.

**Still to discuss**
Translations of labels for languages different from English are not provided, I wonder if we want to guess them via automatic translation services, leave them as they are, or ask native-speaking people to provide translations.

We will add new stuff in DCAT 3, so I would suggest planning language revisions of the translation when delivering the last version of DCAT3.
Perhaps we need a separate issue to track translations?


andrea-perego (approved):

andrea-perego:
Thanks for taking care of these revisions, @riccardoAlbertoni .

About this point:

> **Still to discuss**
> Translations of labels for languages different from English are not provided, I wonder if we want to guess them via automatic translation services, leave them as they are, or ask native-speaking people to provide translations.
> 
> We will add new stuff in DCAT 3, so I would suggest planning language revisions of the translation when delivering the last version of DCAT3.
> Perhaps we need a separate issue to track translations?

I think we'd better create a separate issue.

riccardoAlbertoni:
> I think we'd better create a separate issue.

done,   see https://github.com/w3c/dxwg/issues/1352

agbeltran (approved):

Issue #1155, opened 2019-11-05
What kind of dataset is dataset-001?


aisaac:
Are the small examples given in the DCAT spec supposed to be building up a more complete picture of a bigger example?

If yes, then I'm puzzled by the story told for dataset-001. In example 6 its primary type says it is a dataset of manuscripts. But its content type in example 7 is "standard office documents", which hints these were not created purely by hand. And its temporal coverage elsewhere says it covers 2006. It's unlikely manuscripts would be created as a main source of data at or after that date. And it has a distribution as csv files, later.

I reckon that it is meant to be an imaginary dataset. Still we could aim at presenting slightly more consistent chimeras :-)

This is probably for future work anyway.

riccardoAlbertoni:
Thanks for this editorial suggestion. Though it is not substantial, I agree that we could aim at presenting slightly more consistent chimaeras. So,  as a reminder, I have labelled this issue as future work.

riccardoAlbertoni:
@aisaac, please take a look at PR #1361 , it changes the examples considering your remarks.

riccardoAlbertoni:
> @aisaac, please take a look at PR #1361 , it changes the examples considering your remarks.

@aisaac: PR 1361 has been merged, if there are no objections, I think we can close this issue...

aisaac:
@riccardoAlbertoni sorry for the delay. It doesn't look so inconsistent anymore, so this can be closed!

PR #1361, opened 2021-05-05
Dcat issue 1155

riccardoAlbertoni:
An attempt to make the story of dataset-001 more consistent considering the remarks made by @aisaac in issue #1155.

- the temporal coverage is closed to the issue date making the whole publication story more credible.
- dataset-001 was typed as manuscript now is a dataset/database

Preview: https://raw.githack.com/w3c/dxwg/dcat-issue-1155/dcat/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-issue-1155%2Fdcat%2Findex.html


dcat/index.html
```diff
@@ -436,7 +436,7 @@ <h3>Basic example</h3>
   dcterms:issued "2011-12-05"^^xsd:date&nbsp;;
   dcterms:modified "2011-12-15"^^xsd:date&nbsp;;
   dcat:contactPoint &lt;http://example.org/transparency-office/contact&gt; ;
-  dcterms:temporal &lt;http://reference.data.gov.uk/id/quarter/2006-Q1&gt; ;
+  dcterms:temporal &lt;http://reference.data.gov.uk/id/quarter/2011-Q3&gt; ;
```
andrea-perego: I wonder whether we should consider replacing the specification of the temporal coverage with the one supported in DCAT (`dcterms:Period` + `dcat:startDate` + `dcat:endDate`), and included in the other examples.

riccardoAlbertoni: Example 3 is one of the first examples, I guess it has a lot of attention from "casual distracted readers". For that reason, we might want to stress the existence of the new DCAT 2 properties. At the same time,  I find the reuse of data.gov.uk quite useful.

What about leaving the example as it is, and adding a reference to the temporal properties section (https://www.w3.org/TR/2021/WD-vocab-dcat-3-20210504/#temporal-properties) in the explanation which follows Example 3?

The use of  (`dcterms:Period` + `dcat:startDate` + `dcat:endDate`) is very well visible there, so keeping the example as it is can be beneficial for giving the visibility of the whole range of options. 

For example, we could  add the sentence in bold

> The temporal coverage or extent is given in dcterms:temporal using an item from the Interval dataset (originally available from http://reference.data.gov.uk/id/interval) from data.gov.uk. **Other options to express dcterms: temporal are discussed  in `<a href="#temporal-properties"></a>`** .

dr-shorthair: +1

andrea-perego: I see your point, @riccardoAlbertoni , but I think there are a few issues here:

1. The example was included when DCAT didn't have properties for specifying the start / end of a period, and when the time URIs from reference.data.gov.uk were released and actively used. 
2. DCAT has now 2 pairs of specific properties, which are meant to be the preferred options to address the most common (`dcat:startDate` / `dcat:endDate`) and more complex (`time:hasBeginning` / `time:hasEnd`) use cases. If we are referring to the reference.data.gov time URIs as another option, it should be the third in cascade, but we should also motivate when it should be used.
3. I'm not sure if the reference.data.gov time URIs are actually still used, so I am not comfortable in suggesting it as a possible option.

For the reasons above, I still think the example should be revised, but it can be complemented with a note about the use of the reference.data.gov time URIs in DCAT 2014, saying that now DCAT addresses this with specific properties.

riccardoAlbertoni: Hi @andrea-perego, I have updated the example as you suggested. Please double-check if  I have correctly understood your proposal.

As for the note about reference.data.gov, perhaps I would avoid it if we are not considering this as an option.  What do you think?

andrea-perego: > Hi @andrea-perego, I have updated the example as you suggested. Please double-check if I have correctly understood your proposal.

+1 from me. Thanks, @riccardoAlbertoni .

> As for the note about reference.data.gov, perhaps I would avoid it if we are not considering this as an option. What do you think?

I agree. But maybe we should add an entry in the changelog.

riccardoAlbertoni: Added the entry in the changelog

andrea-perego: Thanks, @riccardoAlbertoni . I'm happy with the changes.

andrea-perego (approved):

davebrowning (approved):

PR #1379, opened 2021-06-07
Dcat issue 1364 second solution

riccardoAlbertoni:
An alternative solution to PR #1380 

This  PR drops the range for `dcat:theme`, but explicitly defining it as `owl:ObjectProperty` as suggested by @dr-shorthair  in https://github.com/w3c/dxwg/issues/1364#issuecomment-843616224

Considering that `dcterm:subject` has been recently changed to let both literals and  URI,  by defining `dcat:theme`s as an `owl:ObjectProperty`  and dropping the `dcat:theme`'s range,  `dcat:theme` would still restrict `dcterms:subject` to URIs,  while `dcat:keyword` might restrict `dcterms:subject` to literals.

This would also limit the automatic inference of `skos:Concept`, which could be potentially problematic when applied to instances whose classes are defined disjoint from `skos:Concept`.

Issue #1364, opened 2021-05-18
Relax range restriction on dcat:theme


andrea-perego:
@dr-shorthair 

Why not relax range constraint for `dcat:theme`? In the biomedical domain we have more ontologies and often these ontologies have `owl:Class` typing for concepts.

_Originally posted by @rajaram5 in https://github.com/w3c/dxwg/issues/123#issuecomment-843290546_

agbeltran:
There was also a discussion around the range for ```dcat:theme``` at: https://github.com/w3c/dxwg/issues/1153

dr-shorthair:
I would be comfortable relaxing the ranges of theme and themeTaxonomy. But it is important to keep them as object-properties, since the object must be denoted by a URI - i.e. taken from a managed source somewhere.

makxdekkers:
All, there was an earlier discussion on this (see [here](https://github.com/w3c/dxwg/issues/1153#issuecomment-549750210)), where I noted that `dcat:theme` was explicitly minted for DCAT version 1 because it was felt that the range of `dct:subject` was too broad and that it would be useful to have a property with a range restricted to `skos:Concept`. Otherwise, `dct:subject` could have been used.
Now, with the relaxation of the range of `dct:subject`, which now also allows for textual values, I guess that `dcat:theme` with relaxed range, but still requiring it to be an object property, can still be seen as a restriction on `dct:subject`, effectively making it the same as the 'old' (pre-relaxation) `dct:subject`. Right?

rajaram5:
@dr-shorthair yes, it could work for our use cases if `dcat:theme` is an `object property`

rajaram5:
@makxdekkers I see your point. As I mentioned early in biomedical domain we have more `owl` based ontologies and vocabularies. So if you still like to restrict `range` of `dcat:theme` property it would be nice to have list of values like `skos:Concept`, `owl:Class`.

Issue #1153, opened 2019-11-04
Inconsistency between usage note of dcat:themeTaxonomy and range of dcat:theme?


aisaac:
The usage note of dcat:themeTaxonomy says "It is recommended that the taxonomy is organized in a skos:ConceptScheme, skos:Collection, owl:Ontology or similar, which allows each member to be denoted by an IRI and published as Linked Data.". 
But the range of dcat:theme is skos:Concept, so it seems that the expectation for the members of the objects of dcat:themeTaxonomy statements is a bit more precise than what the usage note says.

Future work?

dr-shorthair:
Concepts are not always organized into ConceptSchemes, Collections or Ontologies. Sometimes they are free-floating (e.g. specified at run time).

nicholascar:
@aisaac  is correct that the range gives an ***expectation*** of use, even if in general Semantic Web work you don't *have* to use `skos:Concept`s within a `skos:ConceptScheme` or `skos:Collection`.

akuckartz:
What does `It is recommended` mean? Is it normative or only a suggestion?

dr-shorthair:
'recommended' is still normative language, but it is a lower bar than 'required'. 

https://tools.ietf.org/html/rfc2119
https://www.iso.org/sites/directives/current/part2/index.xhtml#_idTextAnchor030

kcoyle:
The range actually assigns all values of dcat:theme the class of skos:Concept. 

>3.1 rdfs:range

>dfs:range is an instance of rdf:Property that is used to state that the values of a property are instances of one or more classes.

"... *are* instances ..." This is regardless of whether the value of dcat:theme has elsewhere been defined in a taxonomy using SKOS. 

It would be interesting to understand the functional purpose; perhaps this aids in queries by class rather than by property? Presumably this was discussed long ago, so no one may remember at this point.

aisaac:
Adding to what was said, and clarifying my words... 

It is perfectly fine to use a (regular) ontology like Schema.org, or even a (instance-level) database like Wikidata for dcat:themeTaxonomy. And with the current axiomatization, formally it's fine to use the members of these, e.g., an OWL class from the ontology (schema:Painting), or a city from DBpedia (dbpedia:london), as the object of dcat:theme.

Yet there is indeed an expectation that what's used for dcat:theme would qualify as skos:Concept. Some users may thus be reluctant to use schema:Painting or dbpedia:london on that basis.

[...]

PR #1380, opened 2021-06-07
Dcat issue 1364 first solution

riccardoAlbertoni:
This PR goes in the direction we discussed [last DCAT teleconference](https://www.w3.org/2021/05/26-dxwgdcat-minutes), it relates to issue #1364, #1153.
It explains in an Editorial Note that `skos:Concept`s can also be `owl:Classe`s when setting `dcat:theme`. 
It also changes a usage note to be more coherent to the dcat:themeTaxonomy's range relaxation introduced in DCAT 2.

An alternative solution is implemented in PR #1379.


dr-shorthair:
Editorial notes will not satisfy the OWLies :-( 
Might as well just drop `skos:Concept` in range and trust that `ObjectProperty` will ensure use of a well described term, probably from a 'controlled' vocabulary - see #1379

riccardoAlbertoni:
We keep this PR open, as it illustrates one of the possibilities we have considered so far. However,  the last dcat teleconference, see https://www.w3.org/2021/06/09-dxwgdcat-minutes, expressed a preference for the solution implemented in  PR #1379. So this PR is likely to be killed as soon as the other PR is finalized/merged.

PR #1382, opened 2021-06-13
Providing a svg for the main class diagram

### LINKED ISSUE #1294: Editorial issue: change UML diagrams to SVG
The UML diagrams in DCAT on a modern high-res monitor look terrible, because they are low-resolution bitmap graphics in PNG format.

I suggest switching to SVG vector graphics.

riccardoAlbertoni:
This PR addresses issue #1294  by providing a manipulable drawio diagram and an SVG diagram for the DCAT UML class.

As a side comment, even before this PR, the diagrams we are including in DCAT are made from different tools and have a different look and feel. 

preview: https://rawgit.com/w3c/dxwg/dcat-issue-1294/dcat/index.html#dcat-scope


dcat/index.html
```diff
@@ -298,7 +298,7 @@ <h3>DCAT scope</h3>
               </p>
            </aside>
     <figure id="UML_DCAT_All_Attr">
```
```suggestion
    <figure id="fig-dcat-all-attributes">
```

dcat/index.html
```diff
@@ -298,7 +298,7 @@ <h3>DCAT scope</h3>
...
-      <img alt="UML model of DCAT classes and properties" src="./images/DCAT-summary-all-attributes.png">
+      <img alt="UML model of DCAT classes and properties" src="./images/DCAT-summary-all-attributes.svg">
```
```suggestion
      <img alt="UML model of DCAT classes and properties" src="./images/dcat-all-attributes.svg">
```

riccardoAlbertoni: Shouldn’t we change in uncapitalised letter dropping
“-summary” also in the name of the svg file, if we change it in the html?

riccardoAlbertoni:
Thanks, @andrea-perego, for your suggestions. I have accepted all and made some changes to keep the references to the figure consistent.
If there are no other things you consider problematic, Please consider giving your approval to this PR.

andrea-perego (approved):
Thanks, @riccardoAlbertoni

andrea-perego:
> Thanks, @andrea-perego, for your suggestions. I have accepted all and made some changes to keep the references to the figure consistent.
> If there are no other things you consider problematic, Please consider giving your approval to this PR.

@riccardoAlbertoni , I think it may be worth changing accordingly also the name of the `.drawio` file.

riccardoAlbertoni:
> > Thanks, @andrea-perego, for your suggestions. I have accepted all and made some changes to keep the references to the figure consistent.
> > If there are no other things you consider problematic, Please consider giving your approval to this PR.
> 
> @riccardoAlbertoni , I think it may be worth changing accordingly also the name of the `.drawio` file.

Right, I've updated the name of the `.drawio` file. Thanks.

dr-shorthair:
Should Figure 6 (Qualified Relations) be updated as well?

riccardoAlbertoni:
> Should Figure 6 (Qualified Relations) be updated as well?

I have updated it.
It does not harm to have all the diagrams for DCAT with the same look and feel; also, I have taken the chance to change a 'dct:relation', which was not yet converted in dcterms.

Instead, I suggest not to update the one on Schema.org, fig 7, as it would take too many energies, and  I think we'd better spend those energies closing other issues.

agbeltran (approved):

agbeltran:
@riccardoAlbertoni thanks for this, it is looking good - it would be nice to update fig 7 too, as it looks inconsistent now - have you shared the link to draw.io? we can do the new figure and updates collaboratively

riccardoAlbertoni:
> @riccardoAlbertoni thanks for this, it is looking good - it would be nice to update fig 7 too, as it looks inconsistent now - have you shared the link to draw.io? we can do the new figure and updates collaboratively

A separate issue is created for Fig 7, see https://github.com/w3c/dxwg/issues/1383

aidig:
Seeing that the original UML class diagram was created in Sparx EA, might it have been (or still could be an) an idea to export the diagram from Sparx EA as .wmf file and then convert that to .svg?

PR #1384, opened 2021-06-18
making dcat:keyword a sub-property of dcterms:subject in the DCAT document

### LINKED ISSUE #175: Erratum: in DCAT 1.0 RDF dcat:keyword is a sub-property of dct:subject
In the RDF representation of DCAT v1, dcat:keyword is a sub-property of dct:subject - see https://www.w3.org/ns/dcat.ttl . 
In the recommendation document this is not mentioned - see tabulation at https://www.w3.org/TR/2014/REC-vocab-dcat-20140116/#class-dataset 
so this row in the description was omitted in the draft moving forward, 

I think this is an unintentional omission, but perhaps worth checking with @makxdekkers and @aisaac if you have any insight? Else it should be added here https://w3c.github.io/dxwg/dcat/#Property:dataset_keyword

### LINKED ISSUE #1374: In DCAT3, dcat:keyword is a subproperty of dcterms:subject
```dcat3.ttl``` includes ```dcat:keyword rdfs:subPropertyOf dcterms:subject ;```:
https://github.com/w3c/dxwg/blob/836b201802c0859c20bd066816e0c2cebb239f12/dcat/rdf/dcat3.ttl#L1034

which had been discussed as an errata here:
https://github.com/w3c/dxwg/issues/175

The solution is in @andrea-perego's comment:
https://github.com/w3c/dxwg/issues/175#issuecomment-375808963

reproduced here for convenience:
 "I think it is probably more correct to say that `dcat:keyword`, whose range is `rdfs:Literal`, is a subproperty of [`dc:subject`](http://dublincore.org/documents/dcmi-terms/#elements-subject). By contrast, [`dct:subject`](http://dublincore.org/documents/dcmi-terms/#terms-subject) (quoting) "is intended to be used with non-literal values" - as `dcat:theme`."

riccardoAlbertoni:
This would close #175 and #1374, 
it aligns the REC with the RDF making dcat:keyword a sub-property of dcterms:subject.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1405,7 +1405,7 @@ table.simple {width:100%;}
                 In DCAT 1 [[?VOCAB-DCAT-20140116]] the domain of <code>dcat:keyword</code> was <code>dcat:Dataset</code>, which limited use of this property in other contexts. The domain has been relaxed in DCAT 2 - see Issue <a href="https://github.com/w3c/dxwg/issues/121">#121</a>.
               </p>
              <p>
-                 DCAT 3 defines dcat:keyword  as subproperty of <code>dcterms:subject</code> as the latter might accepts literals.          
+                 DCAT 3 defines <code>dcat:keyword</code> as sub-property of <code>dcterms:subject</code> as the latter might accepts literals.          
                 </p>
             </aside>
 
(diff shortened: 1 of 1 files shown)


andrea-perego (commented):
@riccardoAlbertoni , I still think we should just drop the range restriction in the RDF definition.

PR #1410, opened 2021-10-01
DCAT removing inverse properties from Vocabulary Specification

riccardoAlbertoni:
This PR removes the properties  <code>dcat:next</code> and <code>dcterms:isReplacedBy</code> from the Vocabulary specification (section  6).  

Still to decide if other properties mentioned in the inverse property section (e.g.,  <code>dcat:hasVersion</code> and <code>prov:generated</code>) need to be removed as well.
 
Preview: https://raw.githack.com/w3c/dxwg/DCAT-adjustementForInverseProperties/dcat/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2FDCAT-adjustementForInverseProperties%2Fdcat%2Findex.html

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -852,7 +852,7 @@ table.simple {width:100%;}
 
 <li><a href="#Property:resource_first">first</a></li>
 <li><a href="#Property:resource_last">last</a></li>
-<li><a href="#Property:resource_next">next</a></li>
+<!-- <li><a href="#Property:resource_next">next</a></li> -->
 <li><a href="#Property:resource_previous">previous</a></li>
 
 </ul>
@@ -1028,7 +1028,7 @@ table.simple {width:100%;}
 <!-- Series properties -->			
             <li><a href="#Property:resource_first">first</a></li>
             <li><a href="#Property:resource_last">last</a></li>
-            <li><a href="#Property:resource_next">next</a></li>
+            <!-- <li><a href="#Property:resource_next">next</a></li> -->
             <li><a href="#Property:resource_previous">previous</a></li>
 			
         </ul>
@@ -2064,7 +2064,7 @@ table.simple {width:100%;}
 <th class="prop">See also:</th>
 <td>
 <a href="#Property:resource_last"></a>,
-<a href="#Property:resource_next"></a>,
+<!--<a href="#Property:resource_next"></a>, -->
 <a href="#Property:resource_previous"></a>.
 </td>
 </tr>
@@ -2108,7 +2108,7 @@ table.simple {width:100%;}
 <th class="prop">See also:</th>
 <td>
 <a href="#Property:resource_first"></a>,
-<a href="#Property:resource_next"></a>,
+<!-- <a href="#Property:resource_next"></a>, -->
 <a href="#Property:resource_previous"></a>.
 </td>
 </tr>
@@ -2118,6 +2118,7 @@ table.simple {width:100%;}
 
 </section>
 
+<!--    
 <section id="Property:resource_next">
 
 <h4>Property: next</h4>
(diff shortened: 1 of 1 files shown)


andrea-perego:
@riccardoAlbertoni said:

> Still to decide if other properties mentioned in the inverse property section (e.g., `dcat:hasVersion` and `prov:generated`) need to be removed as well.

I created PR https://github.com/w3c/dxwg/pull/1413 to deal with `dcat:hasVersion` / `dcat:isVersionOf`. The proposal implemented there is to keep `dcat:hasVersion` in §6, and move instead `dcat:isVersionOf` to the section on inverse properties.

The rationale is that this reflects the PAV approach, where only `pav:hasVersion` (≡ `dcat:hasVersion`) is defined, and not its inverse. Moreover, `pav:hasVersion` is the superproperty of `pav:hasCurrentVersion` (≡ `dcat:hasCurrentVersion`), which needs to be specified to link the abstract resource to the current version of the actual resource. For this reason, `dcat:hasCurrentVersion` should be kept in §6; in such a case, its superproperty (`dcat:hasVersion`) cannot be dropped from §6.

andrea-perego:
About property `prov:generated` (not included in §6), it seems to be used only in the examples. Should this be the case, the fix may be limited to ensuring that all the examples use it in addition to `prov:wasGeneratedBy` - see, e.g., [Example 45](https://raw.githack.com/w3c/dxwg/DCAT-adjustementForInverseProperties/dcat/index.html#ex-conformance-degree) and [Example 48](https://raw.githack.com/w3c/dxwg/DCAT-adjustementForInverseProperties/dcat/index.html#ex-conformance-test-results-earl).

riccardoAlbertoni:
Example 54 uses an inverse property `dcterms:references` but not `dcterms:isReferencedBy`, I wonder if we should add 
`:d33937.owl dcterms:isReferencedBy :d33937
`to make the example more consistent with DCAT policies on inverse property

riccardoAlbertoni:
> @riccardoAlbertoni said:
> 
> > Still to decide if other properties mentioned in the inverse property section (e.g., `dcat:hasVersion` and `prov:generated`) need to be removed as well.
> 
> I created PR #1413 to deal with `dcat:hasVersion` / `dcat:isVersionOf`. The proposal implemented there is to keep `dcat:hasVersion` in §6, and move instead `dcat:isVersionOf` to the section on inverse properties.
> 
> The rationale is that this reflects the PAV approach, where only `pav:hasVersion` (≡ `dcat:hasVersion`) is defined, and not its inverse. Moreover, `pav:hasVersion` is the superproperty of `pav:hasCurrentVersion` (≡ `dcat:hasCurrentVersion`), which needs to be specified to link the abstract resource to the current version of the actual resource. For this reason, `dcat:hasCurrentVersion` should be kept in §6; in such a case, its superproperty (`dcat:hasVersion`) cannot be dropped from §6.

That was one of the things into which I was stuck.   I thought of defining `dcat:isVersionOf` as the inverse of `pav:hasVersion`, but I was not much convinced of such a solution. Your proposal sounds more reasonable to me. 

A concern is that the metadata of the abstract resources must be updated each time a new version is added. However, I am not sure there is a way to avoid this, even if we had chosen `dcat:isVersionOf` instead of `dcat:hasVersion`,  we would have needed to update the `dcat:hasCurrentVersion`.
 
The solution you have implemented which mimics the directions used by PAV looks like the proper equilibrium.

andrea-perego:
@riccardoAlbertoni said:

> Example 54 uses an inverse property `dcterms:references` but not `dcterms:isReferencedBy`, I wonder if we should add `:d33937.owl dcterms:isReferencedBy :d33937 `to make the example more consistent with DCAT policies on inverse property

Makes sense to me.

riccardoAlbertoni:
> About property `prov:generated` (not included in §6), it seems to be used only in the examples. Should this be the case, the fix may be limited to ensuring that all the examples use it in addition to `prov:wasGeneratedBy` - see, e.g., [Example 45](https://raw.githack.com/w3c/dxwg/DCAT-adjustementForInverseProperties/dcat/index.html#ex-conformance-degree) and [Example 48](https://raw.githack.com/w3c/dxwg/DCAT-adjustementForInverseProperties/dcat/index.html#ex-conformance-test-results-earl).

Done.


> @riccardoAlbertoni said:
> 
> > Example 54 uses an inverse property `dcterms:references` but not `dcterms:isReferencedBy`, I wonder if we should add `:d33937.owl dcterms:isReferencedBy :d33937 `to make the example more consistent with DCAT policies on inverse property
> 
> Makes sense to me.

As discussed, we are going to postpone this, as it is possible that `dcterms:references` is not the property we want there. In case it was the right property, the addition of `dcterms:references` can be done when dealing with issue #1416.

andrea-perego:
Thanks, @riccardoAlbertoni . I think we can merge this PR.

andrea-perego (approved):

dr-shorthair (approved):

PR #1423, opened 2021-11-09
Editorial revision to definitions of locn:geometry, dcat:bbox, dcat:centroid

### LINKED ISSUE #1392: Semantics of the dcat:bbox attribute could (should?) be more explicit
The http://www.w3.org/ns/dcat#bbox attribute is somewhat vague (or perhaps lends itself to misinterpretation) dcat describes resources (data-sets) and while I suspect that this attribute is intended to say something about what the resource is about, it could also be interpreted as the (physical) nature of the resource itself. A somewhat forced  example: an old fashioned phone-book is about a region or a municipality but simultaneously a phonebook has a bounding-box (describing how thick it is for example and where it lies on my shelf). 

In general, naming an attribute after its type (or range in rdfs terms) is not specific enough to be clear about its intent (although that might be a personal preference). While the the spec mentions the range is 'intentionally generic' the technical origin of this concept (CAD drawing, BIM etc, where this is a mathematical construct to represent a 3 dimensional object ) will most likely cause most people (and people programming machines) to ignore the 'intentionally generic' part of the spec.

My suggestion would either be to rename this attribute to make its intent more clear (dcat:subjectBoundingBox for example), or elaborate in the description on its intended meaning (phone book vs region the phonebook covers)

andrea-perego:
See https://github.com/w3c/dxwg/issues/1392#issuecomment-907498910

This PR tries to address the ambiguity of the term "resource" in the definitions of properties `locn:geometry`, `dcat:bbox`, and `dcat:centroid` by replacing it with "spatial thing", in the SDW-BP sense - see https://www.w3.org/TR/sdw-bp/#spatial-things-features-and-geometry

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -3666,7 +3666,7 @@ The interval can also be open - i.e., it can have just a start or just an end.</
 <!--
 <tr><th class="prop">Definition:</th><td>Associates any resource with the corresponding geometry. [[LOCN]]</td></tr>
 -->
-<tr><th class="prop">Definition:</th><td>Associates any spatial thing [[?SDW-BP]] with the corresponding geometry.</td></tr>
+<tr><th class="prop">Definition:</th><td>Associates a spatial thing [[?SDW-BP]] with a corresponding geometry.</td></tr>
 <tr><th class="prop">Range:</th><td><a data-cite="LOCN#locn:Geometry"><code title="https://www.w3.org/ns/locn#Geometry">locn:Geometry</code></a></td></tr>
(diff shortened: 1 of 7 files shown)


riccardoAlbertoni (approved):

andrea-perego:
Thanks for endorsing this proposal, @riccardoAlbertoni .

Once we agree on the revision to be made, we need to update the RDF vocabulary accordingly before merging this PR.

dcat/index.html
```diff
@@ -3663,7 +3663,10 @@ <h4>Property: geometry</h4>
 <th>RDF Property:</th>
 <td><a data-cite="LOCN#locn:geometry"><code title="http://www.w3.org/ns/locn#geometry">locn:geometry</code></a></td>
 </tr>
+<!--
 <tr><th class="prop">Definition:</th><td>Associates any resource with the corresponding geometry. [[LOCN]]</td></tr>
+-->
+<tr><th class="prop">Definition:</th><td>Associates any spatial thing [[?SDW-BP]] with the corresponding geometry.</td></tr>
```
dr-shorthair: is it 'a' rather than 'the' corresponding geometry?

andrea-perego:
```suggestion
<tr><th class="prop">Definition:</th><td>Associates a spatial thing [[?SDW-BP]] with a corresponding geometry.</td></tr>
```

davebrowning (approved):
Noting that the RDF needs to be updated to bring it into line, but the changes look like the right step.

andrea-perego:
> Noting that the RDF needs to be updated to bring it into line, but the changes look like the right step.

Thanks, @davebrowning . I just revised the English definitions in `dcat3.ttl` and `dcat-external.ttl`. We need to generate the other serialisations before merging.

riccardoAlbertoni:
The other serializations have been generated.

PR #1450, opened 2022-02-04
deleted the reference to issue 1336 in DCAT PWD

### LINKED ISSUE #1336: Applying PROV's lightweight approach for DCAT inverse properties
Discussions on adding inverse DCAT properties have appeared in distinct issues (e.g., #1322 and #1335).
Some people feel DCAT should recommend only one direction of properties, essentially to avoid extra efforts when maintaining and consuming the metadata. Others argue that inverse properties are not that difficult to maintain. 

A middle-out option would be supporting inverse properties with a "lightweight approach" such as the one adopted by PROV-O (https://www.w3.org/TR/prov-o/#inverse-names).

The idea is to avoid inverse properties (unless grounded explicitly by use cases). But rather, to claim in the DCAT Appendix the name of the inverse properties to ensure interoperability in case any specific application needs to use the non-recommended inverse properties.
In a recent DCAT teleconference,  we started discussing whether this strategy is suitable for DCAT, and we agreed on following/exploring the PROV approach for inverse properties (see https://www.w3.org/2021/03/17-dxwgdcat-minutes#r02). 

What do group members and contributors think about this approach?

riccardoAlbertoni:
Considering the issue 1336 have been labelled "due for closing" for more than a month, no one has complained about the proposal of closing it, this PR remove the reference to the issue in the DCAT PWD and closes #1336

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -3863,7 +3863,7 @@ The interval can also be open - i.e., it can have just a start or just an end.</
 
 <h2>Use of inverse properties</h2>
 
-<p>The properties described in <a href="#vocabulary-specification"></a> do not intentionally include inverses, with the purpose of ensuring interoperability also in systems not making use of OWL reasoning.</p>
+<p>The properties described in <a href="#vocabulary-specification"></a> do not include inverses intentionally, with the purpose of ensuring interoperability also in systems not making use of OWL reasoning.</p>
 
 <p>However, recognizing that inverses are needed for some use cases, DCAT supports them, but with the requirement that they MAY be used only <em>in addition to</em> those described in <a href="#vocabulary-specification"></a>, and that they MUST NOT be used to replace them.</p>
 
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -3863,14 +3863,6 @@ The interval can also be open - i.e., it can have just a start or just an end.</
 
 <h2>Use of inverse properties</h2>
 
-<aside class="ednote">
-<p>This section is a draft for discussion about support for inverse properties in DCAT.</p>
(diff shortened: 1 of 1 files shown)


andrea-perego (approved):

dcat/index.html
```diff
@@ -3863,14 +3863,6 @@ <h4>Property: checksum value</h4>
...
 
-<aside class="ednote">
-<p>This section is a draft for discussion about support for inverse properties in DCAT.</p>
-<p>Related issues:</p>
-
-<div class="issue" data-number="1336"></div>
-
-</aside>
-
 <p>The properties described in <a href="#vocabulary-specification"></a> do not intentionally include inverses, with the purpose of ensuring interoperability also in systems not making use of OWL reasoning.</p>
```
agbeltran: OK with removing the note section

One issue I see with the following text is that we say:

"The properties described in <a href="#vocabulary-specification"></a> do not intentionally include inverses"...

but I believe it should be:

"The properties described in <a href="#vocabulary-specification"></a> do not include inverses intentionally"

agbeltran (changes requested):

dcat/index.html
riccardoAlbertoni: Changing as in Alejandra's suggestion
```suggestion
<p>The properties described in <a href="#vocabulary-specification"></a> do not include inverses intentionally, with the purpose of ensuring interoperability also in systems not making use of OWL reasoning.</p>
```

riccardoAlbertoni: @agbeltran I have included your correction. Could you accept the PR?

agbeltran (approved):

PR #1451, opened 2022-02-04
adjusting dcat:version's definition to include both number and text m…

### LINKED ISSUE #1442: The definition of a version should not be limited to numbers
In section 6.4.27, we have Property: version  = "The version number of a resource", but we don't want to prescribe numbers only.

riccardoAlbertoni:
…ore explicitly

closes #1442

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1908,7 +1908,7 @@ table.simple {width:100%;}
 <tr><th>RDF Property:</th><td><a href="http://www.w3.org/ns/dcat#version"><code>dcat:version</code></a></td></tr>
 <tr>
 <th class="prop">Definition:</th>
-<td>The version number of a resource [[?PAV]].</td>
+<td>A number or text that identifies a particular revision of the resource.</td>
 </tr>
 <tr>
 <th class="prop">Equivalent property:</th>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -1908,7 +1908,7 @@ table.simple {width:100%;}
 <tr><th>RDF Property:</th><td><a href="http://www.w3.org/ns/dcat#version"><code>dcat:version</code></a></td></tr>
 <tr>
 <th class="prop">Definition:</th>
-<td>A number or text that identifies a particular revision of the resource.</td>
+<td>The version indicator (name or identifier) of a resource.</td>
 </tr>
 <tr>
 <th class="prop">Equivalent property:</th>


andrea-perego (commented):
@riccardoAlbertoni , I suggest we discuss this change in the next DCAT meeting before merging.

I see a number of issues if we replace "version" with "revision":
1. If the problem is what "version" means, then we need also to define what a revision is.
2. If  we use "revision", then it's a bit odd to call the property dcat:version
3. The two points above apply also to all the other properties related to versioning.

About the original issue (i.e., support to both text and numbers), I suggest we revise the definition as per what said in the usage note:

> The version indicator (name or identifier) of a resource.

dcat/index.html
```diff
@@ -1908,7 +1908,7 @@ <h4>Property: version</h4>
 <tr><th>RDF Property:</th><td><a href="http://www.w3.org/ns/dcat#version"><code>dcat:version</code></a></td></tr>
 <tr>
 <th class="prop">Definition:</th>
-<td>The version number of a resource [[?PAV]].</td>
+<td>A number or text that identifies a particular revision of the resource.</td>
```
```suggestion
<td>The version indicator (name or identifier) of a resource.</td>
```

riccardoAlbertoni:
> @riccardoAlbertoni , I suggest we discuss this change in the next DCAT meeting before merging.
> 
> I see a number of issues if we replace "version" with "revision":
> 
> 1. If the problem is what "version" means, then we need also to define what a revision is.
> 2. If  we use "revision", then it's a bit odd to call the property dcat:version
> 3. The two points above apply also to all the other properties related to versioning.
> 
> About the original issue (i.e., support to both text and numbers), I suggest we revise the definition as per what said in the usage note:
> 
> > The version indicator (name or identifier) of a resource.

I agree. So I've changed the phrasing as you have suggested.

@kcoyle, @agreiner: Are you ok with this phrasing?

agbeltran:
should we continue the discussion in https://github.com/w3c/dxwg/issues/1442? one of the arguments was that we shouldn't use 'version' to define 'version', and that's why 'revision' was seen as an alternative - I agree we would probably need further clarification

davebrowning (approved):

riccardoAlbertoni:
> should we continue the discussion in #1442? one of the arguments was that we shouldn't use 'version' to define 'version', and that's why 'revision' was seen as an alternative - I agree we would probably need further clarification

@agbeltran see https://github.com/w3c/dxwg/issues/1442#issuecomment-1033158236

Does it sound convincing to you?

andrea-perego (approved):

PR #1452, opened 2022-02-15
Charter 2022

plehegar:
Based on
  https://www.w3.org/2022/02/01-dxwg-minutes.html#t02


plehegar:
The WG is still hesitant to keep "[Content Negotiation by Profile](https://www.w3.org/TR/dx-prof-conneg/)" on the REC track, due to very minimum progress/interest. See [February 15 minutes](]https://www.w3.org/2022/02/15-dxwg-minutes.html#t03).

cc @rob-metalinkage @nicholascar @larsgsvensson

plehegar:
As a reminder, we did a poll in 2019 to move [Conneg to CR](https://www.w3.org/2002/09/wbs/99375/CONNEG_to_CR/results).

agreiner:
And we also voted as a group to move it to being a WG Note, sometime between that poll and the end of October 2019.

nicholascar:
It is currently a Note and I do recall voting for it's publication as such. 

We've, only now, finished larger-scale, independent, production deployments of ConnegP systems (Rob at OGC, me at Aust Government) which means we need to update the [Implementation Report](https://w3c.github.io/dx-connegp/connegp-implementation-report/). I think such an updated report is needed for any moves from Note back to CR?

plehegar:
@nicholascar , ConnegP is currently on the REC, according to the charter. Some in the Group would like to remove from the REC track...

nicholascar:
@plehegar do you think we need to indicate a likely timeline for implementation reports and so on to retain the REC track status? This is assuming the reason for wanting to remove it from REC is the lack of recent progress.

plehegar:
Approved during https://www.w3.org/2022/03/15-dxwg-minutes.html#t03

PR #1456, opened 2022-02-21
PR Issue 1443 - renaming prev, next, last, first into previousInSeries..ect

### LINKED ISSUE #1443: Properties relating to series can be confused with those for versions
For a cataloged resource, previous and first and last properties can easily be confused with versions. Would it be more helpful to users if we called them previousInSeries, firstInSeries, lastInSeries?

riccardoAlbertoni:
closes #1443

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -848,10 +848,10 @@ table.simple {width:100%;}
             <li><a href="#Property:resource_version">version</a></li>
             <li><a href="#Property:resource_version_notes">version notes</a></li>
 
-<li><a href="#Property:resource_first">first</a></li>
-<li><a href="#Property:resource_last">last</a></li>
+<li><a href="#Property:resource_first_in_series">first in series</a></li>
+<li><a href="#Property:resource_last_in_series">last in series</a></li>
 <!-- <li><a href="#Property:resource_next">next</a></li> -->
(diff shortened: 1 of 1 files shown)


riccardoAlbertoni:
Still to decide whether or not  dropping the occurences of "ordered collections" in the definitions of  previousInSeries, firstInSeries, lastInSeries.

andrea-perego (commented):
I suggest we keep this PR on hold until we discuss the overall issue in the subgroup meeting.

dr-shorthair (approved):

riccardoAlbertoni:
> I suggest we keep this PR on hold until we discuss the overall issue in the subgroup meeting.

Yes, I agree we  need to discuss this PR in the DCAT call. I have just noticed some choices that we might need to doublecheck  before any merging.

agreiner:
I'm having a hard time thinking of an example of an ordered collection that would not be a series.

riccardoAlbertoni:
Do not merge!
According to resolution  https://www.w3.org/2022/02/22-dxwgdcat-minutes#r04
an alternative PR will be prepared adding to dcat:prev etc. properties an explanation of which is their scope and purpose, and why they are associated with dcat:Resource. [in short, they are  used in dataset series but they are not only for dataset series]

PR #1458, opened 2022-02-22
PR to refine the message about dataset series' member  (issue 1429)

### LINKED ISSUE #1429: Model Series of Data as Distributions of a single Dataset
In the Swiss DCAT profile we are modeling Dataseries such as yearly elections in a single dataset, where each distribution contains the election data of one year. I know that is considered an antipattern by DCAT. And I also know that all properties of Distributions are designed in such a way, that it is assumed that the all Distributions of a Dataset have comparable content. 

But there is also a sentence at https://www.w3.org/TR/vocab-dcat-3/#Class:Distribution:

> Nevertheless, the question of whether different representations can be understood to be distributions of the same dataset, or distributions of different datasets, is application specific. Judgement about how to describe them is the responsibility of the provider, taking into account their understanding of the expectations of users, and practices in the relevant community. 

I was recently surprised that we are not the only profile modeling a series of data with Distributions, and that there is a certain resistance to give this up this pattern, since the overall impression is, that this would make for too many datasets and the data portals would get harder to mangage with too many datasets.

On the other hand if DCAT would approve of that pattern or antipattern, then properties would be needed to describe the content in each Distribution. In the Swiss profile we added the attribute `dct:coverage` to Distributions, with a domain of either spatial or temporal values. 

I am just curious what DCAT's opinion on this topic is and I want of launch a discussion about this: Can't it be that sometimes Distributions differ in their content and shouldn't DCAT alos support these use cases with appropriate properties?

riccardoAlbertoni:
This PR is a work in progress attempt aiming at sharpen the message about dataset series members.

Rendered at: http://raw.githack.com/w3c/dxwg/dcat-issue-1429/dcat/index.html

Diff at: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-issue-1429%2Fdcat%2F

This when merged is expected to close #1429.

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2499,7 +2499,7 @@ table.simple {width:100%;}
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
 
-<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
+<p>For guidance on the use of this property, see <a href="#dataset-series"></a>.</p>
 
     </section>
         <section id="Property:dataset_spatial">
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -5125,7 +5125,7 @@ ex:budget-2018-it a dcat:Dataset ;
 
 </section>
 
-<section id="dataset-series-before-dcat3">
+<!--section id="dataset-series-before-dcat3">
 
 <h2>Dataset series in existing DCAT implementations</h2>
 
@@ -5143,7 +5143,7 @@ ex:budget-2018-it a dcat:Dataset ;
 <p>In both cases, the dataset series is sometimes soft-typed by using the [[DCTERMS]] property <code>dcterms:type</code> (e.g., this is the approach used in [[?GeoDCAT-AP]], and adopted in [[?DCAT-AP-IT]] and [[?GeoDCAT-AP-IT]]).</p>
 
 <p>These options are not formally incompatible with DCAT, so they can cohexist with <code>dcat:DatasetSeries</code> during the upgrade to DCAT 3.</p>
-
+-->
 <!--
 <p>Compared with the second option, the first one may have the advantage of simplifying metadata management, and avoiding the creation of datasets having the same values for almost all their metadata elements. On the other hand, this approach reduces the ability of being discovered, as distribution metadata are not rich as datasets' ones. Moreover, using distributions may result cumbersome or unfeasible when the number of child datasets is too high.</p>
 
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2723,6 +2723,8 @@ table.simple {width:100%;}
             <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
             </table>
 
+<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
+
        <div class="issue" data-number="1335"></div>  
        <!--section id="Property:dataset_series_series_member">
             <h4>Property:series member</h4>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2499,6 +2499,9 @@ table.simple {width:100%;}
                 <!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
+
+<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
+
     </section>
         <section id="Property:dataset_spatial">
             <h4>Property: spatial/geographical coverage</h4>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2495,7 +2495,7 @@ table.simple {width:100%;}
                 <!--tr><th class="prop">Inverse property of:</th>
                 <td><a href="#Property:dataset_series_series_member"><code>dcat:seriesMember</code></a></td>
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>.<!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.--></td></tr>
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
     </section>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2722,7 +2722,7 @@ table.simple {width:100%;}
             <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
             </table>
 
-<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
+<p>For guidance on the use of this property, see <a href="#dataset-series"></a>.</p>
 
        <div class="issue" data-number="1335"></div>  
        <!--section id="Property:dataset_series_series_member">
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2495,7 +2495,7 @@ table.simple {width:100%;}
                 <!--tr><th class="prop">Inverse property of:</th>
                 <td><a href="#Property:dataset_series_series_member"><code>dcat:seriesMember</code></a></td>
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
     </section>
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -5129,9 +5129,9 @@ ex:budget-2018-it a dcat:Dataset ;
 
 <h2>Dataset series in existing DCAT implementations</h2>
 
-<aside class="ednote">
+<!--aside class="ednote">
 <p>To be decided whether to keep or not this section.</p>
-</aside>
+</aside-->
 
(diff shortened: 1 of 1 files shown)


dcat/index.html
```diff
@@ -2495,7 +2495,7 @@ <h4>Property: in series</h4>
                 <!--tr><th class="prop">Inverse property of:</th>
                 <td><a href="#Property:dataset_series_series_member"><code>dcat:seriesMember</code></a></td>
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>.<!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.--></td></tr>
```
dr-shorthair: This is quite a strong change. It is clearly justified in principle. However, it may not be sustainable in practice. DCAT does not have wide adoption already. Maybe it is more influential than adopted.

riccardoAlbertoni: > This is quite a strong change. 

It might be not that strong. We are using "SHOULD", not "MUST". 

"SHOULD"  means that there may exist valid reasons in particular circumstances to ignore a  particular item, but the full implications must be understood and carefully weighed before choosing a different course.

The change attempts to clarify that using datasets as members in series is the recommended solution. 
Other solutions are possible, but this is the one DCAT 3 considers the most flexible and versatile and the one recommended for the new implementations.

Existing or even future implementations can do otherwise if they have thoroughly considered the consequences. 

We don't want to alienate the existing implementations. Therefore, we explained that current solutions are not incompatible with DCAT 3 during the upgrade to DCAT 3 (see the non-normative section beginning at line 5128).

However,  unless we want to support more than a solution and thus accept interoperability issues intrinsic in the availability of more than one recommended solution, we need to be clear on the recommended approach. 
I am not  100% sure this is the best way to do it.

Do you think a "MAY" instead of "SHOULD" would be more appropriate? Do you have any suggestions on how to convey the message more appropriately?

riccardoAlbertoni:
Though I guess there will be further discussion and request of adjustments for this PR, I am marking this as ready for review to solicit opinions ;)

dcat/index.html
dr-shorthair: Thanks Riccardo. I'm OK with the proposal. Was just checking, and I note the use of SHOULD.

dr-shorthair (approved):

dcat/index.html
andrea-perego: @riccardoAlbertoni , I wonder whether we'd rather drop altogether the usage note. The fact that members of a series are instances of `dcat:Dataset` is already made explicit in the definition of this property, so there's no benefit in restating it in the usage note.

dcat/index.html
```diff
@@ -2713,6 +2713,7 @@ <h3>Class: Dataset Series</h3>
             <tr><th>RDF Class:</th><td><a href="http://www.w3.org/ns/dcat#DatasetSeries"><code>dcat:DatasetSeries</code></a></td></tr>
             <tr><th class="prop">Definition:</th><td>A collection of datasets that are published separately, but share some common characteristics that groups them. <!--, and could also be made available as a single dataset.--></td></tr>
             <tr><th class="prop">Sub-class of:</th><td><a href="#Class:Dataset"><code>dcat:Dataset</code></a></td></tr>
+            <tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>.<!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.--></td></tr>
```
andrea-perego: @riccardoAlbertoni , I wonder whether we'd rather drop altogether the usage note. The fact that members of a series are instances of `dcat:Dataset` is already made explicit in the definition of this property, so there's no benefit in restating it in the usage note.

dcat/index.html
riccardoAlbertoni: I am quite tempted to drop the note.

Though, I wonder if dropping the note might imply that  Items of the Dataset series MUST be Datasets, namely if readers might perceive the message stronger than what I thought we wanted to deliver.

"SHOULD" discourages other solutions, but it accepts exceptions.

andrea-perego: I see the point, but using SHOULD without indicating alternatives would be close to a MUST. As we have a specific section about alternative solutions, I would just add a link to that section from both `dcat:DatasetSeries` and `dcat:inSeries`.

dr-shorthair: Yes - that is the problem. A SHOULD alongside a MAY is probably needed here.

riccardoAlbertoni: Andrea's suggestion has been implemented via [62945d9](https://github.com/w3c/dxwg/pull/1458/commits/62945d971c2c41f2d8f14c86db9323c6518ebdf7)

dcat/index.html
riccardoAlbertoni: added the see also to  section "Dataset series in existing DCAT implementations"  via [62945d9](https://github.com/w3c/dxwg/pull/1458/commits/62945d971c2c41f2d8f14c86db9323c6518ebdf7)

riccardoAlbertoni:
@andrea-perego, please double-check if I have implemented your suggestion correctly via  [62945d9](https://github.com/w3c/dxwg/pull/1458/commits/62945d971c2c41f2d8f14c86db9323c6518ebdf7),

dcat/index.html
```diff
@@ -2495,7 +2495,8 @@ <h4>Property: in series</h4>
...
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>    
+                <!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
     </section>
```
andrea-perego, line 2567:
```suggestion

<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>

    </section>
```

dcat/index.html
```diff
@@ -2495,7 +2495,8 @@ <h4>Property: in series</h4>
                 <!--tr><th class="prop">Inverse property of:</th>
                 <td><a href="#Property:dataset_series_series_member"><code>dcat:seriesMember</code></a></td>
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>    
+                <!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
```
```suggestion
                <!--<tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>.The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
```

dcat/index.html
```diff
@@ -2495,7 +2495,8 @@ <h4>Property: in series</h4>
                 <!--tr><th class="prop">Inverse property of:</th>
                 <td><a href="#Property:dataset_series_series_member"><code>dcat:seriesMember</code></a></td>
                 </tr-->
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
```
```suggestion
```

dcat/index.html
```diff
@@ -2713,8 +2714,10 @@ <h3>Class: Dataset Series</h3>
...
             <tr><th class="prop">Sub-class of:</th><td><a href="#Class:Dataset"><code>dcat:Dataset</code></a></td></tr>
+            <!--tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>.<!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
             <tr><th class="prop">Usage note:</th><td> Dataset series can be also soft-typed via property <code>dcterms:type</code> as in the approach used in [[?GeoDCAT-AP]], and adopted in [[?DCAT-AP-IT]] and [[?GeoDCAT-AP-IT]]).</td></tr>
             <tr><th class="prop">Usage note:</th><td>Common scenarios for dataset series include: time series composed of periodically released subsets; map-series composed of items of the same type or theme but with differing spatial footprints.</td></tr>
+            <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
```
```suggestion
            <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
```

dcat/index.html
```diff
@@ -2713,8 +2714,10 @@ <h3>Class: Dataset Series</h3>
...
             </table>
```
andrea-perego, line 2787:
```suggestion

<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>

```

andrea-perego (changes requested):
I have added a number of inline change proposals.

Basically, the idea is to put the links to the section on dataset series as done for the versioning properties, and not in the "See also" row of class/property definitions.

dcat/index.html
```diff
@@ -2495,9 +2495,12 @@ <h4>Property: in series</h4>
...
-                <tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>. The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.</td></tr>
+                <!--<tr><th class="prop">Usage note:</th><td> Normally, child datasets in dataset series are represented as <code>dcat:Dataset</code>.The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
                 <!--tr><th class="prop">See also:</th><td>Sub-properties of <code>dcterms:hasPart</code> in particular <a href="#Property:catalog_dataset"><code>dcat:dataset</code></a>, <a href="#Property:catalog_catalog"><code>dcat:catalog</code></a>, <a href="#Property:catalog_service"><code>dcat:service</code></a>. </td></tr-->
                 </table>
+
+<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
```
riccardoAlbertoni: @andrea-perego : I wonder if we should point to the section #dataset-series instead of the subsection about  existing implementations.

```suggestion
<p>For guidance on the use of this property, see <a href="#dataset-series"></a>.</p>
```

dcat/index.html
```diff
@@ -2713,10 +2716,14 @@ <h3>Class: Dataset Series</h3>
...
+            <!--tr><th class="prop">Usage note:</th><td> Child datasets in dataset series SHOULD be represented as <code>dcat:Dataset</code>.<!--The use of <code>dcat:Distribution</code> for typing child datasets is however recognized as a possible alternative, whenever it addresses more effectively the requirements of a given application scenario.></td></tr-->
             <tr><th class="prop">Usage note:</th><td> Dataset series can be also soft-typed via property <code>dcterms:type</code> as in the approach used in [[?GeoDCAT-AP]], and adopted in [[?DCAT-AP-IT]] and [[?GeoDCAT-AP-IT]]).</td></tr>
             <tr><th class="prop">Usage note:</th><td>Common scenarios for dataset series include: time series composed of periodically released subsets; map-series composed of items of the same type or theme but with differing spatial footprints.</td></tr>
+            <tr><th class="prop">See also:</th><td><a href="#dataset-series-before-dcat3"></a></td></tr>
             </table>
 
+<p>For guidance on the use of this property, see <a href="#dataset-series-before-dcat3"></a>.</p>
```
riccardoAlbertoni: @andrea-perego : I wonder if we should point to the section #dataset-series instead of the subsection about  existing implementations.

```suggestion
<p>For guidance on the use of this property, see <a href="#dataset-series"></a>.</p>
```

dcat/index.html
riccardoAlbertoni: Not sure here what was the intention. I guess we wanted to drop the table line

```suggestion
     
```

dcat/index.html
andrea-perego: Agreed.

dcat/index.html
andrea-perego: Agreed.

dcat/index.html
andrea-perego: Yes, it was. Thanks for spotting it.

andrea-perego (approved):

PR #1461, opened 2022-02-24
PR adding the dcat:inSeries' inverse to section 'use of inverse' - issue 1335

### LINKED ISSUE #1335: Do we need a new property dcat:seriesMember  defined as dcat:inSeries's inverse?
Do we need a new property (let's name it `dcat:seriesMember`) defined as `dcat:inSeries`'s inverse?

Requests for `dcat:inSeries`'s inverse to connect the dataset series to its children have been discussed in issue #1307. 

Part of the previous discussion is reported below.

_Originally posted by @riannella in https://github.com/w3c/dxwg/issues/1307#issuecomment-798902564_ 
> In our sector, for example, a Regulator would say the more important relationship was from the series to the parts.
Let's have both.

Most of the concerns against adding this property elaborate on maintaining metadata consistent when a dataset is added to the series. 

_Originally posted by @jakubklimek in https://github.com/w3c/dxwg/issues/1307#issuecomment-792503423_
>@dr-shorthair @riannella From the point of view of an application developer, let me again state that I would strongly prefer to standardize only one of the two mutually inverse properties.
Otherwise, compliant applications and queries need to always account for the possibility that in one data catalog, the datasets are connected to the series using one property, in another using the second, which makes them more complex with each inverse property pair.

_Originally posted by @agreiner in https://github.com/w3c/dxwg/issues/1307#issuecomment-793225598_
> I would suggest keeping only the inSeries property so that one needn't worry about updating metadata for the series every time a new dataset is added to that series.

Keeping the metadata of  series consistent while adding new datasets seems  to  be required anyway by the assumption of  "upstream inheritance"  discussed in #1273  

_Originally posted by @riccardoAlbertoni in https://github.com/w3c/dxwg/issues/1307#issuecomment-801550988_

>I suspect the metadata of the dataset series needs to be updated anyway. Especially if we keep the upstream inheritance explained in https://raw.githack.com/w3c/dxwg/dcat-dataseries-issue1272/dcat/index.html#dataset-series  where it is said: "The update date (dcterms:modified) of the dataset series should correspond to the latest publication or update date of the child datasets."
>Also,  we are having some parallel discussion of whether supporting inverse properties with a "lightweight approach" adopted by PROV-O (https://www.w3.org/TR/prov-o/#inverse-names) ( we have discussed this in tonight call, see meeting minutes)

riccardoAlbertoni:
This PR implements resolution https://www.w3.org/2022/02/22-dxwgdcat-minutes#r05

- it adds the inSeries's inverse, seriesMember, in the table of inverse, section 'use of the inverse
- it add the 'see also' to section 'use of inverse property' in the dcat:inSeries description
- it deletes the reference to issue 1335

When merged, It is expected to close #1335.


riccardoAlbertoni:
Do we want to add the use of dcat:memberSeries to the example 37 (https://w3c.github.io/dxwg/dcat/#ex-dataset-series-containment) and 38 (https://w3c.github.io/dxwg/dcat/#ex-dataset-series-releases)?

I am asking as I have noticed that in example 38 https://w3c.github.io/dxwg/dcat/#ex-dataset-series-releases, we used dcat:prev and its inverse, dcat:next.

dr-shorthair (approved):

dcat/index.html
```diff
@@ -4003,12 +4001,10 @@ <h2>Use of inverse properties</h2>
 <td id="inverse-of-catalog_catalog"><code>dcat:inCatalog</code></td>
 </tr>
 -->
-<!--
 <tr>
-<td><a href="#Property:resource_in_series"><code>dcat:inSeries</code></a></td>
+<td><a href="#Property:dataset_in_series"><code>dcat:inSeries</code></a></td>
 <td id="inverse-of-resource_in_series"><code>dcat:seriesMember</code></td>
```
agbeltran: should this be ```inverse_of_dataset_in_series``` instead of ```inverse-of-resource_in_series```; it's only the ```id``` but for clarity

riccardoAlbertoni: [09d354f](https://github.com/w3c/dxwg/pull/1461/commits/09d354fe2cd2a98abf8c7a54476f202a0c62ad80) fixed this and  has made the rest of ids in the table consistent with  it replacing "inverse-of-" by "inverse_of_"

@agbeltran  can you give your ok to this PR?

agbeltran (approved):

PR #1463, opened 2022-02-25
Fix typo in Example 15

### LINKED ISSUE #1459: Example 15: mistake in the used URIs?
Hi,

In Example 15 in https://w3c.github.io/dxwg/dcat/#dereferenceable-identifiers the example starts with

```
<https://dcat.example.org/id> a dcat:Dataset;
  adms:identifier  <https://example.org/iddoi> ;
  dcterms:publisher <https://example.org/PoelenJorritH> ;
```

but later in the example the URI  of the publisher is 

```
<https://dcat.example.org/PoelenJorritH> a foaf:Person;
  foaf:name "Jorrit H. Poelen" ;
  adms:identifier <https://example.org/PoelenJorritHID> ;
```

This breaks the graph. Namely the upper part is not connected with the bottem part in the RDF graph. It might be intentional, but the description below the example does not highlight this.

andrea-perego:
Addresses https://github.com/w3c/dxwg/issues/1459

### DIFF
--- a/dcat/examples/vocab-dcat-3/identifier-types.jsonld
+++ b/dcat/examples/vocab-dcat-3/identifier-types.jsonld
@@ -53,14 +53,14 @@
   }, {
     "@id" : "https://example.org/id",
     "@type" : "dcat:Dataset",
-    "dct:identifier" : {
+    "dcterms:identifier" : {
       "@type" : "xsd:anyURI",
       "@value" : "https://example.org/id"
     }
   }, {
     "@id" : "https://example.org/proxyid",
     "@type" : "dcat:Dataset",
-    "dct:identifier" : "id"
+    "dcterms:identifier" : "id"
   } ],
   "@context" : {
     "identifier" : {
@@ -115,8 +115,8 @@
     "rdfs" : "http://www.w3.org/2000/01/rdf-schema#",
     "geosparql" : "http://www.opengis.net/ont/geosparql#",
     "ex" : "https://dcat.example.org/",
-    "dct" : "http://purl.org/dc/terms/",
     "rdf" : "http://www.w3.org/1999/02/22-rdf-syntax-ns#",
+    "dcterms" : "http://purl.org/dc/terms/",
     "time" : "http://www.w3.org/2006/time#",
     "dcat" : "http://www.w3.org/ns/dcat#",
     "locn" : "http://www.w3.org/ns/locn#",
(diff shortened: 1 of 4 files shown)


riccardoAlbertoni (approved):

riccardoAlbertoni:
We might want to propagate the change in the rdf example files available under the directory dcat/examples and in the text explaining  example 15.

Moreover, I have noticed that in the REC we have used three different name spaces for example URIS: https://example.org/,  http://example.org/, and https://dcat.example.org/ 
Perhaps we should consider to use one only. In case, do we want to deal with the example namespace uniformity in a separate issue?

agbeltran (changes requested):
I agree with Riccardo's comments - line 4020 needs the same change: ```<code>https://example.org/PoelenJorritHID</code>``` to ```<code>https://dcat.example.org/PoelenJorritHID</code>```

andrea-perego:
> We might want to propagate the change in the rdf example files available under the directory dcat/examples and in the text explaining example 15.

Could you please take care of it, @riccardoAlbertoni ?

> Moreover, I have noticed that in the REC we have used three different name spaces for example URIS: https://example.org/, http://example.org/, and https://dcat.example.org/ Perhaps we should consider to use one only. In case, do we want to deal with the example namespace uniformity in a separate issue?

Yes, I think we'd better open a new issue on it.

andrea-perego:
> I agree with Riccardo's comments - line 4020 needs the same change: `<code>https://example.org/PoelenJorritHID</code>` to `<code>https://dcat.example.org/PoelenJorritHID</code>`

Thanks, @agbeltran . Just fixed via https://github.com/w3c/dxwg/pull/1463/commits/3b6f1f377ad330b1e52df3584419beb000c221f9

riccardoAlbertoni (approved):

davebrowning (approved):

andrea-perego:
> > I agree with Riccardo's comments - line 4020 needs the same change: `<code>https://example.org/PoelenJorritHID</code>` to `<code>https://dcat.example.org/PoelenJorritHID</code>`
> 
> Thanks, @agbeltran . Just fixed via [3b6f1f3](https://github.com/w3c/dxwg/commit/3b6f1f377ad330b1e52df3584419beb000c221f9)

@agbeltran , could you please check if your happy with this revision, and add your review?

agbeltran (approved):

PR #1468, opened 2022-02-26
correcting other two URIS

riccardoAlbertoni:
I suspect we should include also these changes to avoid mismatch in the URI and build the RDF graph correctly.
Please @andrea-perego, if you agree with these changes  merge them into your PR.

I have also updated the RDF files.


andrea-perego:
Thanks, @riccardoAlbertoni .

PR #1470, opened 2022-02-27
Add property dcat:resource

andrea-perego:
Summary of changes:
- Implemented proposal in https://github.com/w3c/dxwg/issues/1469
- Updated changelog
- Hidden changelog of DCAT2
- Removed self-reference to [VOCAB-DCAT-3]

Preview: https://raw.githack.com/w3c/dxwg/dcat-issue-1469/dcat/index.html

Diff: https://services.w3.org/htmldiff?doc1=https%3A%2F%2Fw3c.github.io%2Fdxwg%2Fdcat%2F&doc2=https%3A%2F%2Fraw.githack.com%2Fw3c%2Fdxwg%2Fdcat-issue-1469%2Fdcat%2Findex.html

Issue #1320, opened 2021-03-10
Changes to Class Diagram (figure 1).


riccardoAlbertoni:
In this issue, we pile the changes to make to the DCAT class diagram  (figure 1 in the DCAT document) to do them all at once.

riccardoAlbertoni:
~~Turn 'dct:' into 'dcterms' as for PR #1314 which has been merged.~~

andrea-perego:
Another related issue to be taken into account is https://github.com/w3c/dxwg/issues/1252

andrea-perego:
I think one of the things to be considered is whether we should limit the information in the diagram to increase its legibility. 

Some options:

1. We keep only classes and relationship, and remove datatype properties.
2. We provide different views of the diagram - e.g., a general and incomplete one for the voc, and more specific and detailed views for the main classes

dr-shorthair:
It was hard work making the Figure legible for DCAT2. So it might be time for option 2.

riccardoAlbertoni:
As a first step, we might simplify the figure by removing the inherited data properties.  This would still provide the overall view. What do you think?

dr-shorthair:
I did that originally, but then it was requested to show the 'inherited' properties as well. 
It might be that we need 
1. one summary class diagram showing the relationships between the DCAT classes (i.e. the backbone)
2. one diagram-per-class with both local and 'inherited' properties

Issue #1471, opened 2022-02-28
Editorial changes in view of DCAT3 4PWD


andrea-perego:
- [x]  Update changelog with the changes after 3PWD -  via https://github.com/w3c/dxwg/pull/1476 and https://github.com/w3c/dxwg/pull/1486
- [x] Drop the DCAT2 changelog - done via PR https://github.com/w3c/dxwg/pull/1470
- [x] Align DCAT3 diagram to current specification (remove inverse properties, add new properties) - done via PR https://github.com/w3c/dxwg/pull/1472
- [x] Update ack section via PR https://github.com/w3c/dxwg/pull/1464

riccardoAlbertoni:
The editor’s note in section https://w3c.github.io/dxwg/dcat/#inverse-properties  says 

>The list of properties in the following table is yet to be consolidated, and it includes also properties currently defined in [6. Vocabulary specification](https://w3c.github.io/dxwg/dcat/#vocabulary-specification).

Can we consider the list consolidated? 
In any case, I do not think we have any inverse properties that have been left in section 6

I would consider removing the note. What do the other editors think?

andrea-perego:
@riccardoAlbertoni said:

> The editor’s note in section https://w3c.github.io/dxwg/dcat/#inverse-properties says
> 
> > The list of properties in the following table is yet to be consolidated, and it includes also properties currently defined in [6. Vocabulary specification](https://w3c.github.io/dxwg/dcat/#vocabulary-specification).
> 
> Can we consider the list consolidated? In any case, I do not think we have any inverse properties that have been left in section 6
> 
> I would consider removing the note. What do the other editors think?

I think there's yet a property to be added, i.e., `dcat:inCatalog`, inverse of `dcat:resource`. After that, I would agree to remove the EDNOTE.

riccardoAlbertoni:
> I think there's yet a property to be added, i.e., `dcat:inCatalog`, inverse of `dcat:resource`. After that, I would agree to remove the EDNOTE.

Changes are done via #1485

andrea-perego:
I moved planned revisions on RDF files to https://github.com/w3c/dxwg/issues/1509

Closing this issue.

PR #1472, opened 2022-03-08
Update class diagram according to #1470

riccardoAlbertoni:
This PR  updates the class diagram according to the changed made by #1470  
- dcterms:hasPart has been added to dcat:Resource 
- dcterms:hasPart linking dcat:Catalog to dcat:Resource is replaced by its subproperty dcat:resource
It relates to #1320

Preview: https://raw.githubusercontent.com/w3c/dxwg/dcat-updateOfDiagram/dcat/images/dcat-all-attributes.svg


andrea-perego:
Thanks, @riccardoAlbertoni .

I wonder whether we should take this opportunity to do some housekeeping of the image files. We have them in two different subfolders, namely, `images` and `UML`, and I'm not sure they are the same versions. This is confusing.

I think it would be better to keep all of them under the `images` subfolder and delete the `UML` one.

WDYT?

andreasgeissner:
As you folks are working on the image already, may I just jump in and point out a typo? The connection dcat:DataService to dcat:Dataset reads dcat:serversDataset instead of dcat:servesDataset

riccardoAlbertoni:
I have implemented the suggestions made by @andrea-perego:
-  [getting rid of UML directory](https://github.com/w3c/dxwg/pull/1472/commits/83066620e03006df3dde68fc8cee671700d08441) 
- [removing from the diagram the inverse properties mentioned in section 7  as for #1413 and #1410](https://github.com/w3c/dxwg/pull/1472/commits/59dbba9efee4a5bd6019450069ae547a626c1f91)

and I have [fixed the typo serversDataset -> servesDataset](https://github.com/w3c/dxwg/pull/1472/commits/f94d5a3cb7d99b89b41021b0fdbbb1d35477e5fe)  thanks @andreasgeissner

andrea-perego:
Thanks, @riccardoAlbertoni .

I think we should also remove the cardinality restrictions:
- The only cardinality restriction specified in the RDF concerns property `dcat:record` - to be decided whether to keep it or not
- The cardinality 1..* of `dcat:resource` (previously on `dcterms:hasPart`) is controversial (see discussion in https://www.w3.org/2022/02/22-dxwgdcat-minutes#t03), and should probably be revised into 0..* 
- Cardinalities as 0..* are the default ones, so they needn't be specified

riccardoAlbertoni:
> I think we should also remove the cardinality restrictions:
> 
> * The only cardinality restriction specified in the RDF concerns property `dcat:record` - to be decided whether to keep it or not

I would keep this.

> * The cardinality 1..* of `dcat:resource` (previously on `dcterms:hasPart`) is controversial (see discussion in https://www.w3.org/2022/02/22-dxwgdcat-minutes#t03), and should probably be revised into 0..*

I am ok with changing this.

> * Cardinalities as 0..* are the default ones, so they needn't be specified

I am afraid without cardinalities some people might read 1 as default instead 0..* 
I think that the note below the figure explains quite well that cardinalities are not normative, and they are placed to provide usual expectations. 

Let's discuss this in tonight's call. 
What is the harm that you see in keeping them? 
Anyway, If we cancel the cardinalities we need to change the note as well.

andrea-perego:
@riccardoAlbertoni said:

> > * Cardinalities as 0..* are the default ones, so they needn't be specified
> 
> I am afraid without cardinalities some people might read 1 as default instead 0..* 

Not sure about that. If no cardinality is specified, there are no cardinality restrictions. But we can clarify this in the note.

> I think that the note below the figure explains quite well that cardinalities are not normative, and they are placed to provide usual expectations.

Yes, but if the cardinality is 0..* there is no need to make it explicit for the reason above.

> Let's discuss this in tonight's call. [...]

Agreed.

agbeltran:
@riccardoAlbertoni the ```dcat:Relationship``` part of the diagram has overlapping parts at the moment:

Screenshot 2022-03-08 at 21 48 22

dr-shorthair:
> I am afraid without cardinalities some people might read 1 as default instead 0..*

Yes, that was the original motivation. In standard UML the default is `1..1`

riccardoAlbertoni:
> @riccardoAlbertoni the `dcat:Relationship` part of the diagram has overlapping parts at the moment:
> 
> Screenshot 2022-03-08 at 21 48 22

@agbeltran  we have solved  this and other glitches via [e6788ef](https://github.com/w3c/dxwg/pull/1472/commits/e6788ef1f34467c1778a13ba88078e07431aab44)

riccardoAlbertoni:
> > I am afraid without cardinalities some people might read 1 as default instead 0..*
> 
> Yes, that was the original motivation. In standard UML the default is `1..1`

During the last DCAT meeting, we voted [resolution 03](https://www.w3.org/2022/03/08-dxwgdcat-minutes#r03) for the removal of all the cardinalities except those originally present in the RDF. The rationale is to reinforce the message that we do not normatively constraint them. 

According to the meeting discussion, providing general expectations might be misleading in some cases, and by the way,  cardinalities are not present for all the relations, which is another source of confusion.

[e0f25df](https://github.com/w3c/dxwg/pull/1472/commits/e0f25dfbef35545b95a31f8dec3c597da9a1a7a8) implements the voted resolution, dropping all the cardinalities and restating in the caption of the diagram that  "Except where specifically indicated, DCAT does not provide cardinality constraints."

andrea-perego (approved):

agbeltran (approved):

andrea-perego:
The changelog needs to be updated accordingly (https://github.com/w3c/dxwg/issues/1471)

PR #1480, opened 2022-03-17
Fixed dryad-globtherm-sdata example file for the DCAT2 examples

### LINKED ISSUE #1479: Issue with example file
This file provided as an example for DCAT2:
https://github.com/w3c/dxwg/blob/gh-pages/dcat/examples/dryad-globtherm-sdata.ttl
has the same problems fixed in https://github.com/w3c/dxwg/pull/1474 for the same file in the folder /examples/vocab-dcat-3

Can we fix directly or do we need to consider some errata given that is for DCAT2?

I note that the spec points to the file directly, see https://www.w3.org/TR/vocab-dcat-2/#dataset-publication

agbeltran:


riccardoAlbertoni (commented):
Thanks @agbeltran, 
I would keep dct as it is consistent with the other examples in the dcat 2.  We have been using dcterms only since dcat 3.

agbeltran:
thanks @riccardoAlbertoni, makes sense - and it is now changed to ```dct```

dcat/examples/dryad-globtherm-sdata.ttl
```diff
@@ -1,9 +1,12 @@
+@prefix dct: <http://purl.org/dc/terms/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
```
riccardoAlbertoni, line 4:
```suggestion
@prefix : <http://dcat.example.org/> .

```

riccardoAlbertoni (approved):

andrea-perego (approved):

PR #1501, opened 2022-04-10
Adding aligment to sdo:CreativeWorkSeries

### LINKED ISSUE #1398: Align to schema:CreativeWorkSeries
as suggested by @dr-shorthair in https://github.com/w3c/dxwg/pull/1397#pullrequestreview-747417547

 We should consider attempting some additional alignment with https://schema.org/CreativeWorkSeries

agbeltran:
Preview: https://raw.githack.com/w3c/dxwg/dcat-issue-1398/dcat/index.html#dcat-sdo

only the class and the dcat:inSeries property can be matched; don't think there is equivalents to dcat:prev, dcat:next; specific subclasses such as https://schema.org/TVSeries have more details but still no ways of linking previous/next item

closes #1398

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -6178,6 +6178,10 @@ ex:Test543L
                 <td>sdo:contactPoint</td>
             </tr>
             <tr>
+                <td>dcat:version</td>
+                <td>sdo:version</td>
+            </tr>
+            <tr>
                 <td><b>dcat:Catalog</b></td>
                 <td><b>sdo:DataCatalog</b></td>
             </tr>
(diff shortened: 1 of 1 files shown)


dr-shorthair (approved):
`next` and `previous` missing from Schema.org. Looks like a hole. Perhaps we should comment there? 

https://github.com/schemaorg/schemaorg/issues/3093

andrea-perego (changes requested):
Thanks for preparing this update, @agbeltran .

Probably we should also add the mapping between [`dcat:version`](https://w3c.github.io/dxwg/dcat/#Property:resource_version) and [`sdo:version`](https://schema.org/version). This could be included right after the mapping of `dcat:contactPoint` under `dcat:Resource`.

I can take care of this, if you agree, @agbeltran .

andrea-perego (changes requested):
I've just realised that there is an error: `dcat:inSeries` must go under `dcat:Dataset`, and not `dcat:DatasetSeries`.

andrea-perego (approved):
@agbeltran , I've moved `dcat:inSeries` under `dcat:Dataset` and added `dcat:version`. I've also moved up `dcat:DatasetSeries` - see https://github.com/w3c/dxwg/pull/1501/commits/fd9dc16956b2660764bdb11efa1e769a45e60d12

Please check if you're happy with the changes.

agbeltran:
thanks @andrea-perego - looks good, will merge

PR #1529, opened 2022-08-29
Correcting the DCAT class Diagram as Suggested by issue #1528

### LINKED ISSUE #1528: dcterms:hasPart is repeated twice in image
The hasPart property in the Class Catalog also exists in the Class resource. Therefore, it seems that the hasPart is written twice in the catalog of the diagram, and the dcat:resource that should be present is missing.

dcat-issue

riccardoAlbertoni:
Closing #1528


dr-shorthair (commented):
The correction is OK. 

However, I see that the `dct:relation` link back to `dcat:Resource` has been dropped from the diagram. Is this intentional?

riccardoAlbertoni:
> The correction is OK.
> 
> However, I see that the `dct:relation` link back to `dcat:Resource` has been dropped from the diagram. Is this intentional?

No, as far as I can recall, it wasn't intentional. I have restored dcterms:relation from dcat:Relationship to dcat:Resource in the diagram.
See https://raw.githack.com/w3c/dxwg/dcat-issue-1528/dcat/images/dcat-all-attributes.svg
Thanks.

dr-shorthair (approved):

davebrowning (approved):

PR #1535, opened 2022-09-28
Adjusting usage note dcterms:language

### LINKED ISSUE #1532: dcterms:language property in dcat:Distribution
In Belgium, we are often working with resources in different languages since we have 3 official languages. It can therefore happen that distributions are published in 3 or 4 languages at federal level, as our publications are intended for all publics. 

In paragraph 6.4.9 Property language, under Class: Cataloged Resource, usage note gives the following instruction:
_If representations of a dataset are available for each language separately, define an instance of dcat:Distribution for each language and describe the specific language of each distribution using dcterms:language (i.e., the dataset will have multiple dcterms:language values and each distribution will have just one as the value of its dcterms:language property)._
Nevertheless, in the vocabulary specification for Class: Distribution, the property language is not listed as specific to this class. 
Is it confirmed that we can use dcterms:language property for each distribution of a dataset available in different languages? 
How to deal with distributions that are the same for different languages? Can we use more than one dcterms:language property for a distribution?

riccardoAlbertoni:
This PR changes a Usage Note for <code>dcterms:language</code> in <code>dcat:Resource</code>
Considering the questions in issue #1532, we might want to be more explicit, suggesting that distribution can be multilingual.


andrea-perego (changes requested):
@riccardoAlbertoni , probably, instead of modifying the existing usage note, we should just add a sentence to cover the case of multilingual distributions - something like:

> In case of multilingual distributions, the distributions will have multiple <code>dcterms:language</code> values.

davebrowning (commented):
I'd vote for the addition @andrea-perego suggested.   It seems clearer to me!

riccardoAlbertoni:
I agree that adding Andrea's sentence is clearer. I will modify the PR accordingly.

dcat/index.html
```diff
@@ -1271,7 +1271,7 @@ <h4>Property: language</h4>
...
-                <tr><th class="prop">Usage note:</th><td>If representations of a dataset are available for each language separately, define an instance of <code>dcat:Distribution</code> for each language and describe the specific language of each distribution using <code>dcterms:language</code> (i.e., the dataset will have multiple <code>dcterms:language</code> values and each distribution will have just one as the value of its <code>dcterms:language</code> property).</td></tr>
+                <tr><th class="prop">Usage note:</th><td>If representations of a dataset are available for each language or group of languages separately, define an instance of <code>dcat:Distribution</code> for each language or group and describe the specific languages of each distribution using <code>dcterms:language</code> (e.g., the dataset will have multiple <code>dcterms:language</code> values and each distribution will have just one as the value of its <code>dcterms:language</code> property).</td></tr>
```
```suggestion
                <tr><th class="prop">Usage note:</th><td>If representations of a dataset are available for each language separately, define an instance of <code>dcat:Distribution</code> for each language and describe the specific language of each distribution using <code>dcterms:language</code> (i.e., the dataset will have multiple <code>dcterms:language</code> values and each distribution will have just one as the value of its <code>dcterms:language</code> property). In case of multilingual distributions, the distributions will have multiple <code>dcterms:language</code> values.</td></tr>
```

andrea-perego (approved):

andrea-perego:
Thanks, @riccardoAlbertoni . I think we also need to update the changelog and the ack section.

andrea-perego:
> Thanks, @riccardoAlbertoni . I think we also need to update the changelog and the ack section.

Just updated the ack section via commit https://github.com/w3c/dxwg/pull/1535/commits/5f358e6d52b2848cb31fc7325e4bdc508badcbf8

andrea-perego:
> Thanks, @riccardoAlbertoni . I think we also need to update the changelog and the ack section.

Changelog updated via commit https://github.com/w3c/dxwg/pull/1535/commits/d179d13a29f726220427e74b3de6c406bbf16bcd

riccardoAlbertoni:
Thanks, @andrea-perego. 
The DIFF is available at https://services.w3.org/htmldiff?doc1=https://w3c.github.io/dxwg/dcat/&doc2=https://raw.githack.com/w3c/dxwg/dcat-issue-1532/dcat/index.html

@davebrowning or other editors: could you accept the PR?

andrea-perego:
@riccardoAlbertoni , I made an editorial fix to avoid that abstract and sotd sections are included in the ToC - see commit https://github.com/w3c/dxwg/pull/1535/commits/f22e792e118d1efc11f48cbb29119e49c5d4bea9

davebrowning (approved):

Issue #1536, opened 2022-10-14
Definition of dcat:spatialResolutionInMeters incompatible/problematic with JSON-LD


jakubklimek:
There is a technical issue with the definition of `dcat:spatialResolutionInMeters` when used with JSON-LD.
Specifically, the issue is the range being `xsd:decimal`.

In JSON-LD, `xsd:decimal` is not supported for numbers, see the note [in the specification](https://www.w3.org/TR/json-ld11/#conversion-of-native-data-types). Therefore, when in JSON this number is actually a JSON number, not JSON string, even if the JSON-LD context specifies the datatype to be explicitly `xsd:decimal`, the number is treated as `xsd:double`, leading to conversions like `1.2E1` instead of `12` when loading JSON-LD as RDF, in turn leading to invalid `xsd:decimal` RDF literal `"1.2E1"^^xsd:decimal`. This can be seen e.g. in [JSON-LD playground](https://json-ld.org/playground/#startTab=tab-nquads&json-ld=%7B%22%40context%22%3A%7B%22xsd%22%3A%22http%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23%22%2C%22decimal%22%3A%7B%22%40id%22%3A%22http%3A%2F%2Flocalhost%2Fdecimal%22%2C%22%40type%22%3A%22xsd%3Adecimal%22%7D%7D%2C%22decimal%22%3A12.1%7D).

Possible solutions:
1. We change the datatype to `xsd:integer` or `xsd:double`, which is a breaking change
2. We explicitly mention, that JSON-LD has this issue, and the property data needs to be represented as JSON String, not JSON Number

pchampin:
> leading to conversions like `1.2E1` instead of `12`

That is not accurate. In JSON-LD, `12` will still render as `12`.

But it is true that `12.1` will render as `1.21E1` (as in your example), and that is indeed an issue.

I believe that xsd:decimal is superior to xsd:double in many respects (see [this paper](https://link.springer.com/chapter/10.1007/978-3-031-06981-9_10) ([pdf](https://2022.eswc-conferences.org/wp-content/uploads/2022/05/paper_98_Keil_et_al.pdf)) for a detailed analysis), so I am in favour of option 2.

pchampin:
Another option would be the following:
define a new RDF datatype dcat:decimal
- its value space is exactly the same as xsd:decimal (si that it is semantically backward compatible with xsd:decimal)
[...]

PR #1543, opened 2022-11-15
Address 1536: extend the range of 'spatial resolution'

pchampin:
The range now allows `xsd:double` in addition to `xsd:decimal`.
I also added a note in 6.6.5 to explain the rationale of that change.

Should the note be copied in 6.8.13 as well?

Addresses #1536

Preview:
- https://raw.githack.com/w3c/dxwg/issue-1536/dcat/index.html#Property:dataset_spatial_resolution 
- https://raw.githack.com/w3c/dxwg/issue-1536/dcat/index.html#Property:distribution_spatial_resolution

Diff:
- https://services.w3.org/htmldiff?doc1=https://w3c.github.io/dxwg/dcat/&doc2=https://raw.githack.com/w3c/dxwg/issue-1536/dcat/index.html#Property:dataset_spatial_resolution
- https://services.w3.org/htmldiff?doc1=https://w3c.github.io/dxwg/dcat/&doc2=https://raw.githack.com/w3c/dxwg/issue-1536/dcat/index.html#Property:distribution_spatial_resolution


dr-shorthair (commented):
I think this will require a matching change to the TTL representation 
https://github.com/w3c/dxwg/blob/gh-pages/dcat/rdf/dcat3.ttl#L1554

jakubklimek:
Do I understand it correctly, that the idea is that JSON-LD contexts of DCAT and DCAT profiles can choose one of:

1. `xsd:double` datatype and JSON number representation
2. `xsd:decimal` datatype and JSON string representation?

pchampin:
@dr-shorthair 
> I think this will require a matching change to the TTL representation

Right, but I'm not sure exactly what the process is for this (are some versions generated from others? difference between dcat3.* and dcat3-external.*...) so I would rather leave this to the editors.

pchampin:
> Do I understand it correctly, that the idea is that JSON-LD contexts of DCAT and DCAT profiles can choose one of:
> 
> 1. `xsd:double` datatype and JSON number representation
> 2. `xsd:decimal` datatype and JSON string representation?

or they can refrain to chose, and leave it to end users... Actually, both options are possible, as demonstrated in [this example](https://json-ld.org/playground/#startTab=tab-nquads&json-ld=%7B%22%40context%22%3A%7B%22rdfs%22%3A%22http%3A%2F%2Fwww.w3.org%2F2000%2F01%2Frdf-schema%23%22%2C%22xsd%22%3A%22http%3A%2F%2Fwww.w3.org%2F2001%2FXMLSchema%23%22%2C%22dcat%22%3A%22http%3A%2F%2Fwww.w3.org%2Fns%2Fdcat%23%22%2C%22spatialResolutionInMeters%22%3A%22dcat%3AspatialResolutionInMeters%22%2C%22spatialResolutionInMetersDouble%22%3A%7B%22%40id%22%3A%22dcat%3AspatialResolutionInMeters%22%2C%22%40type%22%3A%22xsd%3Adouble%22%7D%2C%22spatialResolutionInMetersDecimal%22%3A%7B%22%40id%22%3A%22dcat%3AspatialResolutionInMeters%22%2C%22%40type%22%3A%22xsd%3Adecimal%22%7D%7D%2C%22%40graph%22%3A%5B%7B%22spatialResolutionInMeters%22%3A3.14%2C%22rdfs%3Acomment%22%3A%22this%20will%20produce%20an%20xsd%3Adouble%2C%20which%20is%20now%20correct%22%7D%2C%7B%22spatialResolutionInMeters%22%3A%7B%22%40value%22%3A%223.14%22%2C%22%40type%22%3A%22xsd%3Adecimal%22%7D%2C%22rdfs%3Acomment%22%3A%22this%20will%20produce%20an%20xsd%3Adecimal%2C%20which%20is%20still%20correct%22%7D%2C%7B%22spatialResolutionInMetersDecimal%22%3A%223.14%22%2C%22rdfs%3Acomment%22%3A%22with%20this%20alias%2C%20you%20don't%20neet%20to%20give%20the%20type%20explicitly%2C%20BUT%20you%20MUST%20use%20a%20string%20(otherwise%20an%20invalid%20xsd%3Adecimal%20might%20be%20produced%22%7D%2C%7B%22spatialResolutionInMetersDouble%22%3A%223.14%22%2C%22rdfs%3Acomment%22%3A%22with%20this%20alias%2C%20you%20will%20always%20produce%20an%20xsd%3Adouble%2C%20even%20if%20you%20use%20a%20string%20--%20or%20an%20integer%2C%20for%20that%20matter%22%7D%5D%7D )

jakubklimek:
@pchampin 
> or they can refrain to chose, and leave it to end users... 

Well, technically, yes. But then the role of JSON-LD in "hiding RDF complexity for JSON users" is somehow diminished if we ask "JSON developers" to choose between
```json
"spatialResolutionInMeters": {
   "@value": "3.14",
   "@type": "xsd:decimal"
}
```
and
```json
"spatialResolutionInMeters": {
   "@value": 3.14,
   "@type": "xsd:double"
}
```
for, in JSON world, no apparent reason.

pchampin:
@jakubklimek 
Developers that don't want to worry about the underlying RDF will simply use
```
  "spatialResolutionInMeters": 3.14
```
and that will be compliant with DCAT, thanks to the proposed change.

Developers who care about those details still have the opportunities to be more precise and use an `xsd:decimal`.

jakubklimek:
> "spatialResolutionInMeters": 3.14

But this is not that simple. If they go this way, the will probably also use:
```json
"spatialResolutionInMeters": 3
```
which produces a `xsd:integer` literal, which, yes, can be converted into `xsd:decimal` or `xsd:double`, but strictly speaking, is yet another datatype. That's why it is necessary to state whether the value is a `xsd:decimal` or `xsd:double` in the context (one or the other), or in the value.

pchampin:
@jakubklimek  strictly speaking `xsd:integer` is derived from `xsd:decimal` (see [spec](https://www.w3.org/TR/xmlschema11-2/#integer)), so having an `xsd:integer` where an `xsd:decimal` is perfectly legal semantically. OWL inference engines, for example, will not treat it as a contradiction.

Now, SHACL and ShEX are only concerned with syntax, and a shape requiring `xsd:decimal` will indeed choke on an `xsd:integer` value... A pragmatic approach would be for such shape to allow `xsd:decimal` OR `xsd:integer` whenever the domain is `xsd:decimal`. Note that this is an incomplete workaround, because `xsd:decimal` has many more derived datatype that could semantically be used, but `xsd:integer` is probably by far the most used.

I have added a comment (bcb5990) explaining this in the note below the definition of `dcat:spatialResolutionInMeters`.

NB: this issue with `xsd:integer` is not specific to JSON-LD. The same problem exists with Turtle...

dr-shorthair:
@andrea-perego can you clarify your position on this?

dr-shorthair:
It is arguably out of scope for DCAT to address issues in the JSON implementation of RDF. 
It would be appropriate to add a warning that `xsd:decimal` is not supported in JSON, but I would be reluctant to add an alternative datatype, as too many choices compromises interoperability.

riccardoAlbertoni (approved):

riccardoAlbertoni:
I have approved this PR as a consequence of the "rephasing and removing of normative effect" PR #1573  just merged here.

@dr-shorthair @andrea-perego @agbeltran @davebrowning @pwin : Can any of you approve  and merge the resulting  PR #1543 into the main branch.

dr-shorthair (approved):

PR #1545, opened 2022-11-17
Addressing #1540 - schema.org aligment

### LINKED ISSUE #1540: Alignment with schema.org
Dear all,

I was performing a mapping exercise between DCAT and Schema.org and I found the section:

https://www.w3.org/TR/vocab-dcat-2/#dcat-sdo

Towards the end of the section, there is a a recommended mapping pointing to the RDF file: https://w3c.github.io/dxwg/dcat/rdf/dcat-schema.ttl

The file includes mapping with SOSA concepts as well which I guess they should be outside this file.

In addition, I believe the following mappings are not correct: 

- sdo:DataCatalog and dcat:Catalog, currently mapped as equivalentClass, however dcat:Catalog includes in the definition DataServices, therefore sdo:DataCatalog (defined as "collection of datasets") should be a subClassOf of dcat:Catalog

- sdo:DataDownload and dcat:Ditstribution, currently mapped as equivalentClass, however in the usage note of dcat:Distribution  is written:
 This represents a general availability of a dataset. It implies no information about the actual access method of the data, i.e. whether by direct download, API, or through a Web page. 

thus sdo:DataDownload should be a subClassOf dcat:Distribution.

Thanks for looking into 

Emidio

---

dr-shorthair:
Fix alignments - dcat:Catalog-sdo:DataCatalog dcat:Distribution-sdo:DataDownload

Fixes #1540

### DIFF
--- a/dcat/rdf/dcat-schema.ttl
+++ b/dcat/rdf/dcat-schema.ttl
@@ -51,8 +51,8 @@ dct:temporal
 dct:type
   skos:closeMatch schema:additionalType ;
 .
-dcat:Catalog
-  owl:equivalentClass schema:DataCatalog ;
+schema:DataCatalog
+  rdfs:subClassOf dcat:Catalog ;
 .
 dcat:DataService
   rdfs:comment "Unclear if a DataFeed is a data service, or a data collection. " ;
@@ -61,8 +61,8 @@ dcat:DataService
 dcat:Dataset
   owl:equivalentClass schema:Dataset ;
 .
-dcat:Distribution
-  owl:equivalentClass schema:DataDownload ;
+schema:DataDownload
+  rdfs:subClassOf dcat:Distribution ;
 .
 dcat:Resource
   rdfs:subClassOf schema:Thing ;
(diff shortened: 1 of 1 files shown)


riccardoAlbertoni:
Thanks, Simon, for preparing this PR.
I understand we want to leave out the mapping with SOSA, but I have doubts about deleting the mapping with FOAF.   FOAF, in my opinion, is comparable to DCT, for which we keep the mapping. DCAT directly uses only foaf:Agent, but mapping foaf:Agent's subclasses and properties might be relevant. Do you think it would do any harm to keep them?

dr-shorthair:
I've reinstated the FOAF mappings ... though it does seem a little cheeky given that @danbri is the boss of both.

riccardoAlbertoni (approved):

agbeltran:
Should we update the [turtle file](https://w3c.github.io/dxwg/dcat/rdf/dcat-schema.ttl) in this same PR?

dr-shorthair:
This PR only touches the TTL file.

agbeltran:
sorry, yes! I was thinking about the spec but we don't add the relationship between entities in the table!

agbeltran (approved):

PR #1572, opened 2023-05-25
Fixing issue  #1570

### LINKED ISSUE #1570: adms references
Hi,

lately ADMS got a new release. The new release did not changed any semantics of the terms, but now there is content negotation on the  namespace and the usage beyond DCAT of ADMS is made easier
By processing the DCAT specification, I spotted different usages how the ADMS terms are referenced. In brackets is the motivation for the change.

- [versionNotes](https://www.w3.org/TR/vocab-adms/#adms-versionnotes)
      => the link for the RDF property should be http://www.w3.org/ns/adms#versionNotes (the URI and not to the html specification, like for other properties)

- [status](https://www.w3.org/TR/vocab-adms/#adms-status)
     => the link for the RDF property should be http://www.w3.org/ns/adms#status (the URI and not the html specification, like for the other properties)
 
- under example https://w3c.github.io/dxwg/dcat/#ex-proxy-id
    => the link for the RDF property adms:identifier should be http://www.w3.org/ns/adms#identifier (the URI will dereference to the html now and thus the link is agnostic to the publication location)
     => the link for the RDF property adms:schemaAgency should be http://www.w3.org/ns/adms#schemaAgency (the URI will dereference to the html now and thus the link is agnostic to the publication location)

riccardoAlbertoni:
We used to refer to the ADMS document URL instead of Term URI.

With these changes we refer to w3c URI which turn into  non W3C - ADMS latest documentation, https://semiceu.github.io/ADMS/releases/2.00/.

Closed #1570


davebrowning (dismissed):

davebrowning (commented):
The file changes seem (to me) to address the specifics of the issue #1570 .  There does seem to be some concerns that need to be discussed a little before moving ahead with this at this point, so I've revoked my approval pending that discussion happening in the issue.

andrea-perego (changes requested):
@riccardoAlbertoni , I'm not in favour of these revisions, for the reasons explained in https://github.com/w3c/dxwg/issues/1570#issuecomment-1656943616

I think we should keep the current references, pointing to the ADMS version in W3C space.

PR #1573, opened 2023-05-25
Rephasing  and removing the normative effect of proposal PR #1543

riccardoAlbertoni:
The PR tries to mediate between the proposal made in PR #1543 and the worries about impacting the normative part by providing two datatypes. 

@pchampin, @dr-shorthair: Could this represent a viable compromise? Please refine it if needed.


dr-shorthair (approved):
Modest proposal is fine for me

dcat/index.html
```diff
@@ -2639,7 +2639,7 @@ <h4>Property: spatial resolution</h4>
...
+            <p>As for the use of datatype, note that [[JSON-LD]] converts numbers to <a data-cite="XMLSCHEMA11-2#double"><code>xsd:double</code></a> by default, and properly generating <a data-cite="XMLSCHEMA11-2#decimal"><code>xsd:decimal</code></a> requires the use of strings with an explicit or coerced datatype. In [[Turtle]], seemingly minor modifications can change the datatype of a value: `100.0` is an <a data-cite="XMLSCHEMA11-2#decimal"><code>xsd:decimal</code></a>, while `1e2` is an <a data-cite="XMLSCHEMA11-2#double"><code>xsd:double</code></a>.</p>
```
pchampin:
```suggestion
            <p>As for the use of datatype, note that [[JSON-LD]] converts numbers to <a data-cite="XMLSCHEMA11-2#double"><code>xsd:double</code></a> or <a data-cite="XMLSCHEMA11-2#integer"><code>xsd:integer</code></a>, and properly generating <a data-cite="XMLSCHEMA11-2#decimal"><code>xsd:decimal</code></a> requires the use of strings with an explicit or coerced datatype. In [[Turtle]], seemingly minor modifications can change the datatype of a value: `100.0` is an <a data-cite="XMLSCHEMA11-2#decimal"><code>xsd:decimal</code></a>, while `1e2` is an <a data-cite="XMLSCHEMA11-2#double"><code>xsd:double</code></a>.</p>
```
Numbers are converted to float or integer depending on their fractional part (note that 100.0 in JSON-LD  ends up as "100"^^xsd:integer, despite the trailing ".0" .
I also suggest to remove "by default" because there is no way to change how *numbers* are converted to literals. The workaround is to not use numbers at all.

pchampin (approved):
Sorry for the delay. I'm ok with not changing the schema. I made a suggestion to slightly improve the text, but I can live with the original text as well.

PR #1607, opened 2024-06-10
charter extension

pchampin:


pchampin:
Please note that these changes are already reflected on https://www.w3.org/2022/06/dx-wg-charter.html .
The goal of this PR is merely to keep our repo in sync with the published charter.

PR #1611, opened 2024-09-06
Add Czech translation of new DCAT3 classes and properties to RDF Turtle

jakubklimek:
Add Czech translation of new DCAT3 classes and properties to RDF Turtle.

I do not know which tool was used to convert the Turtle file to RDF/XML and JSON-LD. I tried Apache Jena `riot`, but that ended up with a completely different structure of the output files, which might not be desirable. Therefore, I leave the conversion to the maintainers at the moment.


dr-shorthair (approved):

riccardoAlbertoni (changes requested):
Thanks @jakubklimek. Please include the JSON-LD and RDF/XML files as well. I realize this will trigger a large diff, but I think it is preferable to keep all serializations fully aligned.

In the past, I tried using the "pretty" options for the conversion [(see Jena documentation)](https://jena.apache.org/documentation/io/rdf-output.html), but as far as I remember, they don't guarantee that the triple order is maintained.

One last note—and **please don't worry about this for the current PR**, it’s just for future contributions: when submitting a PR, we usually create a branch directly on the W3C DXWG repo ( usually named DCAT-issue-number_of_Issue) instead of working on our  personal fork of `gh-pages`. I guess this is mainly due to established habits, and we could discuss evolving this practice. However, I suspect that having the branch on the main repo helps others chime in more easily and improves the traceability of the contribution history.

jakubklimek:
@riccardoAlbertoni Will do.

Regarding the the branch/fork, this is due to the PR being from 2024 when I was not part of the WG and could not create a branch.

I tried the pretty option too, but as you pointed out, it does not preserve the ordering, still resulting in large diff.

jakubklimek:
@riccardoAlbertoni done.
Actually, it seems that the original RDF/XML was generated by Jena, while JSON-LD was generated by [Zazuko](https://converter.zazuko.com/) based on the style.

riccardoAlbertoni (approved):

PR #1618, opened 2024-09-30
Changing the files returned by dereferentiation of dcat ns.

### LINKED ISSUE #1614: https://www.w3.org/ns/dcat is still pointing to DCAT 2.0
As the recommendation of DCAT 3 has been published it should be best that the page https://www.w3.org/ns/dcat is updated to version 3.
It still reflects version 2.0

kr,

Bert

riccardoAlbertoni:
Fixes  #1614
Closes #1614 

I have prepared a new version for the HTML accessible via https://www.w3.org/ns/dcat#

@pchampin, My understanding is that we need to accept this PR, and then a copy of the files needs to be included in the proper w3c web directory.

Am I right?


riccardoAlbertoni:
as far as I can understand, the result of this PR is still coherent with what we have written in 
 https://www.w3.org/TR/vocab-dcat-3/#RDF-representation

andrea-perego (approved):

pwin (approved):
Looks ok to me, but may need final move by @pchampin

pchampin:
LGTM, except for two things
* there were two syntax errors in dcat-external.ttl which I just fixed
* dcat2014.ttl was not consistent with https://www.w3.org/ns/dcat2014.ttl, and not isomorphic to dcat2014.rdf

Both issues are fixed by my 2 commits above.

I'm happy to publish this on w3.org/ns once this is merge. Do we need a WG decision for this? (could be done on WBS)?

riccardoAlbertoni:
@pchampin the changes work to me, I am going to merge the PR so that you can publish the file in the website. Thanks

Issue #1630, opened 2025-08-12
catalogRecord/conformsTo Definition does not match Usage note


PeterParslow:
The definition of CatalogRecord/conformsTo is

"An established standard to which the described resource conforms."

https://www.w3.org/TR/vocab-dcat/#Property:record_conforms_to

The usage note says:

"This property SHOULD be used to indicate the model, schema, ontology, view or profile that the catalog record metadata conforms to."

The usage note makes more sense. The definition seems to be an erroneous copy of the definition of CataloguedResource/conformsTo (https://www.w3.org/TR/vocab-dcat/#Property:resource_conforms_to) - where it does make sense.

Note, European Commission / SEMIC DCAT maps ISO 19115 MD_Metadata.standardName to catalogRecord/conformsTo, which makes logical sense but does not fit the current DCAT documentation. https://semiceu.github.io/GeoDCAT-AP/releases/3.0.0/#metadata-standard-name-metadata-standard-version

Suggestion: change the definition at https://www.w3.org/TR/vocab-dcat/#Property:record_conforms_to to "An established standard to which the catalog record conforms."

PR #1631, opened 2025-08-12
Amend definition of catalogRecord/conformsTo

PeterParslow:
See issue https://github.com/w3c/dxwg/issues/1630

I accept that the index.html may have been created from other files, in which case the fix should be put in the appropriate "fragment" file - but I couldn't find that!

### DIFF
--- a/dcat/index.html
+++ b/dcat/index.html
@@ -2400,7 +2400,7 @@ ex:table-service-005
 
             <table class="def">
                 <tr><th>RDF Property:</th><td><a href="http://purl.org/dc/terms/conformsTo"><code>dcterms:conformsTo</code></a></td></tr>
-                <tr><th class="prop">Definition:</th><td>An established standard to which the described resource conforms.</td></tr>
+                <tr><th class="prop">Definition:</th><td>An established standard to which the catalog record conforms.</td></tr>
                 <tr><th class="prop">Range:</th><td><a href="http://purl.org/dc/terms/Standard"><code>dcterms:Standard</code></a> (A basis for comparison; a reference point against which other things can be evaluated.)</td></tr>
                 <tr><th class="prop">Usage note:</th><td>This property SHOULD be used to indicate the model, schema, ontology, view or profile that the catalog record metadata conforms to. </td></tr>
                 </table>


dr-shorthair:
I disagree with this change. 
The text quotes the DC definition, and should not be changed even for this context.

dr-shorthair (commented):
Quotes the original definition. Reject the change.

PeterParslow:
> Quotes the original definition. Reject the change.

Note: this change is specific to 6.5.6 the use of dcterms:conformsTo in the context of Catalog Record

If Simon's is the consensus view, then please amend the Usage note to make it even more explicit that this is the standard to which the catalog record conforms, not the standard to which the resource conforms. i.e. in this context the "described resource" is the catalog record.

jakubklimek:
Actually, there is a similar situation with `dcterms:title` reuse:
- Distribution: `A name given to the distribution.` instead of the original `A name given to the resource.`
- Catalog Record: `A name given to the record.` instead of the original `A name given to the resource.`

So the proposed change would make this consistent.

From this point of view, DCAT is an application profile of Dublin Core (and other reused vocabularies), and therefore can afford to make these context-specific changes.

We could perhaps discuss whether to refer to the record as `record` or `catalog record` in these occurrences.

dr-shorthair:
I'm reluctant to change any part of the canonical definition unless we are minting a new concept with a new URI. Modified text may go in a scope-note.

jakubklimek:
That is a separate, bigger issue, then.

Because there are already multiple cases in DCAT where this is happening (as demonstrated above), the same thing is also happening in DCAT APs. It is then a discussion about whether to make terminological changes when reusing generic terms from other vocabularies or not, and at the moment, DCAT seems to be doing that, so at least it should be consistent with it.

dr-shorthair:
Yes, I agree. 

The text of the definition of a term defined by an external authority must not be different from that provided by that external authority. 
If embellishment related to the context is needed, that should be in a scope note.

PeterParslow:
So,
1. adjust my request to add a note "in this context, the 'described resource' is the Catalog Record"
2. adjust the definition in Distribution back to match the source & add a similar note "in this context, the 'described resource' is the Distribution"

PR #1638, opened 2026-06-11
Fixing DCAT RDF

### LINKED ISSUE #1626: dcat.ttl/rdf/...: Typo in `skos:scopeNote` for `dcat:DataService`
> Line 275 in TTL:
>  skos:scopeNote "Se un dcat:DataService è associato a uno o più Dataset specificati, questi sono indicati dalla proprietà dcat:serveDataset."@it ;

It should be `dcat:servesDataset`.

### LINKED ISSUE #1627: Catalogued resource vs Cataloged Resource
In https://www.w3.org/ns/dcat3.ttl we have:
```turtle
dcat:Resource
  rdfs:isDefinedBy <https://www.w3.org/TR/vocab-dcat-2/> ;
  rdfs:label "Catalogued resource"@en ;
```

In the [spec](https://www.w3.org/TR/vocab-dcat-3/#Class:Resource), we have:

Image

Note:
- Catalogued => Cataloged
- resource => Resource

Is this inconsistency intended? If not, should DCAT contain labels for both `@en-US` and `@en-GB`? Or is there a preference for one?

riccardoAlbertoni:
Closes #1626  
Closes #1627

This PR updates the DCAT RDF serialization based on suggestions from the issues mentioned above.

The changes are editorial in nature and do not affect the normative REC.

### DIFF
--- a/dcat/rdf/dcat.jsonld
+++ b/dcat/rdf/dcat.jsonld
@@ -1,30 +1,25 @@
 {
   "@graph" : [ {
-    "@id" : "_:b0",
-    "homepage" : "https://www.csiro.au",
-    "foaf:name" : "Commonwealth Scientific and Industrial Research Organisation"
-  }, {
-    "@id" : "_:b1",
-    "@type" : "foaf:Person",
-    "seeAlso" : "https://orcid.org/0000-0003-3499-8262",
-    "memberOf" : "_:b2",
-    "homepage" : "https://agbeltran.github.io",
-    "foaf:name" : "Alejandra Gonzalez-Beltran"
-  }, {
-    "@id" : "_:b11",
+    "@id" : "_:b10",
     "@type" : "org:Organization",
     "seeAlso" : "https://www.wikidata.org/entity/Q37033",
     "homepage" : "https://www.w3.org/",
     "foaf:name" : "World Wide Web Consortium (W3C)"
   }, {
+    "@id" : "_:b11",
+    "@type" : "foaf:Person",
+    "seeAlso" : "https://orcid.org/0000-0002-3884-3420",
+    "memberOf" : "_:b2",
+    "foaf:name" : "Simon J D Cox",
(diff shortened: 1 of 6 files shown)


jakubklimek:
@riccardoAlbertoni I understand the resolution in `dcat.ttl` and `dcat3.ttl`, but the diff of the other files (JSON-LD and RDF/XML) is way too big for just fixing 2 typos. Is it just due to the randomness of ordering, or is something else happening there? Wouldn't it be better to fix the typos manually, so that the diff is clear?

riccardoAlbertoni:
Let me provide a little more context.

So far, we've typically updated the TTL file and then generated the JSON and RDF/XML versions using Jena's riot tool or similar. This approach helps reduce both manual effort and the risk of introducing syntax errors.

In practice, this means you'd check the changes in the TTL file and trust the riot translation for the other serializations ;)

Of course, our past approach doesn't mean we can't change how we operate in the future. However, I think the balance between effort and advantages needs to be considered. Personally, I'd support any approach that avoids additional manual editing, especially considering that the RDF serializations are important but mainly meant for machine consumption and are not normative.

jakubklimek:
OK, I thought so, but wanted to make sure.

jakubklimek (approved):

pchampin (approved):

pchampin:
It might be clearer to keep in this `dcat/rdf` only the "master" files (i.e. dcat3.ttl), possibly with a script used to generate the other files.

I would also suggest to remove the duplicate "version-less" files (`dcat.ttl`, `dcat.jsonld`) as those are managed via redirections on `w3.org/ns` (see [here](https://github.com/w3c/ns/blob/main/dcat.ttl.redirect)).

But those are orthogonal concerns to the meat of this PR...

pchampin:
@riccardoAlbertoni any objection to me pushing those changes to w3.org/ns/?

riccardoAlbertoni:
No objections. Thanks

PR #1644, opened 2026-08-14
Replace dcat:target with dct:relation

### LINKED ISSUE #1629: Non-existing DCAT property used in relation examples
I noticed that the non-existent property `dcat:target` is mentioned in the relationship examples. 

Concretely in the turtle example file: 
https://github.com/w3c/dxwg/blob/6d6a954f2ccd085447adf5e02303d8535a6ea722/dcat/examples/relation-examples.ttl#L33 
and in the json-ld example file:
https://github.com/w3c/dxwg/blob/6d6a954f2ccd085447adf5e02303d8535a6ea722/dcat/examples/relation-examples.jsonld#L6
https://github.com/w3c/dxwg/blob/6d6a954f2ccd085447adf5e02303d8535a6ea722/dcat/examples/relation-examples.jsonld#L77

I think the property that should be used in these examples, [according to the specification](https://www.w3.org/TR/vocab-dcat-3/#qualified-relationship), should be `dcterms:relation`.

dr-shorthair:
closes #1629


dcat/examples/vocab-dcat-3/relation-examples.rdf
leskneebone, line 1: fixed dct: --> dcterms:

leskneebone (approved):
just some mismatched dct/dcterms prefixes fixed - applied the latter throughout in the end. not sure if there is a preferred prefix operating in dcat.

riccardoAlbertoni (approved):

